PR #19, opened 2022-01-02
Fixes #15: Added CodeQL Workflow

### LINKED ISSUE #15: Add CodeQL Analysis Workflow
Adding [CodeQL](https://github.com/github/codeql-action) workflow in repo.

If this seems fine please assign this to me

uzaxirr:
It's similar to one in [IntelOwl](https://github.com/intelowlproject/IntelOwl/blob/master/.github/workflows/codeql-analysis.yml)

Fixes and Closes #15 

## Screenshots
Screenshot from 2022-01-02 16-01-16

### DIFF
--- a/.github/workflows/Create codeql-analysis.yml
+++ b/.github/workflows/Create codeql-analysis.yml
@@ -1,0 +1,90 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, dev ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main, dev ]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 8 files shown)


.github/workflows/Create codeql-analysis.yml
```diff
@@ -0,0 +1,90 @@
...
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
```
mlodic: could you just set 3.9 as it is the version used by the application?

.github/workflows/Create codeql-analysis.yml
```diff
@@ -0,0 +1,90 @@
...
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends apt-utils libsasl2-dev libssl-dev libldap2-dev python-dev libfuzzy-dev net-tools git osslsigncode exiftool
```
mlodic: well this is just copy-pasted from IntelOwl. Greedybear does not have the same dependencies

.github/workflows/Create codeql-analysis.yml
uzaxirr: Done!

.github/workflows/Create codeql-analysis.yml
uzaxirr: Yeah, I was about to ask that. I thought it's based on the same tech stack so I assumed it to consume those dependencies. Thanks for the clarification.

Issue #23, opened 2022-01-02
Elasticsearch installation error


uzaxirr:
i'm encountering some error while setting up GreedyBear locally
After doing the `docker-compose -p greedybear up` cmd.
It originates from settings.py where Elasticsearch client is being initialized. The ELASTIC_ENDPOINT variable in my env file is empty
![Screenshot from 2022-01-02 19-51-22](https://user-images.githubusercontent.com/72073401/147880136-9c712db0-3f63-481d-91bb-cc4f63aa5921.png)

mlodic:
- [ ] we can just make a conditional statement in the settings.py that if that value is empty the application just do not run and exit with an error

- [x] we could add an elasticsearch container that should be used only for development (local.yml) and connect the application to it. In this way, it is also possible to perform some manual tests with a working ElasticSearch instance.

mlodic:
other inputs here https://github.com/intelowlproject/IntelOwl/issues/808

devmrfitz:
@mlodic I'd like to work on this issue

mlodic:
solved in #30

PR #24, opened 2022-01-02
Fixes #17: Added API for Enrichment

uzaxirr:
# Description

Added Enrichment Endpoint. To get details of an observable my it's name.
Endpoint: `/api/enrichment?query=<observable_name>`

Please ignore the vague changes in `settings.py` regarding env vars. Did it because of #23 I'll revert them when my PR is good to go.

Added Fake data in DB through admin pannel for testing purpose

## Related issues
Fixes and Closes #17 

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality).

# Checklist

- [x] I have read and understood the rules about [how to Contribute](https://intelowl.readthedocs.io/en/latest/Contribute.html) to this project
- [x] The pull request is for the branch `dev`
- [ ] The tests gave 0 errors.
- [ ] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
- [ ] The commits were squashed into a single one (optional, they will be squashed anyway by the maintainer)
  
# Screenshots

## API Response
Screenshot from 2022-01-02 23-24-14

## Details of the searched observable in DB
Screenshot from 2022-01-02 23-24-40

## All Records in DB
Screenshot from 2022-01-02 23-24-27

PR #25, opened 2022-01-05
Fixes #17: Added API for Enrichment

### LINKED ISSUE #17: create new API endpoint for enrichment (related to #12)
related to #12

uzaxirr:
# Description

Added Enrichment Endpoint. To get details of an observable my it's name.
Endpoint: `/api/enrichment?query=<observable_name>`

Please ignore the vague changes in `settings.py` regarding env vars. Did it because of #23 I'll revert them when my PR is good to go.

Added Fake data in DB through admin pannel for testing purpose

## Related issues
Fixes and Closes #17 

## Type of change

# Checklist

- [x] I have read and understood the rules about [how to Contribute](https://intelowl.readthedocs.io/en/latest/Contribute.html) to this project
- [x] The pull request is for the branch `dev`
- [ ] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
  
# Screenshots

### For a record that exist in DB
Screenshot from 2022-01-05 21-41-55

### For a record that does not exist in DB
Screenshot from 2022-01-05 21-42-07

## Details of the searched observable in DB
Screenshot from 2022-01-02 23-24-40

## All Records in DB
Screenshot from 2022-01-02 23-24-27


uzaxirr:
Hey @mlodic 
I was able to make all the changes mentioned in #24 except to resolve the foreign key `related_ioc` and show there a list of the names of the related IOCs instances.

I spent almost an entire day trying things around with this one but all in vain. Can you lease help me out over here.

api/views.py
```diff
@@ -138,3 +141,17 @@ def _formatted_bad_request(format_):
...
+    observable_name = request.query_params.get('query')
+    logger.info(f"Enrichment view requested for: {str(observable_name)}")
+    try:
+        required_object = IOC.objects.get(name=observable_name)
+    except IOC.DoesNotExist:
+        return Response({'found':"false"}, status=status.HTTP_200_OK)
+    serialized = IOCSerializer(required_object)
+    response_data = serialized.data
+    # Adding a extra "found" feild to response to indicate that IOC was found in D
+    response_data['found']='true'
```
mlodic: I would have expected a boolean value here. Why did you use a string?

api/views.py
```diff
@@ -138,3 +141,17 @@ def _formatted_bad_request(format_):
     else:
         # json
         return JsonResponse({}, status=400)
+
+@api_view(['GET'])
+def enrichment_view(request):
+    observable_name = request.query_params.get('query')
+    logger.info(f"Enrichment view requested for: {str(observable_name)}")
+    try:
```
mlodic: there is no validation of the input here. This is dangerous with a SQL database. At least we should add some regexes.

Considering that only IP addresses and domains are accepted we could check this 2 regexes (would be even better to merge them into a single one):

* `regex_ip = r"^(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})$"`
* `regex_domain = r"^[a-zA-Z\d-]{1,60}(\.[a-zA-Z\d-]{1,60})*$"`

mlodic (changes requested):
for now, we could just make the suggested adjustments and merge the PR

api/views.py
uzaxirr: I was not sure what response to return if the given observable does not pass regex checks so returned "Invalid Observable" please resolve this conversation if the recent changes looks satisfactory to you.

uzaxirr:
If everything seems fine. Tell me ill revert the changes in `settings.py`. Not doing this now beacuse if any changes are requested then i have to make those changes again

api/views.py
```diff
@@ -145,13 +147,17 @@ def _formatted_bad_request(format_):
...
     observable_name = request.query_params.get('query')
+    if not re.match(regex_domain, observable_name):
+        return Response({"message":"Invalid Observable"}, status=status.HTTP_400_BAD_REQUEST)
+    if not re.match(regex_ip, observable_name):
+        return Response({"message":"Invalid Observable"}, status=status.HTTP_400_BAD_REQUEST)
     logger.info(f"Enrichment view requested for: {str(observable_name)}")
     try:
         required_object = IOC.objects.get(name=observable_name)
     except IOC.DoesNotExist:
         return Response({'found':"false"}, status=status.HTTP_200_OK)
```
mlodic: also this must be changed...

api/views.py
```diff
@@ -20,7 +21,8 @@
 from greedybear.serializers import IOCSerializer
 
 logger = logging.getLogger(__name__)
-
+regex_domain = r"^[a-zA-Z\d-]{1,60}(\.[a-zA-Z\d-]{1,60})*$"
+regex_ip = r"^(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})$"
```
mlodic: constants should be put into the consts.py file

api/views.py
```diff
@@ -145,13 +147,17 @@ def _formatted_bad_request(format_):
 @api_view(['GET'])
```
mlodic: please use the constant GET from consts.py

api/views.py
```diff
@@ -145,13 +147,17 @@ def _formatted_bad_request(format_):
...
+    if not re.match(regex_domain, observable_name):
+        return Response({"message":"Invalid Observable"}, status=status.HTTP_400_BAD_REQUEST)
+    if not re.match(regex_ip, observable_name):
+        return Response({"message":"Invalid Observable"}, status=status.HTTP_400_BAD_REQUEST)
     logger.info(f"Enrichment view requested for: {str(observable_name)}")
     try:
         required_object = IOC.objects.get(name=observable_name)
     except IOC.DoesNotExist:
         return Response({'found':"false"}, status=status.HTTP_200_OK)
     serialized = IOCSerializer(required_object)
```
mlodic: why is the serializers.py module in `greedybear` instead of the `api` app?

api/views.py
```diff
@@ -145,13 +147,17 @@ def _formatted_bad_request(format_):
 @api_view(['GET'])
 def enrichment_view(request):
     observable_name = request.query_params.get('query')
+    if not re.match(regex_domain, observable_name):
+        return Response({"message":"Invalid Observable"}, status=status.HTTP_400_BAD_REQUEST)
+    if not re.match(regex_ip, observable_name):
+        return Response({"message":"Invalid Observable"}, status=status.HTTP_400_BAD_REQUEST)
```
mlodic: the correct way is to use the serializers for these checks too.

The serializers checks the data in input and then provides the data to be returned

Please see how we have done it in [IntelOwl here](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/api.py#L54)

greedybear/settings.py
```diff
@@ -13,11 +13,12 @@
 SECRET_KEY = os.environ.get("DJANGO_SECRET", None) or get_random_secret_key()
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.environ.get("DEBUG", False) == "True"
+# DEBUG = os.environ.get("DEBUG", False) == "True"
```
mlodic: there is no need to comment this during development. Use the environment variables in the `env_file`. We cannot just keep these changes until the PR is resolved and then revert them at the end. It is not practical and it is error prone.

This is valid for all the other changes you made in this file.

greedybear/settings.py
```diff
@@ -13,11 +13,12 @@
...
+DEBUG=False
 
-DJANGO_LOG_DIRECTORY = "/var/log/greedybear/django"
+DJANGO_LOG_DIRECTORY = "logrotate"
```
mlodic: why this?

mlodic (changes requested):

api/views.py
uzaxirr: done. but i can't understand the intuition behind storing it as a const in other file. i mean the variable name and its value is equal so what benefit it will provide

mlodic: strings constant are error prone. If you make a little mistake, you won't find it until it's too late. It is cleaner to avoid to write the same string more than once if possible.

Putting them in a clean file, where they can also be shared and imported from, is just sintactic sugar and, even if with a litlle project like this the benefit is almost 0, in the long term you can save possible mistakes + you have cleaner and better readable code.

greedybear/settings.py
uzaxirr: As ik, I had to do this due to #23  thus needed to set it without docker-compose.
Inconvenience regretted

greedybear/settings.py
uzaxirr: That's my laptop specific please ignore

api/views.py
uzaxirr: Wow never thought this way. Really learning a lot by contributing to these projects.

greedybear/settings.py
mlodic: The ElasticSearch variable was the only one who needed a one line change to have it work properly. The others do not need any change. We can just add a conditional statement for that variable as mentioned in the issue you linked.

uzaxirr: i did it while setting that thing up. Reverted now would take care of such things in future

eshaan7:
>  Tell me ill revert the changes in settings.py. Not doing this now beacuse if any changes are requested then i have to make those changes again

There's better ways to handle this than pushing unnecessary changes in the PR. Not only this makes your PR look bad, it also wastes the maintainer's time and breaks the testing suite.  

Fun fact: All developers have changes in their local development environment that they do not wish to push to the remote repository. You can simply just not include those files in `git add` before making a commit or use `git stash` to save these changes and use `git stash pop` to recover them.

api/views.py
```diff
@@ -138,3 +143,17 @@ def _formatted_bad_request(format_):
...
+@api_view([GET])
+def enrichment_view(request):
+    observable_name = request.query_params.get('query')
+    logger.info(f"Enrichment view requested for: {str(observable_name)}")
+    try:
+        required_object = IOC.objects.get(name=observable_name)
+    except IOC.DoesNotExist:
+        return Response({'found':"false"}, status=status.HTTP_200_OK)
+    serialized = IOCSerializer(required_object)
+    response_data = serialized.data
```
eshaan7: - missing `serializer.is_valid()` statement.
- should use `serializer.validated_data` to get data.

api/serializers.py
```diff
@@ -0,0 +1,30 @@
...
+            'scanner', 
+            'payload_request', 
+            'related_ioc',
+        )
+    
+    def regex_validate(self, data):
+        observable = data['observable_name']
+        if not re.match(REGEX_DOMAIN, observable):
+            raise serializers.ValidationError('Invalid Observable')
+        if not re.match(REGEX_IP, observable):
```
eshaan7: these 2 `if` condtionals should have an `OR` operator between them because the observable will either be IP or domain.

greedybear/settings.py
```diff
@@ -37,6 +37,7 @@
     "gui.apps.GuiConfig",
     "api.apps.ApiConfig",
     "greedybear.apps.GreedyBearConfig",
+    'rest_framework',
```
eshaan7: just for cleanliness, this should come before our custom apps like api and greedybear.

api/views.py
```diff
@@ -138,3 +143,17 @@ def _formatted_bad_request(format_):
     else:
         # json
         return JsonResponse({}, status=400)
+
+@api_view([GET])
+def enrichment_view(request):
```
eshaan7, line 148: For best practice, you should use 2 serializers for this view: `IoCSerializer` and `EnrichmentSerializer`.
- The `EnrichmentSerializer` should have 3 fields: 
    - `query` as a CharField with required, write_only attributes set to True.
    - `found = BooleanField(read_only=True, default=False)`. 
    - `ioc = IoCSerializer(read_only=True, default=None)`.
- The `regex_validate` function should be used as the the `validate_query` method in `EnrichmentSerializer`.
- The `IoC.objects.get` check should be inside the `validate` method of `EnrichmentSerializer` and in if/else set appropriately the ioc and found fields.

uzaxirr:
@eshaan7 
Thanks a lot for the insightful info, i've reverted the `settings.py` to its orignal state and would be using the trick mentioned by you in future.

But as of now i've not used 2 serializer as i was getting confused. If its necessary please request changes agin and i shall do the needful.

api/serializers.py
```diff
@@ -0,0 +1,32 @@
+from email.policy import default
```
eshaan7: redundant import

eshaan7:
> But as of now i've not used 2 serializer as i was getting confused. If its necessary please request changes agin and i shall do the needful.

Yes, it is necessary so we can re-use IoCSerializer in the future and also so it's obvious what the role of the EnrichmentSerializer is. We should try our best to follow the "Single Responsibility Principle" design pattern since we are dealing with classes here.

api/serializers.py
```diff
@@ -1,4 +1,6 @@
+from asyncore import write
```
eshaan7: redundant import.

api/serializers.py
```diff
@@ -1,4 +1,6 @@
...
 from email.policy import default
+from http.client import FOUND
```
eshaan7: redundant import.

api/serializers.py
```diff
@@ -21,12 +23,21 @@ class Meta:
             'payload_request', 
             'related_ioc',
         )
-    
+
+class EnrichmentSerializer(IOCSerializer):
```
eshaan7: why are u extending IoCserializer here ? use normal Serializer (`serializers.Serializer`) and specify the `fields` variable in `Meta` class.

api/serializers.py
```diff
@@ -21,12 +23,21 @@ class Meta:
...
+    query = serializers.CharField(max_length=255, write_only=True)
+    found = serializers.BooleanField(read_only=True, default=False)
+    ico = IOCSerializer(read_only=True, default=None)
+
     def validate(self, data):
         """
         Check the given observable against regex expression
         """
         observable = data['observable_name']
```
eshaan7: `query = data["query"]`

api/serializers.py
```diff
@@ -21,12 +23,21 @@ class Meta:
...
         if not re.match(REGEX_IP, observable) or not re.match(REGEX_DOMAIN, observable):
-            raise serializers.ValidationError("Observable must be an IP or a domain")
-        return data
\ No newline at end of file
+            raise serializers.ValidationError("Observable is not a valid IP or domain")
+        try:
+            required_object = IOC.objects.get(name=observable)
+        except IOC.DoesNotExist:
+            raise serializers.ValidationError("Observable does not exist in the database")
```
eshaan7: we do not want to raise exception, we want to mark `found` as False. so replace w/ `data["found"] = False`.

api/serializers.py
```diff
@@ -21,12 +23,21 @@ class Meta:
...
         Check the given observable against regex expression
         """
         observable = data['observable_name']
         if not re.match(REGEX_IP, observable) or not re.match(REGEX_DOMAIN, observable):
-            raise serializers.ValidationError("Observable must be an IP or a domain")
-        return data
\ No newline at end of file
+            raise serializers.ValidationError("Observable is not a valid IP or domain")
+        try:
+            required_object = IOC.objects.get(name=observable)
```
eshaan7, line 30: add `data["found"] = True` after this line.

api/views.py
```diff
@@ -148,10 +148,7 @@ def _formatted_bad_request(format_):
 def enrichment_view(request):
     observable_name = request.query_params.get('query')
     logger.info(f"Enrichment view requested for: {str(observable_name)}")
-    try:
-        required_object = IOC.objects.get(name=observable_name)
-    except IOC.DoesNotExist:
-        return Response({'found':"false"}, status=status.HTTP_200_OK)
+    required_object = EnrichmentSerializer.validate(observable_name)
```
eshaan7: this is not the correct way to call the serializer. Did you try running the code ? `validate` is not a class method so this will raise an error.

```python3
serializer = EnchrichmentSerializer(data=request.query_params, context={"request": request})
serializer.is_valid(raise_exception=True)
return Response(serializer.data)
```

api/serializers.py
eshaan7: and `data["ioc"] = required_object`

api/serializers.py
```diff
@@ -0,0 +1,43 @@
...
+            'cowrie', 
+            'scanner', 
+            'payload_request', 
+            'related_ioc',
+        )
+
+class EnrichmentSerializer(IOCSerializer):
+    query = serializers.CharField(max_length=255, write_only=True)
+    found = serializers.BooleanField(read_only=True, default=False)
+    ico = IOCSerializer(read_only=True, default=None)
```
eshaan7: typo.....should be `ioc`.

api/serializers.py
```diff
@@ -0,0 +1,43 @@
...
+            'last_seen', 
+            'days_seen',
+            'number_of_days_seen',
+            'times_seen', 
+            'log4j', 
+            'cowrie', 
+            'scanner', 
+            'payload_request', 
+            'related_ioc',
+        )
```
eshaan7: don't need to specify these fields explicity, just use `fields = serializers.ALL_FIELDS`.

uzaxirr: As we don't have to expose `related_urls` in the API response i specified the feild which needs to be sent in order to omit the one which we do not want to send.

mlodic: makes sense (`related_urls` could expose info regarding honeypots). Anyway we could explicitly write this down as a comment so we don't forget

eshaan7: > to omit the one which we do not want to send.

You can use the `exclude = ["related_urls",]` attribute for that then.

uzaxirr:
### For a IP that dosen't exist in DB
Screenshot 2022-02-05 at 2 02 14 PM

### For a IP that is present in DB
Screenshot 2022-02-05 at 2 04 09 PM

### For a IP that does not pass Regex check
Screenshot 2022-02-05 at 2 05 18 PM

api/serializers.py
```diff
@@ -0,0 +1,29 @@
...
+    class Meta:
+        model = IOC
+        exclude = ["related_urls",]
+
+class EnrichmentSerializer(serializers.Serializer):
+    found = serializers.BooleanField(read_only=True, default=False)
+    ioc = IOCSerializer(read_only=True, default=None)
+    query = serializers.CharField(max_length=250)
+
+    def validate(self, data):
```
mlodic, line 22: should not be better to just create the custom `validate_query` as Iyou suggested before? @eshaan7 https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

eshaan7: This is okay too since there are other steps in the validation process so the `validate` method is needed anyway.

api/views.py
```diff
@@ -1,5 +1,6 @@
 # This file is a part of GreedyBear https://github.com/honeynet/GreedyBear
 # See the file 'LICENSE' for copying permission.
+import re
```
eshaan7: redundant import.

api/views.py
```diff
@@ -12,9 +13,13 @@
     StreamingHttpResponse,
 )
 from django.views.decorators.http import require_http_methods
-
+from rest_framework.decorators import api_view
+from rest_framework import status
+from rest_framework.response import Response 
 from greedybear.consts import FEEDS_LICENSE, GET, PAYLOAD_REQUEST, SCANNER
 from greedybear.models import IOC
+from api.serializers import IOCSerializer, EnrichmentSerializer
```
eshaan7: redundant import of `IOCSerializer`.

eshaan7:
Thank you!

Issue #26, opened 2022-01-13
use DRF for views and django-rest-durin for authentication


eshaan7:
Approved by @mlodic.

Use how these packages are integrated in IntelOwl for reference.

yaswanthsaivendra:
can I work on this issue?

ManishShah120:
Hi, @eshaan7 Can I please work on this, I want to learn more about this project, and make some useful patches in the coming future.

--Thank You 😄

ManishShah120:
Thanx for assigning. On it

uzaxirr:
@ManishShah120 you still working on this?

mlodic:
A week has passed, I am removing the assignment tomorrow 

See: https://intelowl.readthedocs.io/en/latest/Contribute.html

ManishShah120:
Okay, @mlodic I won't be able to open a PR, by tomorrow.

> @ManishShah120 you still working on this?

I was trying to get started with it, but it seems it is taking too long for me. I am unassigning myself.

uzaxirr:
@mlodic 
By "Use DRF for Views" i assume you mean to convert the existing `feeds` view to a DRF based view and add authentication in the APIs correct?
is there anything i'm missing?

mlodic:
yeah, but remember that you still have an open PR to finish before taking up such an issue

Yshaq:
Hey, can I pick up this issue?

uzaxirr:
Hey, Can I work on this now

uzaxirr:
Hey I've done the Converting views to DRF views part. Just to make sure that we are on same page regarding Rest-Durin i wanna confirm that you want to add configurations for Durin and Create the following endpoints
- `/api/auth/login`
- `/api/auth/logout`
- `/api/auth/logoutall`
- `/api/auth/sessions`
- `/api/auth/apiaccess`
Am i correct or missing something?

In IntelOwl i see only `/api/auth/login` and `/api/auth/logout`

mlodic:
please refer to the branch `dev-v4` there: https://github.com/intelowlproject/IntelOwl/tree/dev-v4/api_app/auth

mlodic:
closing this issue for now cause solved with simple TokenAuthentication

PR #29, opened 2022-02-02
updated feeds  view to make use of DRF and added durin authenication

yaswanthsaivendra:
# Description

- Made changes to feeds View to make use of DRF 
- Added token authentication of django-rest-durin.

## Related issues

This PR solves #26 issue.

## Type of change

# Checklist

- [ ]  I have read and understood the rules about [how to Contribute](https://intelowl.readthedocs.io/en/latest/Contribute.html) to this project
- [ ] The pull request is for the branch `dev`
- [ ] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.


greedybear/settings.py
```diff
@@ -69,6 +75,35 @@
...
+    ],
+    "DEFAULT_PERMISSION_CLASSES": ["rest_framework.permissions.IsAuthenticated"],
+    "EXCEPTION_HANDLER": "rest_framework.views.exception_handler",
+    "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema",
+}
+
+# DRF Spectacular
+SPECTACULAR_SETTINGS = {
+    "TITLE": "GreedyBear API specification",
+    "VERSION": "0.1.3",
```
mlodic: this version would be difficult yo keep updated. Would be better to keep it in a separate file and then write in the release template (see how we do it in IntelOwl [here](https://github.com/intelowlproject/IntelOwl/blob/master/.github/release_template.md) to remember to update it before each new release

requirements.txt
```diff
@@ -3,4 +3,8 @@ elasticsearch-dsl~=7.4.0
 Django==3.2.10
 uWSGI~=2.0.20
 psycopg2-binary~=2.9.2
-slack-sdk~=3.12.0
\ No newline at end of file
+slack-sdk~=3.12.0
+django-rest-durin==1.0.0
+djangorestframework==3.13.1
+drf-spectacular==0.21.2
+django-cache-memoize==0.1.9
```
mlodic: ~= better to get patches without having to change the versions manually

mlodic (commented):
I have just fastly pushed some comments. When your PR is ready, please explicitly ask for a review. Thanks!

greedybear/settings.py
yaswanthsaivendra: I am not sure what would be some other great place to keep that version , so rather placing it in a seperate file I left it there and referenced it `release_template.md` as it is done in Intelowl.

yaswanthsaivendra:
> I have just fastly pushed some comments. When your PR is ready, please explicitly ask for a review. Thanks!

@mlodic I made the changes u have mentioned. can u review the pr again.

greedybear/settings.py
mlodic: a simple `version.py` file so the next time we would need to use that variable in other parts of the code we won't forget that we have already added a version in this file too

greedybear/settings.py
```diff
@@ -34,6 +35,11 @@
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "django.contrib.postgres",
+    # DRF
+    "rest_framework",
+    "durin",
+    "drf_spectacular",
+    # greedybear apps
     "gui.apps.GuiConfig",
```
mlodic: even if right now the Gui application is empty, this should not be dependant from the rest framework so it should be stated _before_ it

api/auth.py
```diff
@@ -0,0 +1,58 @@
...
+    target_class = "durin.auth.CachedTokenAuthentication"
+    name = "durinAuth"
+
+    def get_security_definition(self, auto_schema):
+        return {
+            "type": "apiKey",
+            "in": "header",
+            "name": "Authorization",
+            "description": "Token-based authentication with required prefix: Token",
+        }
```
mlodic, line 59: this was copy-pasted from IntelOwl, am I right? In that case, we cannot say that this file is part of GreedyBear 😅  but we should explicitly say where we took this from with a link to the original code. Otherwise, why did we write that disclaimer in IntelOwl? 😅

api/admin.py
```diff
@@ -0,0 +1,47 @@
...
+
+    exclude = []
+    raw_id_fields = ("user",)
+    readonly_fields = ("token", "expiry", "created", "expires_in")
+
+    def get_fieldsets(self, request, obj=None):
+        if not obj:
+            return [
+                (
+                    "Create token for Greedybear",
```
mlodic, line 24: also this is copy-pasted from IntelOwl but this is a customization made for the PyIntelOwl client and, right now, there is no GreedyBear client.

Even if we could still add something similar for an eventual future PyGreedyBear:
* you badly copy-pasted...you removed `admin.site.unregister(Client)`
* you missed a part of the copy-paste, a critical migration where the clients are created

mlodic (changes requested):
I am honest. We do not need copy-pasted contributions only.

api/auth.py
yaswanthsaivendra: Yaa I accept its a complete copy paste. I have gone through the complete documentation of durin . After getting an understanding of it. I saw the implementation of it in intel owl (as it mentioned to take it as reference) and it was completely the same code as it in examples of documentation expect somewhat good print statements and the schema at the bottom. Then I thought what would be a great thing than copy paste if its completely working fine in our case.

I am not clear of what all views I have to include in for the authentication(as there were 4 token based and 2 session based views in durin) but again Intelowl is the reference for me. so I thought why do I even add other views when they are enough.

api/admin.py
yaswanthsaivendra: Yes , I am not clear about the code here(the methods used in the class), all the know is to register the authtoken , so that they can be view them via admin portal. 
Its my mistake, I should be asking u quesn regarding this rather than doing this.
Can u just give an idea of what changes I have to make?

yaswanthsaivendra:
> I am honest. We do not need copy-pasted contributions only.

Sorry, I didn't mean to....As I'm not sure about what things I have to achieve, I ended up doing it . 😞  I assure, I won't repeat it again

api/auth.py
mlodic: As mentioned, there is a disclaimer and there are licenses. You cannot just copy-paste only because you "thought that would be a great thing than copy paste if its completely working fine in our case." That is not how open source works

api/admin.py
mlodic: "Use how these packages are integrated in IntelOwl for reference" (see #26) does not mean that you just copy-paste without even knowing what you are doing. These kind of contributions:
* waste time of maintainers
* waste time of contributors
* they bring NO value at all.

We leave these issues open because we want to find new possible contributors and this is a way to evaluate them. Otherwise, we would have just done this task on our own very fast.

We don't want people who just go at the end of the goal as fast as possible without even understanding what they are doing.

mlodic:
closed because entirely copy-pasted from IntelOwl and violating license of IntelOwl

PR #30, opened 2022-02-02
Dev

### LINKED ISSUE #23: Elasticsearch installation error

devmrfitz:
# Description

Added elasticsearch container for development

## Related issues
closes #23 

## Type of change

# Checklist

- [X] I have read and understood the rules about [how to Contribute](https://intelowl.readthedocs.io/en/latest/Contribute.html) to this project
- [X] The pull request is for the branch `dev`
- [X] The tests gave 0 errors.
- [X] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
  
### Important Rules

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,3 +84,9 @@ docker-compose -p greedybear up
 # shut down the application
 docker-compose -p greedybear down
 ```
+Note that GreedyBear *needs* a running instance of ElasticSearch of a TPoT to function.
+If you don't have one, you can make the following changes to make GreeyBear spin up it's own ElasticSearch and Kibana instances 
+(...Care! This option would require enough RAM to run the additional containers. Suggested is >=16GB):
+
+1) In ```docker/env_file```, set the variable ```ELASTIC_ENDPOINT``` to ```http://elasticsearch:9200```.
+2) Add ```:docker/elasticsearch.yml``` to the last defined ```COMPOSE_FILE``` variable or uncomment the ```# local development with elasticsearch container``` block in ```.env``` file.
--- a/docker/elasticsearch.yml
+++ b/docker/elasticsearch.yml
@@ -1,0 +1,21 @@
+version: '3.8'
+
+services:
+  uwsgi:
+    depends_on:
+      - elasticsearch
+
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.17.0
+    environment:
+      - "discovery.type=single-node"
+
+  kibana:
+    image: docker.elastic.co/kibana/kibana:7.17.0
+    environment:
+      ELASTICSEARCH_HOSTS: '["http://elasticsearch:9200"]'
+    ports:
+      - '5601:5601'
+    depends_on:
+      - elasticsearch
+
--- a/.env_template
+++ b/.env_template
@@ -3,3 +3,6 @@ COMPOSE_FILE=docker/default.yml
 
 # local development
 COMPOSE_FILE=docker/default.yml:docker/local.yml
+
+# local development with elasticsearch container
+#COMPOSE_FILE=docker/default.yml:docker/local.yml:docker/elasticsearch.yml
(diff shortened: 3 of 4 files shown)


devmrfitz:
@mlodic Please review this PR

mlodic (changes requested):
I won't put that default in the `ELASTIC_ENDPOINT` because that would also be inherited by new production deployments but that is only for development so it should not be a default. On the contrary, the right documentation should describe this case and how to handle it.

Then, ok the `elasticsearch` instance but there is something else to consider that I did not mention in the #23.
First, we would need `kibana` too, otherwise it would be really difficult to perform manual tests on ElasticSearch. Second, this would be anyway an optional configuration that should be in a new additional compose file: this is because if I am running a local instance I could still want to connect to a real T-Pot instance and, in that case, this container is not required

devmrfitz:
@mlodic Please have a look now

docker/default.yml
```diff
@@ -27,6 +27,7 @@ services:
       - env_file
     depends_on:
       - postgres
+      - elasticsearch
```
mlodic: this line should be added as a dependency only in the docker/elasticsearch.yml file. It is just a matter of adding uwsgi again with just this line

README.md
```diff
@@ -83,4 +83,9 @@ docker-compose -p greedybear up
 
 # shut down the application
 docker-compose -p greedybear down
-```
\ No newline at end of file
+```
+Note that GreedyBear *needs* a running instance of ElasticSearch to function.
+If you don't have one, you can make the following changes to make GreeyBear spin up it's own ElasticSearch and Kibana instances:
+
+1) In ```docker/env_file```, set the variable ```ELASTICSEARCH_HOST``` to ```http://elasticsearch:9200```.
```
mlodic: the name is different in that file `ELASTIC_ENDPOINT`

README.md
```diff
@@ -83,4 +83,9 @@ docker-compose -p greedybear up
 
 # shut down the application
 docker-compose -p greedybear down
-```
\ No newline at end of file
+```
+Note that GreedyBear *needs* a running instance of ElasticSearch to function.
```
mlodic: "Note that GreedyBear *needs* a running instance of ElasticSearch to function. This should be the ElasticSearch of a TPoT."

README.md
```diff
@@ -83,4 +83,9 @@ docker-compose -p greedybear up
...
+If you don't have one, you can make the following changes to make GreeyBear spin up it's own ElasticSearch and Kibana instances:
+
```
mlodic, line 90: "...Care! This option would require enough RAM to run the additional containers. Suggested is >=16GB."

mlodic (commented):
👍🏻

devmrfitz:
@mlodic Fixed all the mentioned issues

docker/elasticsearch.yml
```diff
@@ -0,0 +1,37 @@
+version: '3.8'
+
+services:
+  uwsgi:
```
mlodic, line 4: it works as a "override" so no need to replicate what already is in the `default.yml`

just uwsgi, depend_on and elasticsearch are enough

devmrfitz: Do I need to mention ```postgres``` in ```depends_on``` again too?

devmrfitz:
@mlodic resolved

mlodic (approved):

Issue #16, opened 2022-01-01
Backup for the PostgreSQL database


mlodic:

0x0elliot:
Matteo, Would you mind elaborating further on this issue? I want to see if this is a issue I can help fix

mlodic:
It would be cool to have a script that can be launched to perform a backup of the PostgreSQL database in a determined folder in the system that is accessible from the host (not only from the container). In this way, people can then choose where to move those backups files (in a safe place they decide. This is not part of this issue)

Considering that the most important part of this project are the data collected **over time**, it is super important to provide a way to configure automatic backups because if cannot lose the database. (This is extremely different from IntelOwl where the data extracted loses quality over time so we automatically clean up analysis to maintain a clean and little db.)

It could make sense to have a cronjob that runs periodically. It should be configurable:
* the user should choose whether this cron has to be run or not 
* how frequently (every X days)

uzaxirr:
Hey I would like to take up this one.

mlodic:
you already have a PR in progress so no

0x0elliot:
@uzaxirr if you like, then you can take up this issue right after you're done with your PRs and I wouldn't pick this one up until then. Best of luck! 😄

uzaxirr:
> @uzaxirr if you like, then you can take up this issue right after you're done with your PRs and I wouldn't pick this one up until then. Best of luck! 😄

Haha Thanks, i got assigned somewhat similar work at my internship so i believe i could do it easily

Shreyaadgitm:
Hello sir , I would like to take this issue

Shreyaadgitm:
Hello ,  I am new here . Please tell how to get started , how can I make my first contribution  ?

CypherpunkSamurai:
Hello Sir, I wanted to start contributing to this project and this issue seemed interesting.
I'm currently learning about databases and how to use PostgreSQL with psycopg2, so for gaining a little more experience I would like to work on this issue.

**Things I've Currently Done to Get Acquainted with the project:**
I have currently finished looking through the readme and installation docs. I've tried installing and running the project myself, and have looked through the REST API code, and docker sources.

One of the doubts I'm facing is, I can't figure out how to properly run test the server. The server returns HTTP errors when accessing the rest endpoints. Do we need to provide/simulate a fake TPoT instance for the dev environment?

uzaxirr:
[...]

PR #44, opened 2022-03-17
Add db_backup cronjob

CypherpunkSamurai:
# Description

Added db_backup cronjob to backup the db to local file storage.

## Related issues
#16

## Type of change

# Checklist

- [X] I have read and understood the rules about [how to Contribute](https://intelowl.readthedocs.io/en/latest/Contribute.html) to this project
- [X] The pull request is for the branch `develop`
- [X] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
  
### Important Rules

### DIFF
--- a/greedybear/celery.py
+++ b/greedybear/celery.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import os
+from datetime import timedelta
 
 from celery import Celery
 from celery.schedules import crontab
@@ -84,6 +85,12 @@ app.conf.beat_schedule = {
     "monitor_logs": {
         "task": "greedybear.tasks.monitor_logs",
         "schedule": crontab(minute=33),
+        "options": {"queue": "default"},
+    },
+    # run once every settings.DB_BACKUP_FREQUENCY * 24 hour
+    "db_backup": {
+        "task": "greedybear.tasks.db_backup",
+        "schedule": timedelta(days=settings.DB_BACKUP_FREQUENCY),
         "options": {"queue": "default"},
     },
 }
--- a/greedybear/cronjobs/db_backup.py
+++ b/greedybear/cronjobs/db_backup.py
@@ -1,0 +1,47 @@
+import os
+from datetime import datetime
+
+from cStringIO import StringIO
+from django.core import management
+
+from greedybear import settings
+from greedybear.apps import GreedyBearConfig
+from greedybear.cronjobs.base import Cronjob
+
+
+class DbBackup(Cronjob):
+    def __init__(self):
+        super().__init__()
+        self.interval = settings.DB_BACKUP_FREQUENCY
+        self.location = settings.DB_BACKUP_LOCATION
+        self.max_backups = settings.DB_BACKUP_MAX_BACKUPS
+
+    @property
+    def minutes_back_to_lookup(self):
+        return self.interval
+
+    def _remove_oldest_backup(self):
+        # removes files older than self.max_backups
+        self.log.info("[db_backup] checking for older backups...")
+        files = os.listdir(self.location)
+        if len(files) >= self.max_backups:
+            self.log.info("[db_backup] removing old backups...")
+            oldest_file = min(files, key=os.path.getctime)
+            os.remove(os.path.abspath(oldest_file))
+
+    def run(self):
+        self.log.info("[db_backup] starting cron job...")
+        self._remove_oldest_backup()
+        app_name = GreedyBearConfig.name
+        backup_file = (
+            f"{self.location}/db_backup_{datetime.now().strftime('%Y%m%d-%H%M%S')}.json"
+        )
+        # Init a new buffer
+        buf = StringIO()
+        # Call command with and pipe stdout to buffer
+        management.call_command("dbbackup", app_name, stdout=buf)
+        buf.seek(0)
+        # Write buffer to file
+        self.log.info(f"[db_backup] writing backup db to {backup_file}...")
+        with open(backup_file, "w", "utf-8") as dump:
+            dump.write(buf.read())
--- a/greedybear/settings.py
+++ b/greedybear/settings.py
@@ -15,6 +15,13 @@ SECRET_KEY = os.environ.get("DJANGO_SECRET", None) or get_random_secret_key()
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DEBUG", False) == "True"
 
+# DB Backup Config
+# Backup Interval in days. Default: 1 day.
+DB_BACKUP_FREQUENCY = int(os.getenv("DB_BACKUP_FREQUENCY", 1))
+# Backup Storage Location and Max Backups. (When Max Backup is reached old backups are removed).
+DB_BACKUP_LOCATION = os.getenv("DB_BACKUP_LOCATION", "/opt/backups")
+DB_BACKUP_MAX_BACKUPS = os.getenv("DB_BACKUP_MAX_BACKUPS", 10)
+
 DJANGO_LOG_DIRECTORY = "/var/log/greedybear/django"
 MOCK_CONNECTIONS = os.environ.get("MOCK_CONNECTIONS", False) == "True"
 ELASTIC_ENDPOINT = os.getenv("ELASTIC_ENDPOINT", "")
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,8 +4,9 @@ ENV PYTHONUNBUFFERED 1
 ENV DJANGO_SETTINGS_MODULE greedybear.settings
 ENV PYTHONPATH /opt/deploy/greedybear
 ENV LOG_PATH /var/log/greedybear
+ENV DB_BACKUP_LOCATION /opt/backups
 
-RUN mkdir -p ${LOG_PATH} \
+RUN mkdir -p ${LOG_PATH} ${DB_BACKUP_LOCATION} \
     ${LOG_PATH}/django \
     ${LOG_PATH}/uwsgi
 
--- a/greedybear/tasks.py
+++ b/greedybear/tasks.py
@@ -38,3 +38,10 @@ def monitor_logs():
     from greedybear.cronjobs.monitor_logs import MonitorLogs
 
     MonitorLogs().execute()
+
+
+@shared_task()
+def db_backup():
+    from greedybear.cronjobs.db_backup import DbBackup
+
+    DbBackup().execute()


greedybear/settings.py
```diff
@@ -13,7 +13,14 @@
 SECRET_KEY = os.environ.get("DJANGO_SECRET", None) or get_random_secret_key()
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.environ.get("DEBUG", False) == "True"
+# DEBUG = os.environ.get("DEBUG", False) == "True"
+DEBUG = True
```
uzaxirr: Kindly revert this change

greedybear/settings.py
```diff
@@ -13,7 +13,14 @@
...
+
+# DB Backup Configs
+DBBACKUP_STORAGE = "django.core.files.storage.FileSystemStorage"
+DBBACKUP_STORAGE_OPTIONS = {"location": "/var/log/greedybear/"}
+# Database Backup Interval (in minutes). default 120 mins = 2 hours.
+DBBACKUP_FREQUENCY = int(os.environ.get("DBBACKUP_FREQUENCY", 120)) == 120
```
uzaxirr: Not sure if it matters much but instead of `DBBACKUP_STORAGE` you can use `DB_BACKUP_STORAGE` and similarly in other variables as well this looks much readable

uzaxirr (changes requested):

greedybear/settings.py
CypherpunkSamurai: Ah, yes. I was honestly thinking about it while naming it. Sure, will do.

CypherpunkSamurai: Oh, I just remembered. The module depends on those settings [\[ref\]](https://django-dbbackup.readthedocs.io/en/master/installation.html#add-it-in-your-project), so I don't think its possible to rename them (except `DBBACKUP_FREQUENCY`, which I manually added)

uzaxirr: Ohh i see, that's their default name. It's fine then.

greedybear/settings.py
CypherpunkSamurai: Oops. Really Sorry, forgot about it.

greedybear/settings.py
CypherpunkSamurai: Should I change `DBBACKUP_FREQUENCY` or should I keep it as is to represent common configs?

uzaxirr: I recommend not to change in order to keep all the DB Backup related vars in a uniform manner.

CypherpunkSamurai: Ok Sir, Got it 👍🏼

greedybear/settings.py
```diff
@@ -15,6 +15,12 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DEBUG", False) == "True"
 
+# DB Backup Configs
+DBBACKUP_STORAGE = "django.core.files.storage.FileSystemStorage"
+DBBACKUP_STORAGE_OPTIONS = {"location": "/var/log/greedybear/"}
+# Database Backup Interval (in minutes). default 120 mins = 2 hours.
+DBBACKUP_FREQUENCY = int(os.environ.get("DBBACKUP_FREQUENCY", 120)) or 120
```
mlodic: imo once a day it is enough

greedybear/settings.py
```diff
@@ -15,6 +15,12 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DEBUG", False) == "True"
 
+# DB Backup Configs
+DBBACKUP_STORAGE = "django.core.files.storage.FileSystemStorage"
+DBBACKUP_STORAGE_OPTIONS = {"location": "/var/log/greedybear/"}
```
mlodic: these are not logs, I would leverage a new directory like `/opt/backups`

greedybear/settings.py
```diff
@@ -56,6 +62,8 @@
     "gui.apps.GuiConfig",
     "api.apps.ApiConfig",
     "greedybear.apps.GreedyBearConfig",
+    # db backups
+    "dbbackup",
```
mlodic: can you explain advantages of this external app when we have https://docs.djangoproject.com/en/4.0/ref/django-admin/#dumpdata as default in Django?

greedybear/cronjobs/db_backup.py
```diff
@@ -0,0 +1,17 @@
+from django.core import management
+
+from greedybear import settings
+from greedybear.cronjobs.base import Cronjob
+
+
+class DbBackup(Cronjob):
```
mlodic, line 12: without adding this cron as a celery task, it would never run

greedybear/settings.py
mlodic: we also need to provide a mechanism for deletion of too old files in this directory.

I would say a maximum of 10 backups we keep there as default and then we delete the old ones

mlodic (changes requested):

greedybear/settings.py
CypherpunkSamurai: Oh, I haven't looked into that yet. That seems like a better alternative than adding another dependency. I'll try implementing it with `dumpdata` instead 👍🏼

greedybear/cronjobs/db_backup.py
CypherpunkSamurai: Ok Sir.

greedybear/settings.py
CypherpunkSamurai: I've fixed that with the recent commit. Please check

greedybear/settings.py
CypherpunkSamurai: Hello Sir, I've updated the implementation with `dumpdata`. Please review changes :)

CypherpunkSamurai:
Hello Sir, I've re-implemented it with `dumpdata` please check :)

greedybear/settings.py
```diff
@@ -15,6 +15,12 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DEBUG", False) == "True"
 
+# DB Backup Config
+# Backup Interval in minutes. Default: 1440 min = 1 day.
+DB_BACKUP_FREQUENCY = int(os.environ.get("DB_BACKUP_FREQUENCY", 1440)) or 1440
+# Backup Storage Location and Max Backups. (When Max Backup is reached old backups are removed).
+DB_BACKUP_STORAGE_OPTIONS = {"location": "/opt/backups", "max_backups": 10}
```
mlodic: I think we can have 2 different variables for those, each configurable like the previous one.

greedybear/settings.py
```diff
@@ -15,6 +15,12 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DEBUG", False) == "True"
 
+# DB Backup Config
+# Backup Interval in minutes. Default: 1440 min = 1 day.
+DB_BACKUP_FREQUENCY = int(os.environ.get("DB_BACKUP_FREQUENCY", 1440)) or 1440
```
mlodic: have you tried to customize this value? cause this does not work.

you have to leverage the env file that is loaded by the application

greedybear/cronjobs/db_backup.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _remove_oldest_backup(self, dir):
+        # removes files older than self.max_backups
+        files = os.listdir(dir)
+        if len(files) >= self.max_backups:
+            oldest_file = min(files, key=os.path.getctime)
+            os.remove(os.path.abspath(oldest_file))
+
+    def run(self):
+        self._remove_oldest_backup(self.location)
```
mlodic: no need to pass this value as a variable...

greedybear/cronjobs/db_backup.py
```diff
@@ -0,0 +1,45 @@
...
+            str(settings.DB_BACKUP_STORAGE_OPTIONS["location"]) or "/opt/backups"
+        )
+        self.max_backups = int(settings.DB_BACKUP_STORAGE_OPTIONS["max_backups"]) or 10
+
+    @property
+    def minutes_back_to_lookup(self):
+        return self.interval
+
+    def _remove_oldest_backup(self, dir):
+        # removes files older than self.max_backups
```
mlodic, line 24: add log

greedybear/cronjobs/db_backup.py
```diff
@@ -0,0 +1,45 @@
...
+            f"{self.location}/db_backup_{datetime.now().strftime('%Y%m%d-%H%M%S')}.json"
+        )
+        # Init a new buffer
+        buf = StringIO()
+        # Call command with and pipe stdout to buffer
+        management.call_command("dbbackup", app_name, stdout=buf)
+        buf.seek(0)
+        # Write buffer to file
+        with open(backup_file, "w", "utf-8") as dump:
+            dump.write(buf.read())
```
mlodic, line 47: can you show an example of a successful run of this cron?

plus, there are no logs in all the cron, it is impossible to debug what is happening without them

greedybear/celery.py
```diff
@@ -86,4 +86,10 @@ def setup_loggers(*args, **kwargs):
         "schedule": crontab(minute=33),
         "options": {"queue": "default"},
     },
+    # once a day
```
mlodic: ok but the DB_BACKUP_FREQUENCY varialbe is never used. It will always be executed once a day...

mlodic (changes requested):

greedybear/settings.py
CypherpunkSamurai: Ok Sir.

greedybear/cronjobs/db_backup.py
CypherpunkSamurai: Oh Right we can access it from the method itself 👍🏼

greedybear/cronjobs/db_backup.py
CypherpunkSamurai: Ok Sir.

greedybear/settings.py
CypherpunkSamurai: Ok Sir. I'll try debugging it.

greedybear/celery.py
CypherpunkSamurai: Sir, I've added the backup frequency to the celery.py. It should work now. Please review.

greedybear/celery.py
```diff
@@ -86,4 +88,10 @@ def setup_loggers(*args, **kwargs):
...
+    "db_backup": {
+        "task": "greedybear.tasks.db_backup",
+        "schedule": crontab(minute=DB_BACKUP_FREQUENCY),
```
mlodic: this does not work as you intend. Same syntax as a normal cronjob is required

greedybear/settings.py
```diff
@@ -15,6 +15,13 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DEBUG", False) == "True"
 
+# DB Backup Config
+# Backup Interval in minutes. Default: 1440 min = 1 day.
+DB_BACKUP_FREQUENCY = int(os.getenv("DB_BACKUP_FREQUENCY", 1440))
```
mlodic: it is enough to allow users to just choose to do the download every X days. Then we set static the hour/minute time

greedybear/celery.py
```diff
@@ -10,6 +10,8 @@
 from django.conf import settings
 from kombu import Exchange, Queue
 
+from greedybear.settings import DB_BACKUP_FREQUENCY
```
mlodic: no need, we already import `from django.conf import settings`

mlodic (changes requested):

greedybear/celery.py
CypherpunkSamurai: Ok Sir

greedybear/settings.py
CypherpunkSamurai: Ok Sir. I'll add that

greedybear/celery.py
CypherpunkSamurai: Ok Sir, I've updated it with a manual crontab format. Using: `f"0 0 * * */{settings.DB_BACKUP_FREQUENCY}"`.

mlodic: please check the documentation: https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html?highlight=cron

CypherpunkSamurai: Ok, I've updated it now according to the documentation sir. Please check.

greedybear/celery.py
```diff
@@ -86,4 +86,10 @@ def setup_loggers(*args, **kwargs):
         "schedule": crontab(minute=33),
         "options": {"queue": "default"},
     },
+    # run once every settings.DB_BACKUP_FREQUENCY * 24 hour
+    "db_backup": {
+        "task": "greedybear.tasks.db_backup",
+        "schedule": crontab(minute=0, hour=f'*/{24 * settings.DB_BACKUP_FREQUENCY}'),
```
mlodic: this is still wrong, hour and minute must be fixes, days should change

CypherpunkSamurai: Sir I currently had a doubt about the `days` parameter. Sir If you don't mind me asking, isn't the days parameter for 'day of the month' or 'day of the week'? I looked into a few cronjob syntaxes and it all suggested using either `hours` or `days=*/DAYS_INTERVAL`

mlodic: maybe we could just leave this to do it once every day

CypherpunkSamurai: Ok Sir.

I'm very new to celery and cron jobs, I'll try my best to learn it as soon as possible to be capable of working on Greedybear.

CypherpunkSamurai: Hello Sir, I just noticed that celery schedules support `datetime.timedelta` formats. I replaced the `crontab()` implementation with `timedelta(days=....)` implementation. Is this ok, or should we resort to settings daily cron jobs?

mlodic: where did you get this? have you tried it working?

CypherpunkSamurai: Sir I got that from the celery documentation Available Feilds section. I haven't yet tested it, I'll update soon once I test it.

https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html?highlight=cron#available-fields

CypherpunkSamurai: I tried to test the cron job, but there seems to some problem. Maybe I'm running it incorrectly.

I'm using the following commands for testing:
```
export DEBUG=True
docker-compose -p greedybear build
docker-compose -p greedybear up
```

I'm not sure why but my docker runs dont print any logs from celery cron jobs. I've even tried adding `print()` to current cron job, and other cron jobs, but none of them seem to be running.

mlodic: why don't you just populate the `env_file` inside the `docker` folder with that DEBUG=True value?

anyway the logs should work and should appear in the `greedybear.log` file

mlodic: can you please explain which are your tests? cause you have implemented no tests here

CypherpunkSamurai: I've edited the env file this time and tried looking up the log files. the celery logs aren't present there.

> can you please explain which are your tests? cause you have implemented no tests here

Oh, I was actually referring to changing the celery schedules for all the cron jobs during debug. Just for testing if the cron jobs are working at all. It turns out the cron jobs are not working for some reason, or i'm looking for logs in the wrong place.

mlodic: It is difficult to guess from here why in your env celery does not work while in mine yes without any problems...in every case, there are no tests implemented for this thing. We need them so we can accept the PR. See in the `tests` folder and try to prove that what you did work, without thinking about celery

CypherpunkSamurai: Ok Sir. I'll try it. Also, are there any other ways to check if the celery cron job is running? or maybe manually trigger it?

mlodic: this are just some basic hints:

- You can set up a time that is close to yours if you want to see that the cron works
- you can use `.delay()` from `celery` to have a task called via celery. This excludes that there is a problem in celery but does not exclude that it could be one from the cronjob configuration
- a proper test will exclude bugs in normal launches, even without celery. So this is primary importance

PR #109, opened 2023-02-09
Dashboard: Feeds Types Chart 

### LINKED ISSUE #95: Feeds dashboard does not show new honeypots
image

This should show also other honeypots data added in v1.0.1

carellamartina:
# Description

Adjusted feeds types chart, now it shows all active honeypots

## Related issues
close #95

## Type of change

# Checklist

- [x] I have read and understood the rules about [how to Contribute](https://greedybear.readthedocs.io/en/latest/Contribute.html) to this project
- [x] The pull request is for the branch `develop`
- [x] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://greedybear.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
- [ ] If changes were made to an existing model/serializer/view, the docs were updated and regenerated (check [CONTRIBUTE.md](./Contribute.md)).
  
### Important Rules


frontend/src/components/dashboard/utils/charts.jsx
```diff
@@ -127,10 +126,20 @@ export const FeedsTypesChart = React.memo(() => {
...
+        return Object.entries(data)
+          .slice(1)
+          .map(([dKey], i) => (
+            <Bar
+              stackId="feedtype"
+              key={dKey}
+              dataKey={dKey}
+              fill={colors[i]}
+            />
+          ));
```
drosetti: This is correct, but quite difficult to read:
* `const [data] = respData` I don't get this: i test with some data in jan and some in feb, everything is plotted correctly, but data contains only the jan. How is this possible ? Please add a comment
* slice remove an element from the list in the position passed as arg, it's correct because we don't want to plot the date field, but i suggest to make it clearer. We can delete it from the object before we convert it.
* I suggest to add a comment in the map: we need to use only the types/keys ([key] does this work) and also have the feed type mapped to a color, for this reason we need a index var (i). I suggest to add a comment to this or to find an easier way.

drosetti (changes requested):
Good work!

I have two requests:
* Changes in the charts module to make the code easier to read
* Add a test for the function `getRandomColorsArray`. I know you didn't write it, but we need to add some tests for the existing code and I think the better way is to add little tests while we are working in feature that use it.

drosetti:
Edit: `getRandomColorsArray` is a function from the certego-ui, it's correct to not test it.

drosetti (approved):

PR #397, opened 2024-12-09
Improve API performance. Closes #394

### LINKED ISSUE #394: Feed API slow
Querying the API, e.g. `/api/feeds/all/all/recent.json`, takes a while. Requesting the feed on my GreedyBear instancees with `curl -so /dev/null -w '%{time_total}\n' http://HOSTNAME/api/feeds/all/all/recent.json` takes about 6 seconds. This also produces significant CPU load on the server. However the Honeynet instance seems to be much quicker. Is it just an issue of my installations?

regulartim:
# Description
Two changes are introduced here:

1. By using prefetch_related I could reduce the time the mentioned API request takes from ~6 seconds to ~2 seconds. This change is completly contained in [eda16cd](https://github.com/intelowlproject/GreedyBear/pull/397/commits/eda16cdde738f96302ec88b14c894e04c39cb48f)
2. By implementing custom serialization / validation and removing the serializers from djangos `rest_framework`, the request time could be reduced further to ~200ms. The downside is that some IOC properties (e.g. value, first_seen, ...) are not checked as carefully as before. But since these values come from our DB anyway, I personally think that the performance improvement is worth the inferior validation. 

## Related issues
- #394

## Type of change

# Checklist

### Important Rules


api/views.py
```diff
@@ -214,21 +212,8 @@ def feeds_response(request, iocs, feed_type, format_, dict_only=False):
...
-                "first_seen": ioc.first_seen.strftime("%Y-%m-%d"),
-                "last_seen": ioc.last_seen.strftime("%Y-%m-%d"),
-                "times_seen": ioc.times_seen,
-                "feed_type": ioc_feed_type,
-            }
-
-            serializer_item = FeedsResponseSerializer(data=data_)
-            serializer_item.is_valid(raise_exception=True)
-            json_list.append(serializer_item.data)
+                    ioc_feed_type = str(ioc.general_honeypot.all()[0]).lower()
```
mlodic, line 229: why changing `first` here?

api/serializers.py
```diff
@@ -1,11 +1,15 @@
...
 
-from greedybear.consts import REGEX_DOMAIN, REGEX_IP
+from greedybear.consts import PAYLOAD_REQUEST, REGEX_DOMAIN, REGEX_IP, SCANNER
 from greedybear.models import IOC, GeneralHoneypot
 from rest_framework import serializers
 
 logger = logging.getLogger(__name__)
+VALID_ATTACK_TYPES = {"scanner", "payload_request", "all"}
+VALID_AGES = {"persistent", "recent"}
+VALID_FORMATS = {"csv", "json", "txt"}
```
mlodic: ideally it would be better to have these values separated in a `consts.py` file so other modules can leverage it without importing all the stuff here.

api/serializers.py
```diff
@@ -47,37 +51,45 @@ def validate(self, data):
...
-    times_seen = serializers.IntegerField()
-
-    def validate_feed_type(self, feed_type):
-        logger.debug(f"FeedsResponseSerializer - validation feed_type: '{feed_type}'")
-        return feed_type_validation(feed_type)
+@cache
+def feed_type_validation(feed_type: str) -> bool:
+    general_honeypots = GeneralHoneypot.objects.all().filter(active=True)
+    valid_types = set(["log4j", "cowrie", "all"] + [hp.name.lower() for hp in general_honeypots])
+    return feed_type in valid_types
```
mlodic: caching the query is a great idea, thank you!

However, this would make the output of the `GeneralHoneypot.objects.all().filter(active=True)` always the same, even if an admin changed something in that table. So this should not be cached in this way.

In my opinion, the better solution would be to calculate the items from GeneralHoneypots only once at the start of the available API views (`feeds` and `feeds_pagination`) and then pass the result as a parameter to the `get_queryset` and the `feeds_response` function and then, utlimately, to the Serializer for the usage.

api/serializers.py
```diff
@@ -47,37 +51,45 @@ def validate(self, data):
...
+        SCANNER: ioc.scanner,
+        PAYLOAD_REQUEST: ioc.payload_request,
+        "first_seen": ioc.first_seen.strftime("%Y-%m-%d"),
+        "last_seen": ioc.last_seen.strftime("%Y-%m-%d"),
+        "times_seen": ioc.times_seen,
+        "feed_type": feed_type,
+    }
+    if not ioc_validation(data):
+        raise serializers.ValidationError(f"Unknown error while validating {data}")
+    return data
```
mlodic: even if I am not an admirer of the Django Rest Framework at all, the custom Serializers are way more easy to read and to maintain. I would not change that stuff to simple functions.

api/views.py
regulartim: `first` does not work with prefetching here for some reason. I tried to find out why but did not really find an answer. One explanation might be that `first`tries to order the objects by their ID which is not getting prefetched and therefore needs an additional database query.

api/serializers.py
regulartim: Ok, I'll look into it again. Maybe we can find a way to use the Django Serializers without such a massive performance loss.

api/serializers.py
regulartim: Problem: I am not able to pass the set of valid feed types through the Serializer to the validation function. If I understand the documentation correctly, the validation functions are build to only validate the data that was serialized by the serializer. But I do not serialize the set of valid feed types, so it ca not be passed to the validation function.
Suggestion: Instead of calling the `feed_type_validation` function inside the serializer class, we call it after the instantiation of the serializer in `get_queryset` and  `feeds_response`.  Are you happy with that?

api/serializers.py
regulartim: Unfortunately I have not found a possibility to do that. Since we do not pass the IOC object itself, but a dictionary into the Serializer, validation is mandatory. This seems to be expensive. I would like to add a constant that allows us to skip the serialisation via the Serializer and thus bypass the validation.

api/serializers.py
regulartim: I removed them completely as they are not needed anymore.

api/views.py
mlodic: perfect, so let's just keep it in this way 🤣 
Maybe we could add a comment there cause I am afraid that if anyone else see this, they could change it back to `first`

api/serializers.py
mlodic: yep, I mean, I don't want to force any kind of specific pattern, like the DRF, because I think it is better to find a solution fast than to focus to code perfectionism because that can be always opinionable.

Another option can be using the [object-level validation](https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation) and [extra-context](https://www.django-rest-framework.org/api-guide/serializers/#including-extra-context). In this way you can pass all the data that you want through the serializers and with the more generic validator you can do all the stuff that you want

greedybear/consts.py
```diff
@@ -7,6 +7,7 @@
 POST = "POST"
 
 FEEDS_LICENSE = "https://github.com/honeynet/GreedyBear/blob/main/FEEDS_LICENSE.md"
+SKIP_FEED_VALIDATION = False
```
mlodic: - if you want to dynamically change this without having to re-build the code, you can add it as an environment variable in `env_file_template` and then extract it as a Django constant in `settings.py`

mlodic (approved):

greedybear/consts.py
regulartim: Great idea, thanks! :) Will do that and merge.

api/serializers.py
regulartim: The [extra-context](https://www.django-rest-framework.org/api-guide/serializers/#including-extra-context) hint was what i needed! Thanks a lot! :)

Issue #405, opened 2024-12-11
Add interaction count to the IOC model


regulartim:
The IOC model already has a `times_seen` property that counts on how many extraction processes an IOC was extracted. But that does not represent the actual number of the attacker's interactions with the honeypot. But this information might also be valuable in some scenarios.

regulartim:
To make it easier to distinguish between them, maybe we could rename `times_seen` in something like `attack_count` and name the new property `interaction_count`?

mlodic:
agree, there is also the `number_of_days_seen` parameter that is similar and maybe this could stay the same.

In django, a data migration would be required for such change. At the same time, all the other references in the code should be updated

regulartim:
I need to read up on the data migration subject, but I'll give it a try.

regulartim:
Should we maybe keep the old field name in the API for compatibility?

mlodic:
which API are you referring to?

regulartim:
I'm referring to the feeds API. For example for someone who regularly retrieves `/api/feeds/all/all/recent.json` from some GreedyBear instance, renaming a property of the JSON object may break their process. That's why I thought about keeping the old field name in the current feeds API.

mlodic:
you mean whether they have some client doing computation on that field and expecting it, right? 
Well yeah that's possible but we could also add that as a breaking change and make a new major release.

Another option could be to leave that feed like that as you mentioned and then possibly create new feeds with the additional data that you added.

regulartim:
I think I'd prefer the second option. I will revert renaming the field in the JSON feed in #407 and then create a PR for a feeds API v2 later.

regulartim:
After renaming it back at some places I changed my mind. I think this will become too confusing when the same value is called `times_seen` in half of the places and `attack_count` in the other half. Let's add it as a breaking change in a major version together with a reworked feeds API. Do you agree? Sorry for the back and forth.

mlodic:
np yes let's go through that

Issue #36, opened 2022-02-14
Create different type of feeds based on mathematical models


mlodic:
Right now, there are 2 feeds available based on the `age` of the indicator:

* recent: most recent IOCs seen in the last 3 days
* persistent: these IOCs are the ones that were seen regularly by the honeypots. This feeds will start empty once no prior data was collected and will become bigger over time.

But they are really simple. Maybe we could think about creating different kind of lists based on more reliable values.
This issue will need a complex research analysis.
Starting point could be the Stratosphere's research: https://www.stratosphereips.org/aip-tool

regulartim:
I would like to work on this as part of my Master's thesis. Is there a possibility to talk to you about the topic?

mlodic:
sure, can you contact me in private in the GSOC Honeynet Slack or Twitter?

* [Slack chat](https://gsoc-slack.honeynet.org/)
* [Twitter](https://twitter.com/intel_owl)

regulartim:
The PR #392 closed itself when I renamed the source branch. I will open another PR when I'm ready. Sorry for that.

mlodic:
np

regulartim:
After testing several models I have some results. The best performing ones are Random Forest models. Here is a short comparison between them and the age filters already implemented in GreedyBear:

| Model | Interaction Recall | IP F1 Score | IP Precision | IP Recall |
|---|---|---|---|---|
| Recent (GreedyBear) | 0.5490 | 0.4675 | 0.4212 | 0.5253 |
| Persistent (GreedyBear) | 0.3871 | 0.6063 | 0.5462 | 0.6812 |
| Random Forest Classifier | 0.4532 | **0.6263** | **0.5642** | **0.7037** |
| Random Forest Regressor | **0.8315** | 0.1629 | 0.1468 | 0.1831 |

In simple terms the recall scores describe what fraction of interactions / IPs would have been blocked today if I had used yesterday's feed as a blocklist (assuming a feed size of 5000 IOCs). 

As you can see, the Random Forest Regressor is optimised for predicting the number of interactions that can be expected from an IP address and is clearly superior to all other models in this respect.

The Random Forest Classifier performs only slightly better than the persistent feed. However, it is more flexible as it can be of any size, whereas the persistent feed is limited by the number of IOCs that meet its criteria. 

What do you think? Which of them should we integrate into GreedyBear?

mlodic:
Wonderful analysis! :) 

[...]

PR #419, opened 2025-01-03
New Feeds API. Closes #97 and #98. 

### LINKED ISSUE #97: Filter IP addresses from known scanners
We should periodically download this batch of data: https://raw.githubusercontent.com/stamparm/maltrail/master/trails/static/mass_scanner.txt
and add those IP to whitelists to reduce number of false positives

### LINKED ISSUE #98: Allow to do customized feeds lookups
We could add more ways to extract data feeds from GB other than "recent" and "persistent" which are free.

These new ways must be protected with authentication to avoid abuse.

We could give the users the chance to:
* download the data extracted in the last X hours (customization of "recent")
* download the data that was seen more than X times in the last X days (customization of "persistent")

regulartim:
# Description

This PR introduces a new API endpoint for feeds and adds some data to the existing ones. Changes in detail:

1. There is a new API endpoint for feeds (`/api/feeds/v2`) that provides highly configurable and more detailed feeds. This endpoint is only accessible for authenticated users. 
2. The old API endpoints for feeds now provide some more information about IOCs:
   1. the interaction_count (see #405)
   2. the reputation of an IP according to [listbot](https://github.com/telekom-security/listbot) (see #97) 
   3. the ASN of an IP
   4. the number of attempted logins from an IP 
   5. the number of destination ports that were attacked by an IP
3. The retrieval of valid feed types is now a separate function.

## Related issues
- Closes #97
- Closes #98
- Provides information described in #405 
- Important for #36 

## Type of change

# Checklist

- [ ] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/GreedyBear/Contribute/) to this project.
- [ ] The pull request is for the branch `develop`.
- [ ] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
  
### Important Rules

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -67,12 +67,16 @@ class FeedsSerializer(serializers.Serializer):
 
 class FeedsResponseSerializer(serializers.Serializer):
     feed_type = serializers.CharField(max_length=120)
-    value = serializers.CharField(max_length=120)
+    value = serializers.CharField(max_length=256)
     scanner = serializers.BooleanField()
     payload_request = serializers.BooleanField()
     first_seen = serializers.DateField(format="%Y-%m-%d")
     last_seen = serializers.DateField(format="%Y-%m-%d")
-    attack_count = serializers.IntegerField()
+    attack_count = serializers.IntegerField(min_value=1)
+    interaction_count = serializers.IntegerField(min_value=1)
+    ip_reputation = serializers.CharField(allow_blank=True, max_length=32)
+    asn = serializers.IntegerField(allow_null=True, min_value=1)
+    login_attempts = serializers.IntegerField(min_value=0)
 
     def validate_feed_type(self, feed_type):
         logger.debug(f"FeedsResponseSerializer - validation feed_type: '{feed_type}'")
--- a/api/urls.py
+++ b/api/urls.py
@@ -1,6 +1,6 @@
 # This file is a part of GreedyBear https://github.com/honeynet/GreedyBear
 # See the file 'LICENSE' for copying permission.
-from api.views import StatisticsViewSet, enrichment_view, feeds, feeds_pagination, general_honeypot_list
+from api.views import StatisticsViewSet, enrichment_view, feeds, feeds_pagination, feeds_v2, general_honeypot_list
 from django.urls import include, path
 from rest_framework import routers
 
@@ -11,6 +11,7 @@ router.register(r"statistics", StatisticsViewSet, basename="statistics")
 # These come after /api/..
 urlpatterns = [
     path("feeds/", feeds_pagination),
+    path("feeds/v2/", feeds_v2),
     path("feeds/<str:feed_type>/<str:attack_type>/<str:age>.<str:format_>", feeds),
     path("enrichment", enrichment_view),
     path("general_honeypot", general_honeypot_list),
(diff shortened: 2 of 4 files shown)


regulartim:
@mlodic : One thing I forgot: As the new API is only accessible by authenticated users, should we add the option to exclude mass scanners to the old APIs?

mlodic:
> As the new API is only accessible by authenticated users, should we add the option to exclude mass scanners to the old APIs?

for "mass scanners" you mean legit scanners like Censys, right? if so, yes

api/serializers.py
```diff
@@ -54,25 +56,54 @@ def feed_type_validation(feed_type: str, valid_feed_types: frozenset) -> str:
...
+def ordering_validation(ordering: str) -> str:
+    if not ordering:
+        raise serializers.ValidationError("Invalid ordering: <empty string>")
+    if ordering[0] == "-":
+        ordering = ordering[1:]
+    try:
+        IOC._meta.get_field(ordering)
+    except FieldDoesNotExist as exc:
+        raise serializers.ValidationError(f"Invalid ordering: {ordering}") from exc
+    return ordering
```
mlodic: I am not completely sure to understand what happens here. You remove the `-` if exists and then return the field value without it. But isn't it needed to properly order the query afterwards?

api/serializers.py
```diff
@@ -54,25 +56,54 @@ def feed_type_validation(feed_type: str, valid_feed_types: frozenset) -> str:
     return feed_type
 
 
-class FeedsSerializer(serializers.Serializer):
+@cache
```
mlodic, line 71: is this useful? I think I am missing the point of this function. Could you clarify it?

api/views/utils.py
```diff
@@ -0,0 +1,260 @@
...
+            case "recent":
+                self.max_age = "3"
+                self.min_days_seen = "1"
+                if "feed_type" in self.ordering:
+                    self.ordering = "-last_seen"
+            case "persistent":
+                self.max_age = "14"
+                self.min_days_seen: "10"
+                if "feed_type" in self.ordering:
+                    self.ordering = "-attack_count"
```
mlodic: looking at this closely, this could be all converted inside the Django Rest Framework but don't worry, I won't ask you to change it 🤣 . The important thing is that you don't find any particular perfomance issues, as I said today. I'll check that in the Honeynet production servers too

regulartim: Ok :D If there's the option to do this with a built-in class from Django Rest Framework I would like to use it. Would you show me where to find it?

api/serializers.py
regulartim: You are right, this is indeed confusing. I will make the function clearer.

api/serializers.py
regulartim: The `@cache` decorator? I assume that the function will very often be called with the same `ordering` argument, e.g. `-last_seen` which is the default ordering. If the `@cache` decorator is used and the function has been called at least once with a given argument, the result of the function will be obtained by a dictionary lookup instead of executing the function. This should be much quicker.

api/views/utils.py
regulartim: But I would like to change that in a separate PR.

mlodic: yes no problem. I am finally back

mlodic:
thanks for the patience. I merged this. The refactor about DRF can be done later

PR #481, opened 2025-02-25
Cleanup. Closes #440

### LINKED ISSUE #440: Add IOC and Cowrie Session retention
It makes sense to add some sort of periodic cleanup and customizable retention of the data.

In about 2,5 years, in Honeynet we collected about 500k different IP addresses by using Suricata, Glutton, Cowrie and Log4Pot. That's a managable number but it is increasing.

At the same time, with the 1.4.0 version, we added additional information for Cowrie and that generates about 4k logs in an hour so the retention can be managed more closely for this case.

regulartim:
# Description

Added a scheduled task that automatically removes outdated records based on configurable retention periods. The job cleans up old IOCs, Cowrie sessions, and command sequences to prevent database bloat and improve performance. All retention periods are configurable via environment settings.

## Related issues
- #440 

## Type of change

# Checklist
  
### Important Rules


docker/env_file_template
```diff
@@ -49,4 +49,14 @@ EXTRACTION_INTERVAL=10
 
 # Set True to cluster command sequences recorded by Cowrie once a day
 # This might be computationaly expensive on large Databases
-CLUSTER_COWRIE_COMMAND_SEQUENCES=False
\ No newline at end of file
+CLUSTER_COWRIE_COMMAND_SEQUENCES=False
+
+# Days to keep unseen IOCs before deletion
```
mlodic, line 54: ok I noticed now that you already opened the PR so let's keep the retention management for IOCs and Command Sequences too. I would put a higher value as default (10 years)

regulartim: Two important considerations:
- CowrieSession serves as the n:m relationship bridge between IOC and CommandSequence objects. Deleting a session removes the association between an IOC and the commands it executed during that session.
- With 10 years of CommandSequence data, clustering may become too computationally expensive due to the time complexity of these algorithms.

mlodic: > CowrieSession serves as the n:m relationship bridge between IOC and CommandSequence objects. Deleting a session removes the association between an IOC and the commands it executed during that session.

Because of that, I would keep CowrieSessions with at least one associated command alive for more time.
If the CommandSequence objects are deleted on cascade with CowrieSessions, does it make sense to clean that table too directly? (Maybe I am missing something so please tell me if you see any flaw)

> With 10 years of CommandSequence data, clustering may become too computationally expensive due to the time complexity of these algorithms

Ok so nevermind, I am more interested in keeping a long retention for the IOC itself

regulartim: > Because of that, I would keep CowrieSessions with at least one associated command alive for more time.

Yes. I think it is sensible to keep them at least as long as they have an associated command. 

> If the CommandSequence objects are deleted on cascade with CowrieSessions, does it make sense to clean that table too directly? (Maybe I am missing something so please tell me if you see any flaw)

CommandSequence objects are NOT deleted on cascade. One CommandSequence can be executed by multiple IOCs. If we would cascade delete it with the deletion of a single CowrieSession, that would also delete it for Sessions from other IOCs, which is not what we want. 

Conclusion:
- we set  IOC_RETENTION to 3650 (10y)
- we leave the default for COWRIE_SESSION_RETENTION and COMMAND_SEQUENCE_RETENTION at 365
- we only delete CowrieSessions if they have no associated CommandSequence (either because they never had one or because it got cleaned up)

Do you agree?

mlodic: super!

mlodic (approved):

PR #482, opened 2025-02-25
Update scores after extraction. Closes #462

### LINKED ISSUE #462: Calculate scores after extraction
Currently the scores for the IoCs are only calculated and updated once a day. It would be nice if we could do this every time an IoC is seen during an extraction run.

regulartim:
# Description

Until now the scores for the IoCs are only calculated and updated once a day. With this change, the scores for an IoC are updated every time it is seen during an extraction run.

## Related issues
Closes #462 

## Type of change

# Checklist

- [ ] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/GreedyBear/Contribute/) to this project.
- [ ] The pull request is for the branch `develop`.
- [ ] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
  
### Important Rules

### DIFF
--- a/greedybear/cronjobs/scoring/scoring_jobs.py
+++ b/greedybear/cronjobs/scoring/scoring_jobs.py
@@ -1,6 +1,7 @@
 import json
 import logging
 from collections import defaultdict
+from datetime import date
 
 import pandas as pd
 from django.core.files.base import ContentFile
@@ -8,7 +9,7 @@ from django.core.files.storage import FileSystemStorage
 from django.db.models import F, Q
 from greedybear.cronjobs.base import Cronjob
 from greedybear.cronjobs.scoring.random_forest import RFClassifier, RFRegressor
-from greedybear.cronjobs.scoring.utils import correlated_features, get_current_data, get_features
+from greedybear.cronjobs.scoring.utils import correlated_features, get_current_data, get_data_by_pks, get_features
 from greedybear.models import IOC
 from greedybear.settings import ML_MODEL_DIRECTORY
 
@@ -138,7 +139,7 @@ class UpdateScores(Cronjob):
         super().__init__()
         self.data = None
 
-    def update_db(self, df: pd.DataFrame, score_names: list[str]) -> int:
+    def update_db(self, df: pd.DataFrame) -> None:
         """
         Update IOC scores based on new data from a DataFrame.
 
@@ -149,13 +150,12 @@ class UpdateScores(Cronjob):
         Args:
             df: DataFrame containing new score data.
                 Must have a 'value' column with IOC names/IPs and columns for each score.
-            score_names: List of column names in DataFrame representing scores to update.
-                These must match the field names in the IOC model.
 
         Returns:
             int: The number of objects updated.
         """
         self.log.info("begin updating scores")
+        score_names = [s.score_name for s in SCORERS]
         scores_by_ip = df.set_index("value")[score_names].to_dict("index")
         iocs = IOC.objects.filter(Q(cowrie=True) | Q(log4j=True) | Q(general_honeypot__active=True)).filter(scanner=True).distinct().only("name", *score_names)
         iocs_to_update = []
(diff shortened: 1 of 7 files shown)


greedybear/cronjobs/scoring/scoring_jobs.py
```diff
@@ -182,6 +182,46 @@ def update_db(self, df: pd.DataFrame, score_names: list[str]) -> int:
...
 
+    def update_iocs(self, iocs: set[IOC], df: pd.DataFrame) -> None:
+        """
+        Update scores for a specific list of IoC objects.
+
+        Args:
+            iocs: Set of IoC objects to update
+            df: DataFrame containing new score data
+        """
+        self.log.info("begin updating scores")
```
mlodic: this is the same message of update_db so it can be confused while looking at the logs

greedybear/cronjobs/scoring/scoring_jobs.py
```diff
@@ -182,6 +182,46 @@ def update_db(self, df: pd.DataFrame, score_names: list[str]) -> int:
...
+        Update scores for only the specific IoCs provided.
+
+        Args:
+            iocs: List of IoC objects to update scores for
+
+        Returns:
+            int: Number of objects updated
+        """
+        iocs = set(iocs)
+        primary_keys = set(ioc.pk for ioc in iocs)
```
mlodic, line 207: If I am correct, the complexity (number of iterations) of this can be reduced by just doing a single set, the second one.

greedybear/cronjobs/scoring/scoring_jobs.py
```diff
@@ -182,6 +182,46 @@ def update_db(self, df: pd.DataFrame, score_names: list[str]) -> int:
...
+    def update_iocs(self, iocs: set[IOC], df: pd.DataFrame) -> None:
+        """
+        Update scores for a specific list of IoC objects.
+
+        Args:
+            iocs: Set of IoC objects to update
+            df: DataFrame containing new score data
+        """
+        self.log.info("begin updating scores")
+        score_names = [s.score_name for s in SCORERS]
```
mlodic: I think there are a lot of similarities between this method and with update_db, can you extend the previous one to include this case? For instance, `score_names` is one shared variable and you can avoid duplication of code with some well put if/else clauses

mlodic (changes requested):

greedybear/cronjobs/scoring/scoring_jobs.py
regulartim: Yes, your are right, these functions should be merged.

greedybear/cronjobs/scoring/scoring_jobs.py
regulartim: I would like to keep both. 
- `iocs = set(iocs)` removes duplicate IOCs from the list to avoid duplicate updates later in `update_db()`
- `primary_keys = set(ioc.pk for ioc in iocs)` is needed to fetch additional data from the DB

I don't see a more elegant way to solve this because we will always need to (1) deduplicate the IOC objects and (2) obtain the PKs of all IOCs. I also do not think that this has a huge performance impact as the number of IOCs  here will usually be really small.

greedybear/cronjobs/scoring/scoring_jobs.py
regulartim: I merged the methods so now it's unambiguous.

greedybear/cronjobs/scoring/scoring_jobs.py
mlodic: yes I didn't notice the usefulness of the first set, now I got that. thanks for explaining

mlodic (approved):

PR #487, opened 2025-03-02
single task for train and update

mlodic:


greedybear/tasks.py
```diff
@@ -52,35 +52,18 @@ def monitor_logs():
...
 
     trainer = TrainModels()
     trainer.execute()
-    return trainer.current_data
-
-
-@shared_task()
-def update_scores(current_data=None):
-    from greedybear.cronjobs.scoring.scoring_jobs import UpdateScores
+    current_data = trainer.current_data
```
regulartim: You can remove the variable and directly set `updater.data = trainer.current_data`.

regulartim (approved):

PR #541, opened 2025-07-14
add mass scanner extraction

mlodic:


greedybear/cronjobs/attacks.py
```diff
@@ -98,6 +99,18 @@ def _get_attacker_data(self, honeypot, fields: list) -> list:
             iocs.append(ioc)
         return iocs
 
+    def _get_ip_reputation(self, ip, hit):
+        ip_reputation = hit.get("ip_rep", "")
+        if not ip_reputation:
+            try:
+                MassScanners.objects.get(ip_address=ip)
```
regulartim, line 106: What do you think about creating an index on the `ip_address` column?

regulartim (approved):

greedybear/cronjobs/attacks.py
mlodic: that makes sense, I'll add it

mlodic: added

PR #545, opened 2025-07-17
add mass scanner exclusion as default

### LINKED ISSUE #479: Add option to exclude mass scanners to the Feeds page
We already have this option in the underlying API, so a checkbox on the page would be nice.

mlodic:
only consider this commit https://github.com/intelowlproject/GreedyBear/pull/545/commits/f953887509b695f64c5259405440656857989cfd @regulartim. In this way by default the users of the public instance can get valuable IP addresses without the majority of the noise.

I am merging and deploying even before approval to fasten things up (next week I am away). Still please review so that we can add changes later in case. :) Thanks


api/views/utils.py
```diff
@@ -75,10 +75,11 @@ def __init__(self, query_params: dict):
         self.paginate = query_params.get("paginate", "false").lower()
         self.format = query_params.get("format_", "json").lower()
         self.feed_type_sorting = None
-
-    def exclude_mass_scanners(self):
         self.exclude_reputation.append("mass scanner")
```
regulartim, line 78: Just a minor thing: I think I would move this line directly below `self.exclude_reputation = query_params["exclude_reputation"].split(";") if "exclude_reputation" in query_params else []` to make it more obvious what this property actually contains.

regulartim (commented):
Nice! :)

Issue #517, opened 2025-06-02
Publish Payload Delivery Urls to URLhaus


mlodic:
we could do that with the intelowl project dedicated X account

mlodic:
This is the API doc: https://threatfox.abuse.ch/api/#share
Example code: https://github.com/abusech/ThreatFox/blob/main/threatfox_submit-ioc.py



I created an account for the IntelOwlProject to send the data from this platform.
First published IOC: https://threatfox.abuse.ch/ioc/1558075/

mlodic:
30 automatic contributions to Threatfox in the last month: see: https://threatfox.abuse.ch/user/1860/
FYI @regulartim

PR #552, opened 2025-07-18
added threatfox integration for payload request IOCs

mlodic:
solving this: https://github.com/intelowlproject/GreedyBear/issues/517


greedybear/cronjobs/attacks.py
```diff
@@ -66,8 +69,45 @@ def _add_ioc(self, ioc, attack_type: str, general=None):
...
         return ioc_record
 
+    def _threatfox_submission(self, ioc_record: "IOC"):
+        if not settings.THREATFOX_API_KEY:
+            self.log.warning("Threatfox API Key not available")
+            return
+
+        # we submit only payload request IOCs for now cause they are more reliable
+        if not ioc_record.payload_request:
+            return
```
regulartim, line 82: I think it would be better to first check `if not ioc_record.payload_request` and then `not settings.THREATFOX_API_KEY`. This would drastically reduce the amount of log messages, right?

regulartim: Or, alternatively, call `_threatfox_submission` from `_extract_possible_payload_in_messages` in `cowrie.py` to make sure it is only called for payload requests?

mlodic: I'll do the change you first suggested. I wanted to put this here so that it will be extendable for the future in case we are able to integrate other honeypots effectively like we did for cowrie

PR #585, opened 2025-10-06
2.0.1

regulartim:
# Checklist for creating a new release

- [x] Change version number in `docker/.version` and in `.env_template`
- [x] Verify CI Tests
- [x] Verify that the PR is named with a correct version number like x.x.x
- [x] Merge the PR to the `main` branch. The release will be done automatically by the CI

**Note:** Only use "Merge and commit" as the merge strategy and not "Squash and merge". Using "Squash and merge" makes history between branches misaligned.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1: Frontend 
-FROM node:lts-alpine3.21 AS frontend-build
+FROM node:lts-alpine3.22 AS frontend-build
 
 WORKDIR /
 # copy react source code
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN npm install
 RUN PUBLIC_URL=/static/reactapp/ npm run build
 
 # Stage 2: Backend
-FROM python:3.13-alpine3.21
+FROM python:3.13-alpine
```
mlodic: @regulartim the con about this is the non-reproducibility of the builds because the version may change under the hood at every re-build. This could bring unexpected results. That's the reason why the version was pinned.

regulartim: Ah, ok. This change was suggested by dockerhub, thats why I made it. Will revert.

Issue #647, opened 2025-12-23
Orphaned Decorator in feeds.py


srijan2607:
**File:** `api/views/feeds.py:14`

**Description:** There is an orphaned `@api_view([GET])` decorator on line 14 that has no associated function. This appears to be a copy-paste error or incomplete refactoring.

```python
# Line 14 - ORPHANED DECORATOR
api_view([GET])

@api_view([GET])  # Line 17 - Proper decorator
def feeds(request, feed_type, attack_type, prioritize, format_):
```

**Impact:** Code smell, potential maintenance confusion.

**Recommended Fix:** Remove the orphaned decorator on line 14.

srijan2607:
@regulartim can i work on this ?

regulartim:
Yes.

PR #649, opened 2025-12-23
Fix/bug orphaned decorator

srijan2607:
This pull request makes a minor change to the `api/views/feeds.py` file by removing a redundant decorator. This helps clean up the code without affecting functionality.
---
Related issues : #647


srijan2607:
@regulartim can you take a look

PR #624, opened 2025-12-18
Refactor extraction process. Closes #622.

regulartim:
# Description
This PR introduces a complete rework of the extraction process. The idea is to improve testability, extensibility and maintainability by following some best practices:
- repository pattern: repositories handle data access without containing any processing logic
- single responsibility: every class in the process has one clear and recognizable responsibility
- dependency injection: dependencies are injected through constructors which makes testing much easier
- strategy pattern: makes it easier to add new "special treatment" for honeypots

The new process flow looks like this:
```mermaid
sequenceDiagram
    participant Job as ExtractionJob
    participant Pipeline as ExtractionPipeline
    participant Elastic as ElasticRepository
    participant Factory as StrategyFactory
    participant Strategy as ExtractionStrategy
    participant Processor as IocProcessor
    participant Repo as IocRepository
    
    Job->>Pipeline: execute()
    Pipeline->>Elastic: search(minutes_back)
    Elastic-->>Pipeline: hits[]
    
    loop Each honeypot
        Pipeline->>Factory: get_strategy(honeypot)
        Factory-->>Pipeline: strategy
        Pipeline->>Strategy: extract_from_hits(hits)
        Strategy->>Strategy: iocs_from_hits(hits)
        
        loop Each IOC
            Strategy->>Processor: add_ioc(ioc)
            Processor->>Repo: get_ioc_by_name(name)
            alt IOC exists
                Processor->>Processor: merge_iocs()
                Processor->>Repo: save(ioc)
            else New IOC
                Processor->>Repo: save(ioc)
            end
        end
    end
    
    Pipeline->>Pipeline: UpdateScores()
```
A single ExtractionPipeline instance orchestrates the extraction of all available honeypots. Is uses the ElasticRepository to receive a list of all honeypot hits from a certain time window. For each honeypot it gets the corresponding ExtractionStrategy, which contains all the extraction logic that is specific for a certain type of honeypot (e.g. Cowrie). The ExtractionStrategy uses this logic to create IOC objects and hands them to the IocProcessor, which is responsible for - well - processing them so they can be written to the database via the IocRepository. 

## Key changes (functional)
- Sensors are now extracted in every extraction run. No extra job needed.
- General honeypots that are not in the database yet, are automatically added and extracted (until disabled manually).

## Next steps
- Thoroughly test the new process in a production-like environment. Although I wrote a lot of tests, we might still find some bugs, as the extraction process is quite complex. This should be done before we merge the changes to `main`.
[...]

PR #670, opened 2026-01-02
refactor:honeypot extraction using DB-driven exclusion. closes #631

### LINKED ISSUE #631: Honeypot exclusion list
Following up on #624 , it would be great to have a honeypot exclusion list, which contains all honeypots that we do not want to have in our database and stop them from being extracted. The new extraction process takes all honeypots available in T-Pot's Elastic stack and adds them as general honeypots to the GreedyBear database (alse see #530 ). Some of them are unwanted, for example Ddospot (see #437 ). There are multiple ways to implement such an exclusion list. My ideas are:
- write an migration to add them to the database and disable them
- put their names into some sort of text file and read the file during extraction
- hard code them as a constant

drona-gyawali:
# Description

This PR refactors the `is_ready_for_extraction` method to respect the `GeneralHoneypot.active` flag, making honeypot exclusion fully database-driven. No honeypots are hardcoded, and no migrations were added as discussed previously.

### Changes

* Normalizes honeypot names for cache lookup while preserving original casing in the DB.
* Uses `name__iexact` for case-insensitive DB lookup.
* Dynamically creates new honeypots with `active=True`.
* Updates `_honeypot_cache` to reflect the DB `active` state.
* Extraction pipeline now automatically skips honeypots where `active=False`.

### Notes / Observations

* The code assumes `_honeypot_cache` always uses normalized keys. If elsewhere the cache is populated with original casing, it could cause a cache miss and fallback to the DB. This was already the case before this change.
* Currently, the `GeneralHoneypot.name` field is not unique at the DB level. I didn’t change this, but we might consider adding a unique constraint. Am I missing anything here?

## Related issues
closes #631 

## Type of change

# Checklist

### Important Rules

### DIFF
--- a/greedybear/cronjobs/repositories/ioc.py
+++ b/greedybear/cronjobs/repositories/ioc.py
@@ -92,10 +92,7 @@ class IocRepository:
         Returns:
             The matching GeneralHoneypot, or None if not found.
         """
-        try:
-            return GeneralHoneypot.objects.get(name=name)
-        except GeneralHoneypot.DoesNotExist:
-            return None
+        return GeneralHoneypot.objects.filter(name__iexact=name).first()
 
     def is_empty(self) -> bool:
         """
--- a/greedybear/migrations/0027_disable_unwanted_honeypots.py
+++ b/greedybear/migrations/0027_disable_unwanted_honeypots.py
@@ -8,9 +8,9 @@ def disable_unwanted_honeypots(apps, schema_editor):
     GeneralHoneypot = apps.get_model("greedybear", "GeneralHoneypot")
 
     unwanted = [
-        "ddospot",
+        "Ddospot",
         "ssh-rsa",
-        "nginx",
+        "NGINX",
     ]
 
     for name in unwanted:


drona-gyawali:
Hi @regulartim,
I’ve refactored is_ready_for_extraction to fully respect GeneralHoneypot.active, and all tests are passing.

Could you clarify the plan for setting up the initial honeypot exclusion list? Will it be handled via a migration file, raw SQL, the admin interface, or some other method? I want to make sure our implementation aligns with the intended workflow.

greedybear/cronjobs/repositories/ioc.py
```diff
@@ -119,17 +119,33 @@ def is_enabled(self, honeypot_name: str) -> bool:
...
+            return self._honeypot_cache[normalized]
+
+        try:
+            gh = GeneralHoneypot.objects.get(name__iexact=honeypot_name)
+        except GeneralHoneypot.DoesNotExist:
+            # preserving original casing
+            gh = GeneralHoneypot.objects.create(
+                name=honeypot_name,
+                active=True,
+            )
```
regulartim: Isn't this the exact use case for Django's `Queryset.get_or_create()` ? 
https://docs.djangoproject.com/en/6.0/ref/models/querysets/#django.db.models.query.QuerySet.get_or_create

greedybear/cronjobs/repositories/ioc.py
```diff
@@ -119,17 +119,33 @@ def is_enabled(self, honeypot_name: str) -> bool:
...
+        except GeneralHoneypot.MultipleObjectsReturned:
+            # picking the first one deterministically (rare case)
+            gh = GeneralHoneypot.objects.filter(name__iexact=honeypot_name).order_by("id").first()
```
regulartim: If this happens, something went terribly wrong and we should throw an exception. Do you agree?

regulartim:
> Could you clarify the plan for setting up the initial honeypot exclusion list? Will it be handled via a migration file, raw SQL, the admin interface, or some other method? I want to make sure our implementation aligns with the intended workflow.

Yep, will write that into the issue. :+1:

greedybear/cronjobs/repositories/ioc.py
drona-gyawali: Thanks for the feedback! I simplified the cache to use only normalized keys and removed the fallback for MultipleObjectsReturned so we fail fast on invalid DB state. The logic is now much simpler and fully DB-driven.

greedybear/cronjobs/repositories/ioc.py
drona-gyawali: I looked into get_or_create(), but since we need a case-insensitive lookup (name__iexact) and the name field isn’t unique at the DB level, using get_or_create() safely isn’t possible yet.

However we have another option , we can use .filter().first() to prevent MultipleObjectsReturned errors, followed by get_or_create() to reduce (though not entirely eliminate) race conditions. Once we add a case-insensitive unique constraint or normalize names on write, get_or_create() will be a great fit 

what you think about this.

regulartim: I think we are getting to the core of the problem here: honeypot names in the database are not normalized. I would suggest:
1. for this PR, just rely on the cache as this method did before and remove the whole `try` block (if the cache is normalized, this should be fine for now, right?)
2. open a new issue describing the problem of the not-normalized honeypot names in the DB - I will assign it to you after this PR is completed, if you like
3.  to resolve this, add a case-insensitive unique constraint AND normalize names on write in a separate PR

Does that sound good to you?

drona-gyawali: Yes, for this PR I updated is_ready_for_extraction to rely only on the normalized cache and added the data migration to set up the initial honeypot exclusion list. I didn’t touch existing DB entries or create new honeypots here.

Is this aligned with what you’re expecting for this PR?

regulartim: Yes, it is :+1:

regulartim:
Is this ready to get reviewed again?

drona-gyawali:
> Is this ready to get reviewed again?

Hi @regulartim,

Yes, it’s ready for review again.

I added a migration file for the initial honeypot setup. In the migration, I’m using a try/except (get → create) pattern intentionally, since it’s limited strictly to the migration and only used for the one-time initial setup.

Runtime extraction now relies solely on the normalized cache, with no DB creation or fallback logic involved.

Please let me know if you’d like any adjustments.

regulartim:
Ah, I didn't notice it was ready. You can click "re-request review" next time, then that I get notified. I'll take a look now.

greedybear/cronjobs/repositories/ioc.py
```diff
@@ -119,17 +119,16 @@ def is_enabled(self, honeypot_name: str) -> bool:
...
             honeypot_name: Name of the honeypot to check.
 
         Returns:
             True if the honeypot exists and is enabled, False otherwise.
         """
-        if honeypot_name not in self._honeypot_cache:
-            self.create_honeypot(honeypot_name)
-        return self.is_enabled(honeypot_name)
+        normalized = honeypot_name.lower().strip()
+        return self._honeypot_cache.get(normalized, False)
```
regulartim: Where is the `self.create_honeypot(honeypot_name)` call? This completely destroys the logic of the method! 

Also, I suggested to only write normalized keys to the cache. This is an prerequisite of retrieving normalized keys in my opinion. 
Do you agree?

greedybear/migrations/0026_disabled_unwanted_honeypots.py
```diff
@@ -0,0 +1,40 @@
...
+        "ssh-rsa",
+        "nginx",
+    ]
+
+    for name in unwanted:
+        try:
+            hp = GeneralHoneypot.objects.get(name__iexact=name)
+            if hp.active:
+                hp.active = False
+                hp.save(update_fields=["active"])
```
regulartim: In issue #631 i wrote:

> Be aware that we don't have to change the active flag of existing honeypots in a data migration. The changes in https://github.com/intelowlproject/GreedyBear/pull/624 are not in any release yet so there are no update issues that we have to care about.

But your code does exactly that. I don't want to mess with existing configurations of people running GreedyBear, so please don't change existing entries.

greedybear/cronjobs/repositories/ioc.py
drona-gyawali: Sorry for that, I got a little distracted. I’ve now added the self.create_honeypot(honeypot_name) call back into is_ready_for_extraction, so the method still creates the honeypot if it doesn’t exist.

I also refactored the cache to always use normalized keys via the _normalize_name() helper. This ensures that both retrieval and storage in the cache are consistent, which resolves the issue you mentioned.

greedybear/migrations/0026_disabled_unwanted_honeypots.py
drona-gyawali: I’m really sorry, you’re right about that. I’ve updated the PR to address this.

tests/test_repositories.py
```diff
@@ -136,11 +136,9 @@ def test_add_honeypot_to_ioc_multiple_honeypots(self):
...
 
     def test_existing_honeypots(self):
-        self.assertIn("Cowrie", self.repo._honeypot_cache)
-        self.assertIn("Log4pot", self.repo._honeypot_cache)
-        self.assertIn("Heralding", self.repo._honeypot_cache)
-        self.assertIn("Ciscoasa", self.repo._honeypot_cache)
-        self.assertIn("Ddospot", self.repo._honeypot_cache)
+        expected_honeypots = ["Cowrie", "Log4pot", "Heralding", "Ciscoasa", "Ddospot"]
+        for hp_name in expected_honeypots:
+            self.assertIn(hp_name.lower(), self.repo._honeypot_cache)
```
regulartim: Maybe also use the `_normalize_name` method instead of `lower`?

regulartim (commented):
Looks good. Thanks for your work. One last thing that I am concerned about:
Say we have a honeypot named "Cowrie" in our database. Now for some odd reason we extract an event from T-Pot where the name of the honeypot is lower case "cowrie". What do you think would happen then? Can we cover this with one or more tests?

drona-gyawali:
> Looks good. Thanks for your work. One last thing that I am concerned about:
> Say we have a honeypot named "Cowrie" in our database. Now for some odd reason we extract an event from T-Pot where the name of the honeypot is lower case "cowrie". What do you think would happen then? Can we cover this with one or more tests?

Thanks for the review! 

I’ve added three tests that together cover the case-insensitive handling of honeypot names:

1. Ensures an enabled honeypot like `"Cowrie"` works even when called as lowercase `"cowrie"`.
2. Ensures a disabled honeypot returns `False` even with a lowercase lookup.
3. Confirms that special honeypots (`Cowrie`, `Log4Pot`) are always enabled and normal honeypots respect their `active` status, all in a case-insensitive manner.

I hope this tests  satisfy the concern about case-insensitive extraction from T-Pot events.

tests/test_repositories.py
drona-gyawali: sorry , i will update that

greedybear/migrations/0027_disable_unwanted_honeypots.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Ensure unwanted honeypots exist and are disabled.
+    """
+    GeneralHoneypot = apps.get_model("greedybear", "GeneralHoneypot")
+
+    unwanted = [
+        "ddospot",
+        "ssh-rsa",
+        "nginx",
+    ]
```
regulartim, line 14: Just to be on the safe side here, add the names case-sensitive, like I wrote them into the issue.

tests/test_repositories.py
```diff
@@ -136,17 +136,44 @@ def test_add_honeypot_to_ioc_multiple_honeypots(self):
...
+    def test_is_ready_for_extraction_case_insensitive(self):
+        GeneralHoneypot.objects.create(name="Cowrie", active=True)
+        result = self.repo.is_ready_for_extraction("cowrie")
+        self.assertTrue(result)
+        self.assertEqual(GeneralHoneypot.objects.filter(name__iexact="cowrie").count(), 1)
+
+    def test_get_hp_by_name_insensitive(self):
+        GeneralHoneypot.objects.create(name="Cowrie", active=True)
+        result = self.repo.get_hp_by_name("cowrie")
+        self.assertIsNotNone(result)
```
regulartim, line 157: I added this test and it fails. However I think it should not. Do you agree? 

The fix should be easy, just make a insensitive lookup in the corresponding method.

drona-gyawali: Yes, I agree 👍, The test should pass. I’ve updated get_hp_by_name to use a case-insensitive lookup .

greedybear/migrations/0027_disable_unwanted_honeypots.py
drona-gyawali: I’ll update the migration to use the names exactly as written in the issue.

greedybear/cronjobs/repositories/ioc.py
```diff
@@ -92,10 +92,7 @@ def get_hp_by_name(self, name: str) -> GeneralHoneypot | None:
         Returns:
             The matching GeneralHoneypot, or None if not found.
         """
-        try:
-            return GeneralHoneypot.objects.get(name=name)
-        except GeneralHoneypot.DoesNotExist:
-            return None
+        return GeneralHoneypot.objects.filter(name__iexact=name).first()
```
regulartim, line 95: Why did you remove the `DoesNotExist` check?

regulartim: Ah, because `first()` returns `None` if the query set is empty. Good job! :+1:

regulartim (approved):

regulartim:
Thanks for your work! :) As a follow up, would you like to open a new issue describing the problem of the not-normalized honeypot names in the DB and how to handle this?

drona-gyawali:
> Thanks for your work! :) As a follow up, would you like to open a new issue describing the problem of the not-normalized honeypot names in the DB and how to handle this?

Thanks for the review and merge! 
I’ll open a new issue describing the problem  and the approach to handle this.

Issue #632, opened 2025-12-20
Stop adding hard-coded general honeypots


regulartim:
On fresh installations, we currently write a set of supported general honeypots into the database. This is done via a migration file:

https://github.com/intelowlproject/GreedyBear/blob/5ad7792b5be84cd746eee61f54355b3e4a769890/greedybear/migrations/0008_auto_20230120_1548.py#L1-L37

As GreedyBear now dynamically supports new honeypots that it finds in the Elasticsearch data, this is no longer necessary and rather confusing behavior. I suggest to not add any hard-coded general honeypots to fresh installations. (Maybe except for the ones we want to deliberately disable, see #631 )

Sumit-ai-dev:
Hi @regulartim! I'd like to work on this issue.

 I've reviewed all related issues to understand the architectural evolution:

#530: Identified the need for dynamic honeypot discovery
#624: Implemented dynamic extraction from all T-POT honeypots
#437: Highlighted false positives from Ddospot
#631: Implemented database-driven exclusions via GeneralHoneypot.active flag
My Understanding: After PR #624, GreedyBear automatically discovers all honeypots from T-POT's ElasticSearch, making the hard-coded list in migration  0008_auto_20230120_1548.py obsolete. Exclusions are now handled separately through #631's database approach.

Proposed Solution: Convert the generalHoneypot() migration function to a no-op (using pass) while maintaining migration chain integrity. I've implemented this locally:

def generalHoneypot(apps, schema_editor):
    """
    This function is kept as a no-op to maintain migration chain integrity.
Honeypot exclusions are now handled via the GeneralHoneypot.active flag (see #631).
    """
    pass

Could you please assign this issue to me?

regulartim:
Hey @Sumit-ai-dev ! :) 
Your proposed solution changes and old migration file. Do you think that it is save to do that? Other migrations might depend on the changes made in `0008_auto_20230120_1548.py`. Do you see a different way of solving this?

Sumit-ai-dev:
Hi @regulartim,

Yes, this approach is safe. The key point is that migration 0009 depends on 0008 existing in the chain, not on the data it creates. I verified this by checking the dependencies:

From 0009_alter_ioc_general_field.py
dependencies = [
    ("greedybear", "0008_auto_20230120_1548"),
]


[...]

PR #693, opened 2026-01-09
refactor : honeypot creation and enforce case-insensitive uniqueness . Closes #689

### LINKED ISSUE #689: Normalize `GeneralHoneypot` names and enforce uniqueness
Following the merge of #670 ,  we introduced normalization (lowercase ) in the honeypot cache 

However, the `GeneralHoneypot.name` field in the database is **not normalized on creation or update**, which could lead to inconsistencies:

* Duplicate entries if names differ only in case (e.g., `"Cowrie"` vs `"cowrie"`).
* Cache misses or redundant DB queries if incoming honeypot names from sources like T-Pot vary in casing.
* Potential issues with special honeypots (e.g., `"Cowrie"`, `"Log4Pot"`) if their casing doesn't match expectations.

Currently:

* The DB field `name` is a `CharField` without uniqueness or case-insensitivity enforced.
* Normalization happens in the repository layer (`_normalize_name`), but the DB stores the original casing.

This issue aims to **fully normalize honeypot names in the database** to prevent these problems and ensure all operations are case-insensitive and consistent.

### **Approach**

1. **Normalize on Save**: Override the `save` method in `GeneralHoneypot` to normalize `name` (e.g., lowercase) before saving. This ensures all DB entries are consistent.
2. **Add Unique Constraint**: Make `name` unique at the DB level with a case-insensitive index if possible (e.g., using Django's `UniqueConstraint` with `Lower('name')` 
3. **Data Migration**: Create a migration to normalize existing entries in the DB (update all `name` to lowercase). Since these changes are pre-release, this should be low-risk.
4. **Update**  `create_honeypot` in repo to direct support normalized name
5. **Tests**: Add tests to cover creation, lookup, and extraction with mixed-case inputs to ensure normalization works.

---

### **Expected Outcome**

* All honeypot names in the DB are lowercase and unique.
* Future honeypots cannot introduce case-related duplicates.

drona-gyawali:
# Description

This PR addresses issues with honeypot creation and ensures case-insensitive uniqueness:

- Added a unique constraint on GeneralHoneypot.name (case-insensitive) at the database level.
- Updated the create_honeypot method to handle IntegrityError gracefully.
- Used get_or_create in test cases to avoid duplicate key errors and ensure reliable test setup.
- Added and improved test cases for create_honeypot.
  
## Related issues
closes #689 

## Type of change

# Checklist

### Important Rules

### DIFF
--- a/greedybear/cronjobs/repositories/ioc.py
+++ b/greedybear/cronjobs/repositories/ioc.py
@@ -47,13 +47,17 @@ class IocRepository:
 
     def create_honeypot(self, honeypot_name: str) -> GeneralHoneypot:
         """
-        Create a new honeypot and save it to the database.
+        Create a new honeypot or return an existing one.
+
+        If a honeypot with the same name (case-insensitive) already exists,
+        recover and return the existing one instead. This method also updates
+        the internal honeypot cache accordingly.
 
         Args:
             honeypot_name: Name for the new honeypot.
 
         Returns:
-            The newly created GeneralHoneypot instance.
+            A GeneralHoneypot instance (newly created or existing).
         """
         normalized = self._normalize_name(honeypot_name)
 
@@ -62,10 +66,11 @@ class IocRepository:
                 name=honeypot_name,
                 active=True,
             )
-        except IntegrityError:
+        except IntegrityError as e:
+            self.log.error(f"IntegrityError creating honeypot '{honeypot_name}': {e}")
             honeypot = self.get_hp_by_name(honeypot_name)
             if honeypot is None:
-                raise
+                raise e
 
         self._honeypot_cache[normalized] = honeypot.active
         return honeypot
--- a/tests/test_repositories.py
+++ b/tests/test_repositories.py
@@ -157,11 +157,11 @@ class TestIocRepository(CustomTestCase):
         self.assertIsNotNone(result)
 
     def test_disabled_honeypot_case_insensitive(self):
-        GeneralHoneypot.objects.create(name="Feralding", active=False)
+        GeneralHoneypot.objects.create(name="Testpot69", active=False)
 
         # reiniting repo after DB change to refresh the cache
         repo = IocRepository()
-        result = repo.is_ready_for_extraction("feralding")
+        result = repo.is_ready_for_extraction("testpot69")
         self.assertFalse(result)
 
     def test_special_and_normal_honeypots(self):


tests/__init__.py
```diff
@@ -19,9 +19,9 @@ class CustomTestCase(TestCase):
     def setUpTestData(cls):
         super().setUpTestData()
 
-        cls.heralding = GeneralHoneypot.objects.create(name="Heralding", active=True)
-        cls.ciscoasa = GeneralHoneypot.objects.create(name="Ciscoasa", active=True)
-        cls.ddospot = GeneralHoneypot.objects.create(name="Ddospot", active=False)
+        cls.heralding = GeneralHoneypot.objects.get_or_create(name="Heralding", defaults={"active": True})[0]
```
drona-gyawali, line 22: I use get_or_create to ensure the honeypots exist without duplicating entries, so the tests stay consistent and don’t fail because of unique constraints

regulartim:
Hey @drona-gyawali ! :) Independent of the review I wanted to ask you why you think that this is a breaking change? What exactly will break, when we merge this?

tests/__init__.py
regulartim: In this file we set up a fresh DB for the purpose of running the test. This means, before this lines are executed, the DB is empty. I agree that `get_or_create` generally is a safer approach, but in this specific case, it really does not matter. So I would prefer to not have these changes in this PR. Or do I miss something? Do you see a case where using `create` creates a problem?

tests/test_repositories.py
```diff
@@ -157,6 +157,7 @@ def test_get_hp_by_name_insensitive(self):
         self.assertIsNotNone(result)
 
     def test_disabled_honeypot_case_insensitive(self):
+        GeneralHoneypot.objects.all().delete()
```
regulartim: Why is the deletion necessary? Depending on when this test runs, it might influence other tests.

tests/test_repositories.py
```diff
@@ -174,6 +175,31 @@ def test_special_and_normal_honeypots(self):
...
+        with self.assertNoLogs("greedybear.cronjobs.repositories.ioc", level="WARNING"):
+            hp = repo.create_honeypot("log4pot")
+            self.assertEqual(hp.name, "Log4Pot")
+
+    def test_create_new_honeypot_creates_and_updates_cache(self):
+        self.repo._honeypot_cache.clear()
+        hp = self.repo.create_honeypot("NewPot")
+        self.assertEqual(hp.name, "NewPot")
+        self.assertTrue("newpot" in self.repo._honeypot_cache)
+        self.assertTrue(hp.active)
```
regulartim, line 207: This test does not check whether the honeypot is actually created in the DB or not. Are you aware of this?

tests/test_repositories.py
```diff
@@ -174,6 +175,31 @@ def test_special_and_normal_honeypots(self):
         self.assertFalse(repo.is_ready_for_extraction("NormalPot"))
         self.assertFalse(repo.is_ready_for_extraction("normalpot"))
 
+    def test_create_honeypot_case_insensitive_uniqueness(self):
+        GeneralHoneypot.objects.all().delete()
```
regulartim: Why is the deletion necessary? Depending on when this test runs, it might influence other tests.

greedybear/cronjobs/repositories/ioc.py
```diff
@@ -52,10 +54,21 @@ def create_honeypot(self, honeypot_name: str) -> GeneralHoneypot:
...
+
+        try:
+            honeypot = GeneralHoneypot(name=honeypot_name, active=True)
+            honeypot.save()
+        except IntegrityError:
+            honeypot = self.get_hp_by_name(honeypot_name)
+            if honeypot is None:
+                raise
+
+        self._honeypot_cache[normalized] = honeypot.active
```
regulartim, line 75: Try to simplify. You call `self.get_hp_by_name(honeypot_name)` two times in one method. This is a indicator, that there is a way to simplify the logic.

Suggestion: Assume that there is no honeypot named `honeypot_name`, when this function is called. If `create` raises and `IntegrityError`, log the error using `self.log`. Then, try to `get_hp_by_name`. If this also fails, re-raise the `IntegrityError`. What do you think about that?

tests/__init__.py
drona-gyawali: I used get_or_create() here instead of  create() to prevent IntegrityError due to the case-insensitive unique constraint.

Since CustomTestcase is used as a base class for multiple test classes, the initial honeypot entries (Heralding, Ciscoasa, Ddospot are already present in the database when another test class runs. Using create() in that scenario triggers a duplicate key error, because the DB still contains the entries from a previous test class.

get_or_create() ensures that:

1. If the honeypot already exists (even with different casing), it is reused.
2. If it doesn’t exist, it is created.

Although we do have a tearDownClass cleanup, the error occurs during setUpTestData, so tearDownClass is never reached and the cleanup does not run. That’s why the workaround in test_repositories was needed.
 
image

I understand that in the context of a single, fresh DB, create() is necessary,   but in our test setup with multiple classes inheriting CustomTestCase , get_or_create() prevents duplicate key issues. what do you think about this ?

drona-gyawali:
> Hey @drona-gyawali ! :) Independent of the review I wanted to ask you why you think that this is a breaking change? What exactly will break, when we merge this?

Thanks for pointing that out :)
You’re right  this is not actually a breaking change. I mistakenly checked the breaking-change box.

greedybear/cronjobs/repositories/ioc.py
drona-gyawali: That makes sense, thanks for pointing it out
You’re right calling get_hp_by_name() twice is a signal that the logic can be simplified.

I’ll refactor the method following this pattern and push an update soon.

tests/__init__.py
regulartim: Ok, this really surprises me. I would like to check this behavior myself before coming to an conclusion. Anyway, thanks for the detailed explanation! :)

regulartim: I understand it now. Your code is fine. :+1:

drona-gyawali: Thanks!  Just to make sure I fully understand so the reason get_or_create() is needed here is the same scenario I described, where multiple test classes share the DB and setUpTestData runs before tearDownClass, causing potential duplicates? Or is it a slightly different issue?

regulartim: The `setUpTestData` expects a fresh and empty database, but before it runs, [an old migration](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/migrations/0008_auto_20230120_1548.py) populates the database with some entries in `GeneralHoneypot`, one of them is Heralding. So in `setUpTestData` the `create()` call fails, because Heralding already exists. This was not evident in the past, because (1) there was no unique constraint and (2) after the first test `tearDownClass()` deletes the whole database again. 

In my opinion we should not worry about this right now as it is planed to basically revert the problematic migration (see #632). Then we _could_ use `create()` again, but your approach is solid, so we can just keep it long term.

tests/test_repositories.py
```diff
@@ -174,6 +174,54 @@ def test_special_and_normal_honeypots(self):
...
+        self.assertEqual(GeneralHoneypot.objects.count(), initial_count + 1)
+        self.assertEqual(GeneralHoneypot.objects.get(name__iexact="testpot123").name, "TestPot123")
+
+    def test_create_honeypot_integrity_error_handling(self):
+        initial_count = GeneralHoneypot.objects.count()
+        GeneralHoneypot.objects.create(name="Log4PotTest123", active=True)
+
+        try:
+            with transaction.atomic():
+                GeneralHoneypot.objects.create(name="log4pottest123", active=True)
```
drona-gyawali, line 195: In this test, I  try to create a honeypot with a duplicate name. and wrapping  the insert in transaction.atomic()  to isolates the risky operation like if it fails, the transaction rolls back, keeping the session safe for further queries. In , create_honeypot already handles IntegrityError, However,  adding a transaction there depends, what you think ?.

tests/test_repositories.py
drona-gyawali: I had to add this to delete the existing data, as we discussed. In the updated version, I simply changed the name from 'Heralding' to 'Feralding to get rid of unique constraint error.

tests/test_repositories.py
regulartim: I would rather not use a transaction. I don't fully understand what is the benefit of a transaction here. We assume that all test pass. If this one does not, we know something is wrong in the code. And in that case, if it also causes subsequent tests to fail, that's not too bad i my opinion.

tests/test_repositories.py
```diff
@@ -157,11 +157,11 @@ def test_get_hp_by_name_insensitive(self):
         self.assertIsNotNone(result)
 
     def test_disabled_honeypot_case_insensitive(self):
-        GeneralHoneypot.objects.create(name="Heralding", active=False)
+        GeneralHoneypot.objects.create(name="Feralding", active=False)
 
         # reiniting repo after DB change to refresh the cache
         repo = IocRepository()
-        result = repo.is_ready_for_extraction("heralding")
+        result = repo.is_ready_for_extraction("feralding")
```
regulartim: This is confusing, as it just looks like a typo. If it is intentional, then please use a more distinctive name.

greedybear/cronjobs/repositories/ioc.py
regulartim: You did not really do what I was suggesting. Please read again. My opinion did not change. Also, if you change the logic of a method, please update the docstring. This method now does more than just "Create a new honeypot and save it to the database".

tests/test_repositories.py
drona-gyawali: Thanks for the review, in test I try to insert a honeypot with a name that already exists in the database, it immediately raises an IntegrityError.

If this happens inside a test without handling it, the transaction becomes broken, and any further queries in that test fail with a TransactionManagementError. That’s why two of  test cases were failing when I  didn’t use transaction.atomic().

error log:  https://www.notion.so/2e722ce900ac80dfb57ffd61f74da7ea?source=copy_link

However, I understand your point that we should not rely on transactions in the function. In this case, we  need to change how we handle the IntegrityError completely .

tests/test_repositories.py
drona-gyawali: sorry for that , will use a more distinctive name

tests/test_repositories.py
regulartim: Ah, I see. Then just keep the `transaction.atomic()` context as it is. :+1:

greedybear/cronjobs/repositories/ioc.py
drona-gyawali: Thank you for the feedback. I realize I misunderstood your suggestion earlier, Now I’ve updated the method based on your feedback. Could you please review the latest version? I hope I’ve implemented it correctly this time.

PR #697, opened 2026-01-11
Tests(Firehol): Adding and improving tests for Firehol.

### LINKED ISSUE #694: Tests: FireHol tests improvement
Current test file of FireHol, misses several important edge cases: network or HTTP failures, behavior when entries already exist, duplicate IPs within the same source, empty or comment-only blocklists, invalid input lines, and the cleanup of entries older than 30 days. It also does not assert that raise_for_status() or logging calls are executed, nor does it verify that processing continues when one source fails.

I would like to work on improving this test file and decoupling the functions.

amishhaa:
# Description
This PR fixes and adds new test cases for test_firehol.py.
Previously .status_code was incorrectly used and we should be using .raise_for_status instead to mock the behavior of the status code. Additionally we can also test the method which cleans up old entries, and test edges like failure from certain network codes.

## Related issues
Fixes #694.
## Type of change

# Checklist

- [xx] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
  
### Important Rules

### DIFF
--- a/tests/greedybear/cronjobs/test_firehol.py
+++ b/tests/greedybear/cronjobs/test_firehol.py
@@ -58,7 +58,6 @@ class FireHolCronTestCase(CustomTestCase):
         # Setup mock response
         mock_response_blocklist_de = MagicMock()
         mock_response_blocklist_de.text = "# blocklist_de\n1.1.1.1\n2.2.2.2"
-        mock_get.return_value = mock_response_blocklist_de
 
         mock_response_bruteforceblocker = MagicMock()
         mock_response_bruteforceblocker.raise_for_status.side_effect = requests.exceptions.HTTPError("404 Client Error")
@@ -86,7 +85,6 @@ class FireHolCronTestCase(CustomTestCase):
         # Setup mock response
         mock_response_blocklist_de = MagicMock()
         mock_response_blocklist_de.text = "# blocklist_de\n"
-        mock_get.return_value = mock_response_blocklist_de
 
         mock_response_bruteforceblocker = MagicMock()
         mock_response_bruteforceblocker.raise_for_status.side_effect = requests.exceptions.HTTPError("404 Client Error")


amishhhaaaa:
@regulartim, please check this PR out :)!

tests/greedybear/cronjobs/test_firehol.py
```diff
@@ -57,3 +56,114 @@ def side_effect(url, timeout):
...
+        mock_response_bruteforceblocker.raise_for_status.side_effect = requests.exceptions.HTTPError("404 Client Error")
+        mock_response_bruteforceblocker.text = "# bruteforceblocker\n1.1.1.1"
+
+        # Side effect for multiple calls
+        def side_effect(url, timeout):
+            if "blocklist_de" in url:
+                return mock_response_blocklist_de
+            elif "bruteforceblocker" in url:
+                return mock_response_bruteforceblocker
+            return MagicMock(status_code=404)
```
regulartim: This function is duplicated across multiple tests. Can we extract it such that all tests use the same `side_effect` function?

tests/greedybear/cronjobs/test_firehol.py
```diff
@@ -57,3 +56,114 @@ def side_effect(url, timeout):
...
+
+    @patch("greedybear.cronjobs.firehol.requests.get")
+    def test_run_creates_some_firehol_entries(self, mock_get):
+        # Setup mock response
+        mock_response_blocklist_de = MagicMock()
+        mock_response_blocklist_de.text = "# blocklist_de\n1.1.1.1\n2.2.2.2"
+        mock_get.return_value = mock_response_blocklist_de
+
+        mock_response_bruteforceblocker = MagicMock()
+        mock_response_bruteforceblocker.raise_for_status.side_effect = requests.exceptions.HTTPError("404 Client Error")
```
regulartim, line 63: Why do we need this? I don't understand because the mock has a text (same in subsequent tests.

regulartim (changes requested):
I like it! Could you also add a test case for a list containing an invalid IP address?

tests/greedybear/cronjobs/test_firehol.py
amishhhaaaa: @regulartim Thats a great idea! the code looks much cleaner and readable after refactoring it out into a method.

tests/greedybear/cronjobs/test_firehol.py
amishhhaaaa: very sorry, but i dont understand, are we talkign about the raise for status side effect?

amishhhaaaa:
@regulartim regarding adding a test for invalid ip, from following the function call hierarchy where these requests are actually processed from firehol.py, firehol.py uses FireHolList from models.py, i was not able to find any place where invalid IPs are pre-processed/validated, and if we add tests here for invalidating ips, we might also need to do it from the root level in the FireHolList data structure in models.py which is actually responsible for invalidating ips but no exceptions are raised there basically everything is valid currently so we cannot test it. I was also not able to find any unit tests for it in test_models.py, please let me know what you think, because in my opinion is something like invalidating ip is required it would have to be implemented in a fresh pr because from my knowledge it doesnt exist currently.

tests/greedybear/cronjobs/test_firehol.py
regulartim: Yes.

regulartim:
> @regulartim regarding adding a test for invalid ip, from following the function call hierarchy where these requests are actually processed from firehol.py, firehol.py uses FireHolList from models.py, i was not able to find any place where invalid IPs are pre-processed/validated, and if we add tests here for invalidating ips, we might also need to do it from the root level in the FireHolList data structure in models.py which is actually responsible for invalidating ips but no exceptions are raised there basically everything is valid currently so we cannot test it. I was also not able to find any unit tests for it in test_models.py, please let me know what you think, because in my opinion is something like invalidating ip is required it would have to be implemented in a fresh pr because from my knowledge it doesnt exist currently.

Sorry, my bad. I  confused the FireHolCron with MassScannersCron, where we actually have validation. So, you don't need to add tests for this. However I do think we should implement some form of validation, maybe the same process we have in MassScannersCron. Would you like to open an issue for that?

tests/greedybear/cronjobs/test_firehol.py
amishhhaaaa: > Why do we need this?

because the test name is test_run_creates_some_firehol_entries so i want this mock to fail thats why i am raising a 404 status from .raise_for_status otherwise for other mocks that line is not required as i dont want it to fail on status code.



> I don't understand because the mock has a text (same in subsequent tests.

So sorry didnt get this.

let me know if i am not understanding the point correctly.

amishhhaaaa:
> Would you like to open an issue for that?

@regulartim Yes absolutely, i think we can wait for this to get resolved then the validation can be added on top with the tests that are required for it, i will open an issue and mention this PR as a blocker. let me know if you would like if it is done differently, like adding the validations first and then the tests can be pushed at once, or everything can be done in this pr itself.

tests/greedybear/cronjobs/test_firehol.py
regulartim: I cloned your code an tested some stuff. I think what confuses me here (and also in other tests) is that the same mock, e.g. `mock_response_bruteforceblocker`, has a `text` field AND a `raise_for_status` field. But only one of them is needed in a single tests, right? Either the mock raises 404 OR it returns a text. But not both. Am I right?

regulartim:
Thanks for opening the issue #702 . I would suggest to finish this PR independent of the issue. The person who solves #702 can then add the corresponding tests.

amishhhaaaa:
> I would suggest to finish this PR independent of the issue.

okay got it :)

tests/greedybear/cronjobs/test_firehol.py
amishhhaaaa: @regulartim i think i got your point. so basically the object has two fields one is .text which contains the ips and one is .raise_for_status field, currently text cannot raise any error so it cannot fail on that under any condition as already discussed we do not have a validation in place for ips for the text to fail on.
so we only have .raise_for_status to raise errors and this field is called earlier than .text in the code in firehol.py so once it fails on it we do not actually need to explicitly define .text if we are intentionally making it fail earlier on .raise_for_status, so i have removed the extra lines. It was a great catch!

regulartim: Right. The reason I am so strict here is that the test should be easy to understand. They test our code and are not tested themselves. So it is important that they are unambiguous and lean. If the PR is ready, feel free to ask for a review again.

amishhhaaaa: @regulartim Yes the pr is ready in my opinion, I have removed the unnecessary lines. Yep strict code reviews always make sense! They're a part of having a good codebase and the developer also learns a lot so it's a win win.

PR #718, opened 2026-01-17
feat(api): add ASN-aggregated IOC statistics . CLOSES #458

### LINKED ISSUE #458: AS information API
We track the ASN of each IoC. This allows us to group IoCs by ASN and aggregate statistics for that ASN, which may be useful information for some people.

drona-gyawali:
# Description

This change introduces a new authenticated API endpoint that aggregates IOC data by ASN. The endpoint groups all matching IOCs under their respective ASNs and computes summary statistics, including IOC count, total attack count, total interaction count, total login attempts, expected IOC count (derived from recurrence probability), and expected interactions. It also returns the set of unique honeypots associated with each ASN. The implementation reuses the same filtering and authentication logic as the Advanced Feeds API to avoid code duplication, while intentionally returning a JSON-only response tailored for aggregated data use cases.

- Reuses existing Advanced Feeds query building logic to avoid duplication
- Aggregation logic is isolated in a utility function
- Floating point values are rounded to 4 decimals for stable output
- Includes test coverage following existing feeds test patterns

## Related issues
closes #458 

## Type of change

# Checklist

### Important Rules

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -138,10 +138,7 @@ class ASNFeedsOrderingSerializer(FeedsRequestSerializer):
 
         if field_name not in self.ALLOWED_ORDERING_FIELDS:
             raise serializers.ValidationError(
-                f"""
-                 Invalid ordering field for ASN aggregated feed: '{field_name}'.
-                 Allowed fields: {", ".join(sorted(self.ALLOWED_ORDERING_FIELDS))}
-                """
+                f"Invalid ordering field for ASN aggregated feed: '{field_name}'. Allowed fields: {', '.join(sorted(self.ALLOWED_ORDERING_FIELDS))}"
             )
 
         return ordering
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -8,7 +8,7 @@ from ipaddress import ip_address
 
 from django.conf import settings
 from django.contrib.postgres.aggregates import ArrayAgg
-from django.db.models import Count, F, Max, Min, Q, Sum
+from django.db.models import Count, F, Max, Min, Sum
 from django.http import HttpResponse, HttpResponseBadRequest, StreamingHttpResponse
 from rest_framework import status
 from rest_framework.response import Response
@@ -371,7 +371,6 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
             honeypots=ArrayAgg(
                 "general_honeypot__name",
                 distinct=True,
-                filter=Q(general_honeypot__active=True),
             )
         )
     )


api/views/utils.py
```diff
@@ -326,3 +327,57 @@ def is_sha256hash(string: str) -> bool:
...
+            "exp_inter": 0.0,
+        }
+    )
+
+    for ioc in iocs:
+        asn = ioc.asn
+        if asn is None:
+            continue
+
+        e = aggregated[asn]
```
regulartim: Can you think of a better name than `e`?

api/views/utils.py
```diff
@@ -326,3 +327,57 @@ def is_sha256hash(string: str) -> bool:
...
+
+        e["ioc_count"] += 1
+        e["attack"] += ioc.attack_count or 0
+        e["interactions"] += ioc.interaction_count or 0
+        e["logins"] += ioc.login_attempts or 0
+        e["exp_ioc"] += ioc.recurrence_probability or 0.0
+        e["exp_inter"] += ioc.expected_interactions or 0.0
```
regulartim: You should not have to worry about these fields being `None`, as they have an default value set in `models.py`. Did you encounter any instances where these fields were `None`?

api/views/feeds.py
```diff
@@ -116,3 +118,47 @@ def feeds_advanced(request):
...
+        attack_type (str): Type of attack to filter. Supported: `scanner`, `payload_request`, or `all`. Default: `all`.
+        max_age (int): Maximum number of days since last occurrence. Default: 3.
+        min_days_seen (int): Minimum number of days an IOC must have been seen. Default: 1.
+        include_reputation (str): `;`-separated list of reputations to include, e.g., `known attacker;`. Default: include all.
+        exclude_reputation (str): `;`-separated list of reputations to exclude, e.g., `mass scanner;bot`. Default: exclude none.
+        feed_size (int): Maximum number of IOC items considered for aggregation. Default: 5000.
+        ordering (str): Field to order results by, with optional `-` prefix for descending. Default: `-last_seen`.
+        verbose (bool): Not used in this endpoint; included for consistency with Advanced Feeds API. Default: `false`.
+        paginate (bool): Not supported in this endpoint; ignored if provided. Default: `false`.
+        format (str): Response format type; only `json` is supported. Default: `json`.
```
regulartim: You can delete these lines. Every query parameter that is not explicitly supported will be ignored, so these lines do not really provide useful information..

api/views/feeds.py
```diff
@@ -116,3 +118,47 @@ def feeds_advanced(request):
...
+
+    Args:
+        request: The incoming request object.
+        feed_type (str): Type of feed to retrieve. Supported: `cowrie`, `log4j`, general honeypot names, or `all`. Default: `all`.
+        attack_type (str): Type of attack to filter. Supported: `scanner`, `payload_request`, or `all`. Default: `all`.
+        max_age (int): Maximum number of days since last occurrence. Default: 3.
+        min_days_seen (int): Minimum number of days an IOC must have been seen. Default: 1.
+        include_reputation (str): `;`-separated list of reputations to include, e.g., `known attacker;`. Default: include all.
+        exclude_reputation (str): `;`-separated list of reputations to exclude, e.g., `mass scanner;bot`. Default: exclude none.
+        feed_size (int): Maximum number of IOC items considered for aggregation. Default: 5000.
```
regulartim: I think we should not use a limit like this in this API.

drona-gyawali:
> Nice, thank you!
> 
> A few general things:
> 
>     1. In my mind the aggregation would be done on a database level. Your approach also works, of course, but I am a little bit concerned regarding performance on instances with millions of IoCs.
> 
>     2. Ordering is kind of tricky here. I tested your code and I was not able to order by `ioc_count`. But intuitively I would expect that to work. I think we should support to order by any field that is present in the output, if sensible (ordering by `honeypots` is not).
> 
>     3. The IOC model has more field that have not been included here yet. Do you think you could also include `first_seen` and `last_seen`? Or do you think that this is not useful?

Hi, @regulartim,  Thanks a lot for the review and the feedback!

I wanted to explain a bit why I initially went with Python-level aggregation and where I’m currently unsure, so I’d really appreciate your guidance.

My first thought was also to aggregate at the database level using values("asn") + annotate(...). However, I ran into a couple of structural issues with the current setup of get_queryset, which is shared by other API:

get_queryset applies slicing ([:feed_size]) after ordering. Once a queryset is sliced, Django does not allow further aggregation, which makes DB-level grouping impossible unless we disable slicing using a flag (doSlice=True). However i was afraid to change the fx which is used by other apis without your permission.

> Error i recieved using this method:
> Screenshot from 2026-01-18 18-23-21

And If we slice before aggregation, the aggregation sees only the first feed_size raw IOCs. This produces incorrect totals (ioc_count, attack_count, etc.), because other IOCs outside the slice are ignored. For example, if ASN 13335 has 10,000 IOCs, slicing at 5000 will undercount.

When aggregating by ASN and joining against general_honeypot, each IOC can appear multiple times at the sql level (one row per honeypot). This leads to inflated sums (e.g., attack_count, interaction_count).

For example, a single IOC linked to two honeypots is counted twice in SUM(attack_count) in aggregation . I guees it is many to many joins and duplicated rows issue.

I think that get_queryset needs a big refactor here

I’d love to hear what you think is the best tradeoff here, especially given the shared nature of get_queryset. I’m happy to adapt the implementation based on your recommendation.

regulartim:
Hey @drona-gyawali ! Thanks for your detailed explanation. I think the best approach would be to aggregate on the DB level. If this makes it necessary to refactor or even split up `get_queryset`, that's fine. I recognize that it is too inflexible with the slicing and the way it queries the DB. If that's not possible, it is also possible to write a separate function, exclusively for this API.

> When aggregating by ASN and joining against general_honeypot, each IOC can appear multiple times at the sql level (one row per honeypot). This leads to inflated sums (e.g., attack_count, interaction_count).

This is kind of strange. If a aggregation function is used on the `general_honeypot` filed, this should not happen. Did you use `ArrayAgg`?

drona-gyawali:
> Hey @drona-gyawali ! Thanks for your detailed explanation. I think the best approach would be to aggregate on the DB level. If this makes it necessary to refactor or even split up `get_queryset`, that's fine. I recognize that it is too inflexible with the slicing and the way it queries the DB. If that's not possible, it is also possible to write a separate function, exclusively for this API.
> 
> > When aggregating by ASN and joining against general_honeypot, each IOC can appear multiple times at the sql level (one row per honeypot). This leads to inflated sums (e.g., attack_count, interaction_count).
> 
> This is kind of strange. If a aggregation function is used on the `general_honeypot` filed, this should not happen. Did you use `ArrayAgg`?

Extremely sorry for the late reply! I used ArrayAgg(distinct=True) for honeypots and distinct=True on all other fields to avoid inflated sums, since thing were failing without it. The iocs_qs comes from get_queryset.  raw code for reference:

```
def asn_aggregated_queryset(iocs_qs):   
    return (
        iocs_qs
        .exclude(asn__isnull=True)
        .values("asn")
        .annotate(
            ioc_count=Count("id", distinct=True),
            total_attack_count=Sum("attack_count", distinct=True),
            total_interaction_count=Sum("interaction_count", distinct=True),
            total_login_attempts=Sum("login_attempts", distinct=True),
            expected_ioc_count=Sum("recurrence_probability", distinct=True),
            expected_interactions=Sum("expected_interactions", distinct=True),
            honeypots=ArrayAgg("general_honeypot__name", distinct=True),
        )
    )

```
I’ll push the new version soon.

regulartim:
> Extremely sorry for the late reply!

Don't worry, we all have other stuff to do! :) 

> I’ll push the new version soon.

Cool, looking forward to that!

drona-gyawali:
Hi @regulartim , In this new version, I implemented DB-level aggregation for the ASN feed. While building this, I had to introduce a few things:

New serializer (ASNFeedsOrderingSerializer) – I inherited from FeedsRequestSerializer because the base serializer already provides default handling for parameters like max_age, feed_type, and attack_type. The main reason for creating the new serializer was that the base serializer’s ordering validation is strict and only allows model fields. Since aggregation introduces annotated/non-model fields (like ioc_count, total_attack_count), I needed to add custom validation here.

resolve_aggregation_ordering utility – This ensures that our aggregation endpoint defaults to ordering by -ioc_count instead of -last_seen, bypassing the default injection from feed_params. I tried to make this dynamic so that any future aggregation API can leverage the same pattern without reinventing the wheel.

The overall goal was to make the developer experience better and avoid complexity when building future aggregation endpoints. Anyone adding a new aggregation API only needs to inherit and customize the ordering/validation, without rewriting everything.

I hope this aligns with your expectations. If you feel any part of this design needs changes, I’m always open to feedback and happy to adjust.

api/views/utils.py
```diff
@@ -168,9 +177,14 @@ def get_queryset(request, feed_params, valid_feed_types):
...
         .distinct()
-        .order_by(feed_params.ordering)[: int(feed_params.feed_size)]
     )
 
+    # aggregated endpoints should operate on the full queryset
+    # to compute sums, counts, and other metrics correctly.
+    if not is_aggregated:
+        iocs = iocs.order_by(feed_params.ordering)
+        iocs = iocs[: int(feed_params.feed_size)]
+
```
regulartim: Nice!

api/views/utils.py
```diff
@@ -317,3 +331,74 @@ def is_sha256hash(string: str) -> bool:
...
+    aggregated = (
+        iocs_qs.exclude(asn__isnull=True)
+        .values("asn")
+        .annotate(
+            ioc_count=Count("id", distinct=True),
+            total_attack_count=Sum("attack_count", distinct=True),
+            total_interaction_count=Sum("interaction_count", distinct=True),
+            total_login_attempts=Sum("login_attempts", distinct=True),
+            expected_ioc_count=Sum("recurrence_probability", distinct=True),
+            expected_interactions=Sum("expected_interactions", distinct=True),
```
regulartim: Can you please explain why you are using `distinct=True` here? For the `Count` it does not really do anything, because `id` is unique and for `Sum` I can't find any documentation of what the `distinct` argument actually does.

api/serializers.py
```diff
@@ -118,6 +118,34 @@ def validate_ordering(self, ordering):
...
+            "last_seen",
+        }
+    )
+
+    def validate_ordering(self, ordering):
+        field_name = ordering.lstrip("-").strip()
+
+        if field_name not in self.ALLOWED_ORDERING_FIELDS:
+            raise serializers.ValidationError(
+                {f"Invalid ordering field for ASN aggregated feed: '{field_name}'. Allowed fields: {', '.join(sorted(self.ALLOWED_ORDERING_FIELDS))}"}
```
regulartim: This is a set literal but should be a string, right?

api/views/utils.py
```diff
@@ -317,3 +331,74 @@ def is_sha256hash(string: str) -> bool:
...
+
+    Returns
+        str: A safe ordering string to use directly in the aggregation query.
+    """
+    fallback_fields = fallback_fields or set()
+
+    if not ordering or ordering in fallback_fields:
+        return default
+
+    return ordering
```
regulartim: This is a little overengineered in my opinion. I think instead of this, we can just return `-ioc_count` in the `ASNFeedsOrderingSerializer` if the validation fails. Then, if a user requests a supported ordering, it just works and if not, the results are ordered by `-ioc_count`. Or am I missing something?

api/views/utils.py
```diff
@@ -317,3 +331,74 @@ def is_sha256hash(string: str) -> bool:
...
+    resolved_ordering = resolve_aggregation_ordering(
+        ordering=feed_params.ordering,
+        default="-ioc_count",
+        fallback_fields={"-last_seen"},
+    )
+
+    direction = "-" if resolved_ordering.startswith("-") else ""
+    field = resolved_ordering.lstrip("-").strip()
+
+    aggregated = aggregated.order_by(f"{direction}{field}")
```
regulartim: Also very complicated. Can we also rely on the Serializer here?

api/views/utils.py
drona-gyawali: The default ordering (-ioc_count) cannot be enforced in the serializer because by the time the aggregator runs, feed_params.ordering is already populated (default -last_seen) from FeedRequestParams. So the serializer never sees it as missing.

I acknowledge that using resolve_aggregation_ordering was over-engineered, and I agree that we can simplify it by adding a small override directly in `asn_aggregated_queryset`, like this:

```python
if not ordering or ordering.strip() in {"", "-last_seen"}:
    ordering = "-ioc_count"
```

This keeps things simple while ensuring the default ordering works correctly.  If you’re okay with it, I can push these changes but if you have a better approach, I’d love to apply that instead.

api/views/utils.py
drona-gyawali: Thank you for pointing that out. To be frank, I had used distinct as a quick workaround, but the root cause of the error was row duplication in the SQL query. This happened because of the Many-to-Many join from .filter(general_honeypot__active=True) combined with ArrayAgg on the honeypot field in get_queryset.

This duplication caused Count("id") and other aggregated fields to be inflated (e.g., 4 instead of 2), which led to the test failures. At first, I thought it was related to ordering or feed_size in get_queryset, and even considered a possible test case issue, but the real problem was the join itself.

Screenshot from 2026-01-24 15-30-30

Using distinct=True on Count was only a workaround  it masks the issue but doesn’t fix it properly and can break Sum when values repeat and sorry , I realized this issue later after you pointed out.

The clean solution i think  to separate numeric aggregation from the M2M honeypot aggregation in asn_aggreated_queryset and  little refactor in get_queryset. That way, counts and sums stay accurate without duplication tricks.

The changes in get_queryset would roughly reflect this approach. 
```
iocs = (
        IOC.objects.filter(**query_dict)
        .exclude(ip_reputation__in=feed_params.exclude_reputation)
        .annotate(value=F("name"))
        
        .distinct()
    )

    # aggregated endpoints should operate on the full queryset
    # to compute sums, counts, and other metrics correctly.
    if not is_aggregated:
        iocs= iocs.filter(general_honeypot__active=True)
        iocs = iocs.annotate(honeypots=ArrayAgg("general_honeypot__name"))
        iocs = iocs.order_by(feed_params.ordering)
        iocs = iocs[:int(feed_params.feed_size)]
```

what do you think?

api/views/utils.py
regulartim: Looks good! :)

api/views/utils.py
regulartim: Thanks for the clarification! :) I think I do understand the fundamental problem now. My understanding is that Django can't easily nest a GROUP BY ioc.id (from ArrayAgg) inside a GROUP BY asn. When it tries to merge them, the resulting SQL loses the deduplication. So I think your proposed solution might work. However you need to consider:
1. you also have to filter out inactive honeypots for the aggregation - maybe you can use Django's `Exists` for that (but I am not sure). 
2. be cautious with queries that have ArrayAgg and Sum/Count - right now you still have a ArrayAgg in `asn_aggregated_queryset` which still might cause duplicates

api/serializers.py
drona-gyawali: yes it is now correct i hope. [changes](https://github.com/intelowlproject/GreedyBear/pull/718/changes#diff-1acc662da69290fd8ce4137ddec9d9cc131df53d76d6d107a13c95550781e81e)

api/views/utils.py
drona-gyawali: changes [2e0a](https://github.com/intelowlproject/GreedyBear/pull/718/changes#diff-2e0a114be27761cda8490ac4020115319c4a0282748e18e57c95a745fd77af2eR345-R348)

api/views/utils.py
drona-gyawali: Thanks for the feedback!  I did some changes  hope it matches what you have in mind [2e0a](https://github.com/intelowlproject/GreedyBear/pull/718/changes#diff-2e0a114be27761cda8490ac4020115319c4a0282748e18e57c95a745fd77af2e)

api/views/utils.py
```diff
@@ -317,3 +325,65 @@ def is_sha256hash(string: str) -> bool:
...
+
+    honeypot_agg = (
+        iocs_qs.exclude(asn__isnull=True)
+        .filter(general_honeypot__active=True)
+        .values("asn")
+        .annotate(
+            honeypots=ArrayAgg(
+                "general_honeypot__name",
+                distinct=True,
+                filter=Q(general_honeypot__active=True),
```
regulartim: Isn't this redundant? You already applied this filter in line 368.

api/serializers.py
```diff
@@ -118,6 +118,35 @@ def validate_ordering(self, ordering):
...
+        }
+    )
+
+    def validate_ordering(self, ordering):
+        field_name = ordering.lstrip("-").strip()
+
+        if field_name not in self.ALLOWED_ORDERING_FIELDS:
+            raise serializers.ValidationError(
+                f"""
+                 Invalid ordering field for ASN aggregated feed: '{field_name}'.
```
regulartim: This multi-line f-string might lead to leading whitespace in the error message. Please write it in a single line.

api/views/utils.py
drona-gyawali: you’re right. The filter inside ArrayAgg is redundant since the queryset is already restricted to active honeypots. I’ll remove it and keep the queryset-level filter only.

api/serializers.py
drona-gyawali: You’re right, the multi-line f-string introduces leading whitespace. I’ve rewritten it as a single-line error message.

drona-gyawali:
> Good work! 👍 This basically looks good to merge, I only found two minor things.
> 
> Since this is a new API endpoint, it would be great to include it into our documentation. Would you please do that? If so, do you need guidance for that?

Raised a pr in our docs repo: https://github.com/intelowlproject/docs/pull/46

PR #743, opened 2026-01-28
fix: Respect verbose parameter in feeds API response (Fixes #741)

### LINKED ISSUE #741: Feeds API returns verbose response from Advanced Feeds API
On my instance running the `develop` branch, I get a verbose response when calling `/api/feeds/all/all/recent.json`. Every IoC object returned, has this structure:
```
{
    "asn": null,
    "ip_reputation": "",
    "interaction_count": 45,
    "destination_ports":
    [],
    "days_seen":
    [
        "2026-01-05",
        ...
    ],
    "login_attempts": 0,
    "expected_interactions": 0.0,
    "first_seen": "2026-01-05",
    "attack_count": 45,
    "scanner": false,
    "firehol_categories":
    [],
    "honeypots":
    [
        "Cowrie"
    ],
    "value": "wujie.com.cn_123",
    "recurrence_probability": 0.0,
    "last_seen": "2026-01-28",
    "payload_request": true,
    "feed_type":
    [
        "cowrie"
    ],
    "destination_port_count": 0
}
```

This is unexpected, as the response from this endpoint should contain a lot less fields. Here is the expected response from an older version of the endpoint: 
```
{
    "feed_type":
    [
        "cowrie"
    ],
    "value": "134.122.50.246",
    "scanner": true,
    "payload_request": false,
    "first_seen": "2026-01-05",
    "last_seen": "2026-01-28",
    "attack_count": 16,
    "interaction_count": 1151,
    "ip_reputation": "",
    "asn": 14061,
    "destination_port_count": 1,
    "login_attempts": 223,
    "recurrence_probability": 0.09121786866910975,
    "expected_interactions": 3.8284884306903937
}
```

Sumit-ai-dev:
# Description

Fixed the feeds API to respect the `verbose` parameter properly.

**The Issue:**
The `/api/feeds/` endpoints were always returning all fields (including `days_seen`, `destination_ports`, [honeypots](cci:1://file:///Users/sumitdas/Desktop/GreedyBearAI/GreedyBear/greedybear/tasks.py:15:0-19:32), `firehol_categories`) even when verbose mode wasn't enabled. This made responses unnecessarily large.

**The Fix:**
Split the fields into two groups:
- Base fields (always returned)
- Verbose fields (only when `verbose=true`)

Now the default response is much cleaner, and you only get the extra data when you actually ask for it.

## Related issues
Fixes #741

## Type of change

# Checklist

### DIFF
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -205,13 +205,11 @@ def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose
     Format the IOC data into the requested format (e.g., JSON, CSV, TXT).
 
     Args:
-        request: The incoming request object.
         iocs (QuerySet): The filtered queryset of IOC data.
-        feed_type (str): Type of feed (e.g., log4j, cowrie, etc.).
+        feed_params (FeedRequestParams): Request parameters including format.
         valid_feed_types (frozenset): The set of all valid feed types.
-        format_ (str): Desired format of the response (e.g., json, csv, txt).
         dict_only (bool): Return IOC dictionary instead of Response object.
-        verbose (bool): Include IOC properties that may contain a lot of data.
+        verbose (bool): Include verbose fields (days_seen, destination_ports, honeypots, firehol_categories).
 
     Returns:
         Response: The HTTP response containing formatted IOC data.
@@ -235,7 +233,9 @@ def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose
             )
         case "json":
             json_list = []
-            required_fields = {
+
+            # Base fields always returned
+            base_fields = {
                 "value",
                 "first_seen",
                 "last_seen",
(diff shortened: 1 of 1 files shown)


api/views/utils.py
```diff
@@ -263,9 +270,17 @@ def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose
                     "first_seen": ioc["first_seen"].strftime("%Y-%m-%d"),
                     "last_seen": ioc["last_seen"].strftime("%Y-%m-%d"),
                     "feed_type": ioc_feed_type,
-                    "destination_port_count": len(ioc["destination_ports"]),
```
regulartim, line 266: Why do you remove this? It should stay in the non-verbose response, even if its value is 0 (which is not realistic).

Sumit-ai-dev: You're right, my bad! I'll fix it so destination_port_count stays in the default response. Only the destination_ports array should be verbose-only. Pushing the fix now.

PR #753, opened 2026-01-29
test(migrations): add migrations test . closes #746

### LINKED ISSUE #746: Add migration tests for honeypots and IOC data changes
**Problem**:

Our project  contains a few historical data migrations that modify or remove existing rows rather than only adding schema or defaults. These migrations currently run without errors, but they depend on historical database state and could silently break if related models, constraints, or migrations are modified in the future.

Only the following migrations are proposed for testing, due to their higher risk profile:

1. 0029_remove_hardcoded_honeypots
    - Conditionally deletes pre-seeded honeypots only if no IOC is associated.
    - A logic or relationship change could unintentionally delete in-use data.

2. 0030_migrate_cowrie_log4j
    - Migrates meaning from boolean fields (cowrie, log4j) into a ManyToMany relationship.
    - Incorrect behavior could silently drop honeypot attribution for existing IOCs.

Other migrations mainly insert static defaults or adjust schema and are intentionally left out to keep test scope minimal and focused.

Motivation:

While validating these migrations, it became apparent that re-creating historical honeypot rows is prevented by a case-insensitive uniqueness constraint. This highlights how dependent these migrations are on historical database state and reinforces the value of automated validation to guard against future regressions.

Proposed approach

- Add focused migration tests using django-test-migrations
- Use TransactionTestCase to reflect real migration behavior
- Validate only the expected data outcomes:
- In-use honeypots are preserved
- Unused honeypots are removed
- IOC honeypot relationships are migrated correctly

new deps required 

[django-test-migrations](https://github.com/wemake-services/django-test-migrationsl)

This dependency is test-only and widely used for validating Django data migrations.

drona-gyawali:
# Description

This PR adds focused regression tests for two historical data migrations that modify existing rows and depend on historical database state:

 - 0029_remove_hardcoded_honeypots

- 0030_migrate_cowrie_log4j

These migrations currently run without errors but could silently break if related models, constraints, or relationships are changed in the future.

Please include a summary of the change.

## Related issues
closes #746 

# Checklist

### Important Rules

### DIFF
--- a/.github/workflows/pull_request_automation.yml
+++ b/.github/workflows/pull_request_automation.yml
@@ -87,6 +87,8 @@ jobs:
       upload_coverage: true
       tags_for_slow_tests: main
 
+      custom_command: python manage.py test --tag=migration --failfast
+
       env: >-
         {
         "ENVIRONMENT": "ci",


drona-gyawali:
just realized that ci is currently not installing dev dependencies, so the django-test-migrations module isn’t found. Locally, with dev requirements installed, the tests pass fine. We might need to update the CI to include dev requirements when running tests or some other alternative..

tests/test_migrations.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+
+class TestCowrieLog4jMigration(MigrationTestCase):
+    """Tests migration of cowrie and log4j boolean flags into the GeneralHoneypot M2M relation."""
+
+    migrate_from = "0029_remove_hardcoded_honeypots"
+    migrate_to = "0030_migrate_cowrie_log4j"
+
+    def test_boolean_flags_are_migrated_to_m2m(self):
```
regulartim, line 43: In this test case, can you please also check if the boolean fields were removed from the IOCs?

regulartim (changes requested):
Good work! 👍

regulartim:
I think I fixed the CI issue, but your tests are failing for some reason. Any idea why?

regulartim:
Suggestion from LLM:

This is a PostgreSQL-specific issue with django-test-migrations. The first test (TestRemoveHardcodedHoneypots) creates an IOC and adds it to a GeneralHoneypot via the M2M relation. That M2M operation creates deferred constraint triggers. When the second test's setUp calls migrator.apply_initial_migration, it tries to DROP/TRUNCATE the greedybear_ioc table, and PostgreSQL refuses because those triggers are still pending. The fix is to force PostgreSQL to resolve deferred constraints before migrator.reset() runs. In the MigrationTestCase.tearDown, execute SET CONSTRAINTS ALL IMMEDIATE before the reset:
```python
def tearDown(self):
      from django.db import connection
      with connection.cursor() as cursor:
          cursor.execute("SET CONSTRAINTS ALL IMMEDIATE")
      self.migrator.reset()
      super().tearDown()
```

tests/test_migrations.py
drona-gyawali: thanks for this feedback, I checked the test case with following requirements and  test case failed  because the migration file currently does not change the boolean fields from True to False. They will remain True, so the assertions in the test will fail. I think  the  solution is to refactor the migration to explicitly set these fields to False and save the changes. This way, the test will pass as expected. what do you think?

image

drona-gyawali:
> Suggestion from LLM:
> 
> This is a PostgreSQL-specific issue with django-test-migrations. The first test (TestRemoveHardcodedHoneypots) creates an IOC and adds it to a GeneralHoneypot via the M2M relation. That M2M operation creates deferred constraint triggers. When the second test's setUp calls migrator.apply_initial_migration, it tries to DROP/TRUNCATE the greedybear_ioc table, and PostgreSQL refuses because those triggers are still pending. The fix is to force PostgreSQL to resolve deferred constraints before migrator.reset() runs. In the MigrationTestCase.tearDown, execute SET CONSTRAINTS ALL IMMEDIATE before the reset:
> 
> ```python
> def tearDown(self):
>       from django.db import connection
>       with connection.cursor() as cursor:
>           cursor.execute("SET CONSTRAINTS ALL IMMEDIATE")
>       self.migrator.reset()
>       super().tearDown()
> ```

Thanks for the suggestion! I will apply the following snippet in tearDown and push the changes. Hopefully, this resolves the CI issue.

tests/test_migrations.py
regulartim: Ah, I was thinking that migration 30 would remove the boolean fields completely, but that happens in migration 31. Sorry for the confusion.

drona-gyawali: i will remove all changes that i applied , to prove this

regulartim:
> Locally, with dev requirements installed, the tests pass fine.

Now the CI also installs the dev requirements but your tests still don't pass. I tried them locally on my machine and they also fail. Do you know why? Do they still pass on your machine?

drona-gyawali:
> > Locally, with dev requirements installed, the tests pass fine.
> 
> Now the CI also installs the dev requirements but your tests still don't pass. I tried them locally on my machine and they also fail. Do you know why? Do they still pass on your machine?

 I'm actively working on  this and trying to solve it. I discovered that the issue seems to be related to test isolation in the DB level. When I run only the migration test file in isolation (e.g., python manage.py test tests.test_migrations), it passes successfully. I initially thought it would work in the full test suite as well, so I mentioned it passed, but in reality, it fails when running the entire suite due to some DB state pollution or pending constraint/trigger events from other tests. I'm investigating further to identify the exact cause, and if I find a fix or more details, I'll update you right away.

drona-gyawali:
while doing some investigation on this bug what  I found is  postgresql has a feature called "deferred constraints" where foreign key checks can be postponed until the end of a transaction. Django creates many-to-many (M2M) relationship tables with these deferred constraints by default. 

proof of deferred constraint  in db:

Screenshot from 2026-01-31 16-37-30

At First, i  thought that error was a migration base class itself but, later found something interesting.

 During normal tests : 
  - Tests create IOC objects and use M2M relationships: ioc.general_honeypot.add(honeypot).
  - PostgreSQL says "I'll check this foreign key constraint later" and queues it as a "pending trigger event".
  - The test completes, but PostgreSQL still has these pending checks in its queue.
  
When migration test came:
 - The django-test-migrations library needs to DROP or TRUNCATE tables to test migrations from scratch
 
- Postgresql says: "I can't drop/truncate greedybear_ioc table because I still have pending trigger events (deferred     constraint checks) waiting"
 
- ERROR: cannot DROP TABLE "greedybear_ioc" because it has pending trigger events

 - django-test-migrations assumes it can flush all tables at any point in the test, but Postgres blocks flushes when deferred triggers exist.

What proves our code is not problem:
- When we run only migration tests, no other tests have created pending trigger events
The table can be dropped clean and test case passes.


This is my first time dealing with this kind of PostgreSQL behavior, so I may be wrong, but my understanding is that things should ideally work with the normal Django test setup without needing manual SQL or cursor-based fixes.

In this case, making the migration test base class more complex with DB-specific queries feels heavy  and harder to maintain.

As of now, running the migration tests in isolation seems like the cleanest solution .
I’d really like to hear your thoughts on this and whether you think  i am  missing something important here.

regulartim:
Hey @drona-gyawali ! Thanks for your investigation. 

> I’d really like to hear your thoughts on this and whether you think i am missing something important here.

You're not missing anything that is obvious to me at least. What are the options that we have now, in you opinion?

drona-gyawali:
> Hey @drona-gyawali ! Thanks for your investigation.
> 
> > I’d really like to hear your thoughts on this and whether you think i am missing something important here.
> 
> You're not missing anything that is obvious to me at least. What are the options that we have now, in you opinion?

Thank you very much for taking the time to review this PR.

 I think that the case we have in our project isn’t solvable purely at the programmatic level. Interfering with the schema or forcing PostgreSQL to ignore deferred constraints can lead to bigger problems and break other tests.

My simple and maintainable solution is to isolate migration tests using tags:

* We can mark migration tests with `@tag("migration")`.
* We can configure a custom `test_runner` so that:

  * When a developer runs `python manage.py test`, normal tests run (e.g., 423 test cases)  excluding migration tests.
  * If a developer wants to run migration tests, they can run `python manage.py test --tag=migration`.

This approach keeps the developer experience clean, avoids DB conflicts, and ensures that future schema changes won’t accidentally break normal tests.

Additionally, this separation makes it easier to configure CI pipelines, because normal tests and migration tests can be run independently, keeping CI fast and reliable while still allowing migration tests to run when needed.

As of now, this seems like the practical and safe solution given  db deferred constraints behavior. and I really want to hear your thoughts on this approach.

regulartim:
Thanks for the explanation! :) The problem I have with your approach is that the migration tests won't run by default, reducing their value as regression guards. But if you are convinced that there is no better option, feel free to implement it that way.

drona-gyawali:
> The problem I have with your approach is that the migration tests won't run by default, reducing their value as regression guards.

I completely agree with your point. For now, we can run a separate CI job for these test cases as a regression guard if that work. I assure you, In the future, if I find a reliable solution, workaround, or any relevant update, I’ll implement it or notify you :)

Issue #770, opened 2026-02-04
AS API does not diplay the name of the AS


regulartim:
In our new AS API, we only show the number of the AS, but not its actual name. Although the ASN is the actual identifier of the AS, adding the name to the responses would be useful.

Sumit-ai-dev:
Hi @regulartim,

I know my previous contributions didn't meet your standards, and I sincerely apologize for that.

I've spent a significant amount of time working with GreedyBear, and honestly, I regret that my earlier work wasn't up to par. This project means a lot to me, and I hate that I let you down. I'd really appreciate a chance to work on this issue and demonstrate that I've improved.

Here's my proposed approach:

## The Problem

ASN API returns only numbers, not names. Users have to look up "ASN 15169" to know it's Google.

**Current:** `{"asn": 15169, "ioc_count": 42}`  
**Better:** `{"asn": 15169, "as_name": "GOOGLE", "ioc_count": 42}`

## Proposed Solution

**Changes needed:**

1. **Add field** ([greedybear/models.py](cci:7://file:///Users/sumitdas/Desktop/GreedyBearAI/GreedyBear/greedybear/models.py:0:0-0:0)): `as_name = models.CharField(max_length=256, blank=True)`
2. **Extract from ElasticSearch** ([greedybear/cronjobs/extraction/utils.py](cci:7://file:///Users/sumitdas/Desktop/GreedyBearAI/GreedyBear/greedybear/cronjobs/extraction/utils.py:0:0-0:0)): `as_name=hits[0].get("geoip", {}).get("as_organization", "")`
3. **Update on merge** ([greedybear/cronjobs/extraction/ioc_processor.py](cci:7://file:///Users/sumitdas/Desktop/GreedyBearAI/GreedyBear/greedybear/cronjobs/extraction/ioc_processor.py:0:0-0:0)): `existing.as_name = new.as_name`
4. **Include in API** ([api/views/utils.py](cci:7://file:///Users/sumitdas/Desktop/GreedyBearAI/GreedyBear/api/views/utils.py:0:0-0:0)): `.annotate(as_name=Max("as_name"), ...)`
5. **Migration** to add database column
6. **Comprehensive tests** (8 test cases covering all scenarios)

**Impact:**  
Before: `ASN 15169: 42 attacks` _(What is 15169?)_  
After: `GOOGLE (15169): 42 attacks` _(Immediately clear!)_

[...]

PR #777, opened 2026-02-04
feat(frontend): Dynamic GreedyBear news widget with filtering, sorting, and improved UX. Closes #762

### LINKED ISSUE #762: News feeds on start page are not generated dynamically
The GreedyBear news feeds displayed on the start page are currently hardcoded rather than being generated dynamically. It would be great if new content would appear automatically without code changes.

drona-gyawali:
# Description

This PR replaces the hardcoded news in Home.jsx with a dynamic NewsWidget component that fetches news from public/news.json. It filters, sorts, and displays only news related to GreedyBear, while preserving the original design.

Why news.json includes a project field and Id?
 - To ensure only GreedyBear news appears on the homepage.
 - Allows the JSON to include news for multiple projects in the future.
 - Filtering logic ensures future-proof dynamic loading without modifying code.
 -  React requires a unique key for each item when rendering a list.

## Related issues

closes:  #762 

## Type of change

# Checklist

- [x] If the GUI has been modified:
    - [x] I have a provided a screenshot of the result in the PR.
  
### Important Rules

### DIFF
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -4,10 +4,10 @@ import csv
 import logging
 import re
 from datetime import datetime, timedelta
-from email.utils import parsedate_to_datetime
 from ipaddress import ip_address
 
 import feedparser
+import requests
 from django.conf import settings
 from django.contrib.postgres.aggregates import ArrayAgg
 from django.core.cache import cache
@@ -420,43 +420,37 @@ def get_greedybear_news() -> list[dict]:
         return cached
 
     try:
-        feed = feedparser.parse(RSS_FEED_URL)
+        response = requests.get(RSS_FEED_URL, timeout=5)
+        response.raise_for_status()
+        feed = feedparser.parse(response.content)
+
+        filtered_entries = sorted(
+            [entry for entry in feed.entries if "greedybear" in entry.get("title", "").lower() and entry.get("published_parsed")],
+            key=lambda e: e.published_parsed,
+            reverse=True,
+        )
+
         news_items: list[dict] = []
-
-        for entry in feed.entries:
-            title = entry.get("title", "")
-            if "greedybear" not in title.lower():
-                continue
-
+        for entry in filtered_entries:
             summary = entry.get("summary", "").strip().replace("\n", " ")
-            pub_date = entry.get("published")
-
-            if not pub_date:
-                continue
 
             subtext = summary[:180].rsplit(" ", 1)[0] + "..." if len(summary) > 180 else summary
 
             news_items.append(
                 {
-                    "title": title,
-                    "date": pub_date,
+                    "title": entry.get("title"),
+                    "date": entry.get("published"),
                     "link": entry.get("link"),
                     "subtext": subtext,
                 }
             )
-
-        # RSS feeds are typically chronological, but we sort explicitly
-        # to ensure consistency since the data source is external.
-        news_items.sort(
-            key=lambda x: parsedate_to_datetime(x["date"]),
-            reverse=True,
-        )
 
         cache.set(
             CACHE_KEY_GREEDYBEAR_NEWS,
             news_items,
             CACHE_TIMEOUT_SECONDS,
         )
+
         return news_items
 
     except Exception as exc:
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1,6 +1,6 @@
-from email.utils import parsedate_to_datetime
 from unittest.mock import patch
 
+import requests
 from django.conf import settings
 from django.core.cache import cache
 from django.test import override_settings
@@ -904,18 +904,21 @@ class NewsTestCase(CustomTestCase):
                     title="IntelOwl Update",
                     summary="intelowl news",
                     published="Wed, 01 Jan 2026 00:00:00 GMT",
+                    published_parsed=(2026, 1, 1, 0, 0, 0, 2, 1, 0),
                     link="https://example.com/1",
                 ),
                 FeedParserDict(
                     title="GreedyBear v3 Release",
                     summary="greedybear release notes",
                     published="Thu, 29 Jan 2026 00:00:00 GMT",
+                    published_parsed=(2026, 1, 29, 0, 0, 0, 3, 29, 0),
                     link="https://example.com/2",
                 ),
                 FeedParserDict(
                     title="IntelOwl Improvements",
                     summary="Not related to GreedyBear",
                     published="Mon, 01 Sep 2025 00:00:00 GMT",
+                    published_parsed=(2025, 9, 1, 0, 0, 0, 0, 244, 0),
                     link="https://example.com/3",
                 ),
             ]
@@ -933,12 +936,14 @@ class NewsTestCase(CustomTestCase):
                     title="GreedyBear Old",
                     summary="old post",
                     published="Wed, 01 Jan 2026 00:00:00 GMT",
+                    published_parsed=(2026, 1, 1, 0, 0, 0, 0, 0, 0),
                     link="https://example.com/old",
                 ),
                 FeedParserDict(
                     title="GreedyBear New",
                     summary="new post",
                     published="Thu, 29 Jan 2026 00:00:00 GMT",
+                    published_parsed=(2026, 1, 29, 0, 0, 0, 0, 0, 0),
                     link="https://example.com/new",
                 ),
             ]
@@ -949,12 +954,6 @@ class NewsTestCase(CustomTestCase):
         self.assertEqual(result[0]["title"], "GreedyBear New")
         self.assertEqual(result[1]["title"], "GreedyBear Old")
 
-        # verifying sorting by date
-        self.assertGreater(
-            parsedate_to_datetime(result[0]["date"]),
-            parsedate_to_datetime(result[1]["date"]),
-        )
-
     @patch("api.views.utils.feedparser.parse")
     def test_truncates_long_summary(self, mock_parse):
         long_summary = "word " * 100
@@ -964,6 +963,7 @@ class NewsTestCase(CustomTestCase):
                     title="GreedyBear Long Post",
                     summary=long_summary,
                     published="Thu, 29 Jan 2026 00:00:00 GMT",
+                    published_parsed=(2026, 1, 29, 0, 0, 0, 3, 29, 0),
                     link="https://example.com",
                 )
             ]
@@ -1003,6 +1003,7 @@ class NewsTestCase(CustomTestCase):
                     title="GreedyBear Cached Test",
                     summary="cache test",
                     published="Thu, 29 Jan 2026 00:00:00 GMT",
+                    published_parsed=(2026, 1, 29, 0, 0, 0, 3, 29, 0),
                     link="https://example.com",
                 )
             ]
@@ -1019,3 +1020,11 @@ class NewsTestCase(CustomTestCase):
         result2 = get_greedybear_news()
         self.assertEqual(result1, result2)
         mock_parse.assert_not_called()
+
+    @patch("api.views.utils.requests.get")
+    def test_feed_request_timeout_returns_empty_list(self, mock_get):
+        mock_get.side_effect = requests.Timeout()
+
+        result = get_greedybear_news()
+
+        self.assertEqual(result, [])


regulartim (changes requested):
Hey @drona-gyawali ! :) Sorry, but I think I have not been precise enough when I wrote the issue. You created a local file that contains the links to the blog entries. But that's not what I actually want. 
All links are pointing to the IntelOwl blog https://intelowlproject.github.io/blogs . I would like to have a news widget that dynamically retrieves the links to GreedyBear related blog entries. Such that when a new article is published there, it should automatically appear on the GreedyBear start page. Do you understand what I mean?

If you don't feel like solving this issue anymore under that circumstances, that is fine, just tell me. 

(You linked the wrong issue in this PR. This PR should close #762 not #770 )

drona-gyawali:
> Hey @drona-gyawali ! :) Sorry, but I think I have not been precise enough when I wrote the issue. You created a local file that contains the links to the blog entries. But that's not what I actually want. All links are pointing to the IntelOwl blog https://intelowlproject.github.io/blogs . I would like to have a news widget that dynamically retrieves the links to GreedyBear related blog entries. Such that when a new article is published there, it should automatically appear on the GreedyBear start page. Do you understand what I mean?
> 
> If you don't feel like solving this issue anymore under that circumstances, that is fine, just tell me.
> 
> (You linked the wrong issue in this PR. This PR should close #762 not #770 )


Extremely sorry for the confusion and for misunderstanding your request the first time.
I really appreciate your patience and the clear explanation you gave it helped me a lot.

To make sure I’m on the right track now:

What you want for the GreedyBear homepage news widget is:

 - Truly automatic / dynamic news: Whenever a new blog post appears on [IntelOwl blogs](https://intelowlproject.github.io/blogs)
 that is about GreedyBear, it should show up in the widget without anyone having to edit any file in the GreedyBear repo (no JSON file, no code change, nothing manual).

- The source of truth should be the live blog page itself (the one built from the .md files in the intelowlproject.github.io repo).

Only show posts clearly related to GreedyBear (like the ones with "GreedyBear" in the title: v3 coming, v2.0 released, improvements to GreedyBear, presenting GreedyBear, etc.).

I completely get why the news.json approach wasn’t ideal  it’s still semi-manual.

I understand the requirement and will push the new version soon.

regulartim:
No worries, my issue describtion was too vague. Sorry for that. Yes, what you describe is exactly what I meant. If it is too hard to retrieve the list of blog posts from the IntelOwl website, you could also consider fetching them directly from the repo.

drona-gyawali:
Hi @regulartim, I want to make sure I’m fully aligned before continuing implementation. So i did  some research on possible implementations and wanted to check alignment before coding further.

Thinking by  a long term perspective,  Frontend or backend HTML scraping/dom parsing feels brittle and hard to maintain also limitation of sorting by date.

A more reliable approach seems to be fetching GreedyBear-related blog posts directly from the IntelOwl GitHub repo via the GitHub API, parsing markdown with frontmatter on the backend, and exposing a small JSON endpoint for the frontend. This avoids dependency on the blog’s HTML structure and allows proper sorting/filtering and pagiantion and caching in future.


One small limitations, there’s a GitHub API rate limit for unauthenticated calls(only 60 req perhour), but this can be mitigated via Inmemory caching or an access token(5000 req per hour).

back-end response will look like this :
```
{
  "news": [
    {
      "title": "GreedyBear version 3 coming",
      "date": "2026-01-29",
      "link": "https://intelowlproject.github.io/blogs/greedybear_v3_release",
      "subtext": "Over the last months some new contributors helped us to implement a lot of new stuff in GreedyBear..."
    },

  ],
  "count": 1
}
```

what do you think ? does this approach sound acceptable to you, or would you prefer a different direction? And thank you very much for your valuable time on this pr.

regulartim:
> Thinking by a long term perspective, Frontend or backend HTML scraping/dom parsing feels brittle and hard to maintain also limitation of sorting by date.

The sorting is not that important to me, but I agree, HTML scraping is not optimal. I was hoping that github.io has some kind of API where you can just retrieve a list of posts.

> One small limitations, there’s a GitHub API rate limit for unauthenticated calls(only 60 req perhour), but this can be mitigated via Inmemory caching or an access token(5000 req per hour).

Caching sounds good. If we cache for one hour, we would only have to make one request per hour per instance. Correct?

> what do you think ? does this approach sound acceptable to you, or would you prefer a different direction? And thank you very much for your valuable time on this pr.

As far as I can tell, this sounds very solid! :) Thank you for the time you spend investigating!

drona-gyawali:
Hi,  In this version, I’ve reworked the implementation to align with the original intent of the issue, making the news widget fully automatic and truly dynamic.


The code in utils.py responsible for fetching news might look a bit unpolished because I prioritized keeping the logic within a single function. I’m happy to refactor it if you have suggestions on how you'd like it structured.


I have to  mention about  7-second latency that i notice in our api: 

 One GitHub API request

   - To list all files under Blogs/

Multiple sequential network requests

  - For each markdown file, we parse the raw .md content via download_url
  - Each request is a separate HTTPS round-trip to raw.githubusercontent.com

Network latency dominates  here Each request takes ~200–400 ms depending on network conditions With ~15–20 markdown files, this adds up quickly

Example:
```
1 list request        ~300 ms

15 markdown  ~15 × 300 ms ≈ 4.5–6 s

frontmatter parsing  ~negligible
```
Latency Benchmarks(approx):

    - Cold Start (Uncached): ~7s
    - Warm Start (Cached): 70ms – 162ms

However, Our  implementation has a room for scalability.  If our blog content grows or our user base increases, we can easily transition to a cron job model. This would handle the 'heavy lifting' (the 7s fetch) in the background every hour, ensuring that every user including the first one always receives a near-instant cached result.

Frontend 

Screenshot from 2026-02-06 22-37-38


Thank you for the review!

api/views/utils.py
```diff
@@ -401,3 +405,75 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
...
+    try:
+        # fetching the list of blog post files through github Api
+        response = requests.get(GITHUB_BLOG_API, timeout=10)
+        response.raise_for_status()
+        files = response.json()
+
+        news_items = []
+
+        # processing each markdown file to extract details
+        for file in files[:MAX_FILES_TO_CHECK]:
```
regulartim: What do you think of filtering by filename pattern first here? Assume all GreedyBear related articles have the string "greedybear" in their name.

api/views/news.py
```diff
@@ -0,0 +1,25 @@
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from api.views.utils import get_greedybear_news
+
+
+@api_view(["GET"])
+def news_view(request):
```
regulartim, line 8: This API endpoint is for internal use only. Is there a decorator that can reflect this? Such that only our frontend can access the endpoint?

regulartim (changes requested):
Hey @drona-gyawali ! Thanks a lot for your work and your assessment of the situation. I tested your code, and the widget looks very nice.

When I was looking at your code, another idea came to my mind which I want to discuss with you. We could also make some changes to the website https://intelowlproject.github.io . What if we would generate a RSS feed there and expose it as a `feed.xml` file to the internet. Then, instead of using the github API, we could just load the feed, which would contain all information that we need. Would this be the better approach to solve our issue here?

drona-gyawali:
> When I was looking at your code, another idea came to my mind which I want to discuss with you. We could also make some changes to the website https://intelowlproject.github.io . What if we would generate a RSS feed there and expose it as a `feed.xml` file to the internet. Then, instead of using the github API, we could just load the feed, which would contain all information that we need. Would this be the better approach to solve our issue here?

Yes, using a feed.xml would be a great way to solve this. Having a single file to consume is much simpler and 10 times better than the current approach and would avoid the GitHub API entirely.

Even with this approach, caching would still be important yes, since fetching and parsing XML on every request can still be relatively expensive.

 I also saw that you’ve already opened a PR in another  repo once that gets merged, I’m happy to proceed with updating the implementation accordingly.

regulartim:
Hi @drona-gyawali ! Good news, the PR got merged. :) Here is the link to the RSS feed: https://intelowlproject.github.io/feed.xml

api/views/news.py
drona-gyawali: While doing some research, I found that Django or DRF as such  do not provide a built-in decorator to restrict an API to only our frontend. One possible approach is to send a token from the frontend and verify it on the backend; however, since this token is visible in developer tools, it cannot fully secure a public API.

For now, given that this is a public API, the most practical approach is to monitor its usage. If we observe frequent bot-like activity, we can implement rate limiting or block suspicious requests in the future.  And API isn’t documented, most users won’t find it, so casual exposure is limited.

A similar discussion on this topic can be found [here on Stack Overflow](https://stackoverflow.com/questions/47298148/how-to-ensure-that-only-my-frontend-can-access-my-api-and-not-anyone-else)
. Overall, while no solution is 100% secure for a public API . What do you think?

regulartim: Thanks for your investigation! :) In that case, just leave it as it is. Requests to this API should not be very resource intensive anyway, so I guess it is fine.

api/views/utils.py
```diff
@@ -401,3 +405,63 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
...
+                    "date": pub_date,
+                    "link": entry.get("link"),
+                    "subtext": subtext,
+                }
+            )
+
+        # RSS feeds are typically chronological, but we sort explicitly
+        # to ensure consistency since the data source is external.
+        news_items.sort(
+            key=lambda x: parsedate_to_datetime(x["date"]),
```
regulartim: Feedparser already parses dates, every entry has a published_parsed attribute. I would prefer if you would use that instead. Then you can safely remove the (surprising) import from `email.utils`.

tests/test_views.py
```diff
@@ -865,3 +870,152 @@ def test_is_sha256hash_invalid(self):
...
+        result = get_greedybear_news()
+        self.assertEqual(len(result), 2)
+        self.assertEqual(result[0]["title"], "GreedyBear New")
+        self.assertEqual(result[1]["title"], "GreedyBear Old")
+
+        # verifying sorting by date
+        self.assertGreater(
+            parsedate_to_datetime(result[0]["date"]),
+            parsedate_to_datetime(result[1]["date"]),
+        )
```
regulartim: Same here maybe?

api/views/utils.py
```diff
@@ -401,3 +405,63 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
...
+    Returns:
+        List of dicts with keys: title, date, link, subtext
+        Sorted newest first, or empty list on failure.
+    """
+    cached = cache.get(CACHE_KEY_GREEDYBEAR_NEWS)
+    if cached is not None:
+        return cached
+
+    try:
+        feed = feedparser.parse(RSS_FEED_URL)
```
regulartim: Here we are reading an external resource. What happens if it is not available for some reason? 

To my understanding, feedparser does not time out at any point. Maye it is safer to use `requests.get` with a timeout for fetching the contents and feedparser to parse the result?

Please also add a test for this.

drona-gyawali: Yes, you are right feedparser uses urllib under the hood for http requests, which doesnot enforce timeout.  So I’ve updated the implementation to fetch the feed via requests.get with a timeout and pass the response content to feedparser. I also added a test that verifies we gracefully handle request timeouts by returning an empty list.

tests/test_views.py
drona-gyawali: I’ve updated the test to align with the new implementation. Since sorting now relies on published_parsed, the test simply verifies the order of results rather than re-parsing dates.

api/views/utils.py
drona-gyawali: I’ve updated the implementation to sort entries using published_parsed from feedparser and then build the API response using the human-readable published field. This removes the need for manual date parsing and keeps the response clean.

drona-gyawali:
I noticed that test_views.py has already exceeded 1000 lines and is likely to grow further. To keep our tests maintainable and easier to navigate, I’m thinking of splitting it into separate test files (e.g., test_news.py). Should I open an issue and raise a PR for this in the next workflow?

regulartim:
> Should I open an issue and raise a PR for this in the next workflow?

Yes, please! :)

regulartim (approved):
Good work! 👍

drona-gyawali:
> Good work! 👍

Thanks! I’m really glad to hear that. I enjoyed working on this and I'm happy with how it turned out

PR #784, opened 2026-02-05
Track IoC-Sensor relationship. Closes #779

### LINKED ISSUE #779: Relationship between IoC and Sensor not tracked
Currently there's no way to track which sensor detected which IoC. But especially in scenario with multiple sensors, this information would be quite useful.

opbot-xd:
# Description

This PR implements sensor tracking for Indicators of Compromise (IoCs) by adding a many-to-many relationship between IOC and Sensor models. This enables tracking which sensors detected which IoCs, providing better visibility into threat detection sources.

**Database Changes:**
- Added `sensors` ManyToManyField to IOC model
- Created migration `0036_add_sensors_to_ioc.py`

**Repository Layer:**
- Refactored `SensorRepository`: renamed `add_sensor()` to `get_or_create_sensor()`, changed cache from set to dict
- Added `add_sensor_to_ioc()` method to `IOCRepository` for associating sensors with IoCs

**Extraction Pipeline:**
- Modified `IOCProcessor.add_ioc()` to accept optional sensor parameter
- Updated `ExtractionPipeline` to capture sensor from Elasticsearch hits (`t-pot_ip_ext` field)
- Modified `iocs_from_hits()` utility to return tuples of (IOC, sensors list)
- Updated extraction strategies (Generic and Cowrie) to process IoC once with first sensor, then add remaining sensors separately (prevents duplicate counter increments)

**Admin Panel:**
- Added sensor display in IOC admin with `sensor_list` method
- Added `filter_horizontal` for sensors field

## Related issues
Closes #779

## Type of change

# Checklist

### DIFF
--- a/greedybear/migrations/0036_add_sensors_to_ioc.py
+++ b/greedybear/migrations/0036_add_sensors_to_ioc.py
@@ -3,6 +3,40 @@
 from django.db import migrations, models
 
 
+
+def deduplicate_sensors(apps, schema_editor):
+    """
+    Remove duplicate Sensor records directly from the database to facilitate
+    the unique constraint on the 'address' field.
+    For each address with multiple entries, we keep the one with the lowest ID
+    and remove the others.
+    """
+    Sensor = apps.get_model("greedybear", "Sensor")
+    db_alias = schema_editor.connection.alias
+    
+    # Identify duplicate addresses
+    # Grouping by address:
+    from django.db.models import Count
+    
+    duplicates = (
+        Sensor.objects.using(db_alias)
+        .values("address")
+        .annotate(count=Count("id"))
+        .filter(count__gt=1)
+    )
+
+    for entry in duplicates:
+        address = entry["address"]
+        # Get all sensors for this address, ordered by ID (oldest first)
+        sensors = Sensor.objects.using(db_alias).filter(address=address).order_by("id")
+        
+        # Keep the first one (lowest ID), delete the rest
+        first_sensor = sensors.first()
+        if first_sensor:
+            # Delete all other sensors with this address
+            Sensor.objects.using(db_alias).filter(address=address).exclude(id=first_sensor.id).delete()
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
@@ -15,6 +49,7 @@ class Migration(migrations.Migration):
             name='sensors',
             field=models.ManyToManyField(blank=True, to='greedybear.sensor'),
         ),
+        migrations.RunPython(deduplicate_sensors, reverse_code=migrations.RunPython.noop),
         migrations.AlterField(
             model_name='sensor',
             name='address',


greedybear/migrations/0035_add_sensors_to_ioc.py
```diff
@@ -0,0 +1,23 @@
...
+    dependencies = [
+        ('greedybear', '0034_remove_unused_log4pot'),
+    ]
+
+    operations = [
+        migrations.RenameIndex(
+            model_name='torexitnode',
+            new_name='greedybear__ip_addr_6bc095_idx',
+            old_name='greedybear_ip_addr_tor_idx',
+        ),
```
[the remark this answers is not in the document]
opbot-xd: Hi @regulartim what to do about this?
It got autogenerated when I ran makemigrations for the IoC-Sensor feature, Django picked up this orphaned change.
Should I create 2 migration files for this or let it be as it is?

opbot-xd:
### Test Coverage Report

All modified files have comprehensive test coverage:

| File | Coverage | Status |
|------|----------|--------|
| `greedybear/models.py` | **100%** | Full coverage of new `sensors` ManyToManyField |
| `greedybear/cronjobs/repositories/sensor.py` | **100%** | All methods including refactored `get_or_create_sensor()` tested |
| `greedybear/cronjobs/repositories/ioc.py` | **93.67%** | New `add_sensor_to_ioc()` method fully tested |
| `greedybear/cronjobs/extraction/ioc_processor.py` | **97.92%** | Sensor parameter handling covered |
| `greedybear/cronjobs/extraction/pipeline.py` | **100%** | Sensor extraction from Elasticsearch hits tested |
| `greedybear/cronjobs/extraction/strategies/generic.py` | **100%** | IoC-sensor association logic covered |
| `greedybear/cronjobs/extraction/strategies/cowrie.py` | **83.56%** | Core sensor tracking functionality tested |
| `greedybear/admin.py` | **88.64%** | Sensor display in admin panel covered |

**Overall: 92.04% coverage across all modified files**

greedybear/migrations/0035_add_sensors_to_ioc.py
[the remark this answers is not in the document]
regulartim: I made the pending migration and pushed it to develop. Good catch! :+1:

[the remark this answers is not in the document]
regulartim: (This means you probably have to re-make your migration.)

opbot-xd:
Hi @regulartim the PR is ready for review. Please review it when you get time. Thanks :)

greedybear/cronjobs/repositories/ioc.py
```diff
@@ -42,6 +42,23 @@ def add_honeypot_to_ioc(self, honeypot_name: str, ioc: IOC) -> IOC:
             ioc.general_honeypot.add(honeypot)
         return ioc
 
+    def add_sensor_to_ioc(self, sensor: "Sensor", ioc: IOC) -> IOC:
```
regulartim: Can't you just write Sensor instead of "Sensor" here (without the quotes)?

regulartim: Wait, I don't think we need this method at all. Instead of calling `IocRepository.add_sensor_to_ioc`, you can just call `IOC.sensors.add`.  This works because `ManyToManyField.add` is idempotent. Or do you see any other reason why this method is needed?

greedybear/cronjobs/extraction/utils.py
```diff
@@ -87,9 +87,9 @@ def get_firehol_categories(ip: str, extracted_ip) -> list[str]:
     return firehol_categories
 
 
-def iocs_from_hits(hits: list[dict]) -> list[IOC]:
+def iocs_from_hits(hits: list[dict]) -> list[tuple["IOC", list]]:
```
regulartim: If you just assign the list of sensors to `IOC.sensors` in this method, it simplifies things down the line, I think. Then you don't need to return the list of sensors here and only touch it again in `_merge_iocs`. What do you think?

regulartim (changes requested):
Hey @opbot-xd ! Good work as always! :) But we might be able to simplify the process.

greedybear/cronjobs/repositories/ioc.py
opbot-xd: You're absolutely right! `ManyToManyField.add()` is already idempotent and efficient, so the wrapper method was unnecessary overhead.
I've refactored the code to remove `add_sensor_to_ioc` entirely from the repository, use `ioc.sensors.add(sensor)` directly in the strategies calling code.
Thanks for the catch!

greedybear/cronjobs/extraction/utils.py
opbot-xd: Great suggestion! I've refactored to keep the return type as `list[IOC]` instead of tuples.
Since we can't directly assign to `IOC.sensors` (unsaved objects can't use ManyToMany), I attach sensors via a temporary `_sensors_to_add` attribute on the IOC. Then:
- **For new IOCs**: sensors are added right after save
- **For existing IOCs**: sensors are added in [_merge_iocs()](cci:1://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/cronjobs/extraction/ioc_processor.py:80:4-106:23) (since the existing record is already saved)
This keeps all merge logic consolidated in one place as you suggested.

greedybear/cronjobs/repositories/sensor.py
```diff
@@ -10,49 +10,39 @@ class SensorRepository:
...
+            return self.cache[ip]
         if get_ioc_type(ip) != IP:
             self.log.debug(f"{ip} is not an IP address - won't add as a sensor")
-            return False
+            return None
         sensor = Sensor(address=ip)
         sensor.save()
-        self.cache.add(ip)
+        self.cache[ip] = sensor
         self.log.info(f"added sensor {ip} to the database")
```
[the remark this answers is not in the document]
opbot-xd: Copilot raised a valid point here about potential duplicate [Sensor](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/models.py:21:0-25:27) records in multi-process/concurrent Celery environments.

**The issue**: On a cache miss, the code creates a new Sensor. In multi-worker setups, two workers could simultaneously create sensors for the same IP since the in-memory cache isn't shared.

**Proposed fixes**:
1. Use `Sensor.objects.get_or_create(address=ip)` for atomic creation
2. Add `unique=True` constraint to `Sensor.address` in the model

**My confusion**: Should `Sensor.address` be unique? Logically it makes sense (each sensor IP should exist only once), but I want to confirm before adding a database constraint + migration.

@regulartim What do you think - should I add the unique constraint, or just use [get_or_create](cci:1://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/cronjobs/repositories/sensor.py:21:4-42:21) as a simple fix?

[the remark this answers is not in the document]
regulartim: I agree. Currently we don't have a multi-worker setup, but that can happen in the future. 

> Should Sensor.address be unique?

Yes. 

> What do you think - should I add the unique constraint, or just use get_or_create as a simple fix?

Maybe both!? The unique constraint is the right thing to do on the database level. But then we still need to catch the described edge case on the application level, which can be done with `get_or_create`. Do you agree?

tests/test_ioc_processor.py
regulartim: This should be `self.mock_sensor_repo.cache`. Please check the whole file, there are more occurrences.

greedybear/cronjobs/extraction/utils.py
```diff
@@ -123,6 +128,9 @@ def iocs_from_hits(hits: list[dict]) -> list[IOC]:
             login_attempts=len(hits) if hits[0].get("type", "") == "Heralding" else 0,
             firehol_categories=firehol_categories,
         )
+        # Attach sensors to temporary attribute for later processing
+        ioc._sensors_to_add = sensors
+
```
regulartim: Why do you create this temporary attribute here? Is there a reason to not use `ioc.sensors`?

tests/test_ioc_processor.py
opbot-xd: Fixed the attribute name in the tests. Replaced all occurrences of `self.mock_sensor_repo.sensors` with `self.mock_sensor_repo.cache` (and initialized it as a dict `{}` instead of a set) to match the actual `SensorRepository` implementation.

greedybear/cronjobs/extraction/utils.py
opbot-xd: **Reasoning**: We attach sensors to this temporary attribute because we cannot use `ioc.sensors.add()` on the `IOC` object at that point, as the instance has not yet been saved to the database (Django requires an ID for Many-to-Many relationships). The sensors are then retrieved and added in `IocProcessor` immediately after the IOC is saved or retrieved.

Will add a comment about it.

regulartim: Ah, I see. Did not think about that. Can such an temporary attribute have any unwanted side effects?

opbot-xd: No, it won't have any side effects in my opinion. Since _sensors_to_add is not defined as a field in the Django model, the ORM completely ignores it during save() operations. It exists purely in memory for this specific instance and is discarded once the object goes out of scope.

greedybear/migrations/0036_add_sensors_to_ioc.py
```diff
@@ -0,0 +1,23 @@
...
+        migrations.AddField(
+            model_name='ioc',
+            name='sensors',
+            field=models.ManyToManyField(blank=True, to='greedybear.sensor'),
+        ),
+        migrations.AlterField(
+            model_name='sensor',
+            name='address',
+            field=models.CharField(max_length=15, unique=True),
+        ),
```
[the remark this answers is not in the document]
regulartim: This is a valid point by Copilot, what do you think?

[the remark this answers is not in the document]
opbot-xd: Agreed, that is a very valid point.

I have updated the migration file to include a **data migration step** (`deduplicate_sensors`) that runs **before** the `AlterField` operation. 

This step identifies all `Sensor` records with duplicate addresses. Keeps the oldest record (lowest ID), Deletes the duplicates (safely, by ID).

This ensures the `unique=True` constraint can be applied safely to existing databases without raising an `IntegrityError`.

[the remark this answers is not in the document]
opbot-xd: changes are in this commit `3e0becc4c4a689bdffef3a56508416045cf1aaf5`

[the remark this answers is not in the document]
regulartim: Sorry, I did not see that! 👍

PR #725, opened 2026-01-21
Remove hardcoded Cowrie and Log4pot fields from IOC model. Closes #637

opbot-xd:
# Description

Refactored the [IOC](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/models.py:53:0-83:24) model to remove the legacy hardcoded [cowrie](cci:1://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/tests/test_models.py:40:4-49:77) and [log4j](cci:1://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/tests/api/test_feed_types.py:52:4-63:49) boolean fields. These honeypots are now treated as standard entries in the [GeneralHoneypot](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/models.py:28:0-36:24) table and associated via the [general_honeypot](cci:1://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/api/views/general_honeypot.py:13:0-35:30) ManyToMany relationship. This unifies the data model, simplifies queries, and cleans up the codebase.

Changes include:
- **Database Migration**: Created a data migration to ensure [Cowrie](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/models.py:103:0-125:76) and [Log4pot](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/cronjobs/extraction/strategies/log4pot.py:14:0-149:25) exist as [GeneralHoneypot](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/models.py:28:0-36:24) entries and migrated existing [IOC](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/models.py:53:0-83:24) data to the M2M relationship.
[...]

PR #803, opened 2026-02-12
feat(test): Spliting test_view file for better test structure. Closes #794

### LINKED ISSUE #794: Split test_views.py into Smaller, View-Specific Test Files
**Problem**: 
 - Currently test_views.py contains 1000+ loc.

**Solution**: 
 - Split `test_views.py` into multiple smaller test files

* `test_enrichment_view.py` – Tests for enrichment endpoints
* `test_feeds_view.py` – Tests for feeds (basic + pagination + filters)
* `test_feeds_advanced_view.py` – Tests for advanced feeds
* `test_feeds_asn_view.py` – Tests for ASN aggregated feeds
* `test_statistics_view.py` – Tests for statistics endpoints
* `test_general_honeypot_view.py` – Tests for general honeypot endpoints
* `test_command_sequence_view.py` – Tests for command sequence endpoints
* `test_cowrie_session_view.py` – Tests for cowrie session endpoints
* `test_validation_helpers.py` – Tests for helper/validation utilities
* `test_news_view.py` – Tests for GreedyBear news & caching logic

drona-gyawali:
# Description

In this pr we split test_views.py testcases

## Related issues
closesL #794 

## Type of change

# Checklist

### Important Rules

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -182,10 +182,6 @@ class CustomTestCase(TestCase):
         except User.DoesNotExist:
             cls.regular_user = User.objects.create_user(username="regular", email="regular@greedybear.com", password="regular")
 
-    @classmethod
-    def tearDownClass(cls):
-        super().tearDownClass()
-
 
 class ExtractionTestCase(CustomTestCase):
     def setUp(self):


drona-gyawali:
To my surprise, noticed  that `test_feed_types.py` was never actually running  neither in CI(maybe) nor during normal test execution. The reason was that there was no `__init__.py` file inside the `api` directory, so Django did not treat it as a proper package and therefore did not discover and execute the tests inside it.

While splitting `test_views` and creating a new `views` directory under `api`, I added an `__init__.py` file to ensure proper test discovery. At that point, the tests started failing with a `database connection already closed` error. Initially, I was confused because I followed the same structure used in the `cronjobs` test directory, yet it was still failing.

After investigating further, I realized this was actually a long-standing issue. The failure had been hidden because `test_feed_types.py` was never executed before. Since it wasn’t running, the underlying error never surfaced.

The root cause turned out to be the overridden `tearDownClass` method in `tests/__init__.py`. In Django, when  we typically use overriding lifecycle methods like `tearDownClass`, we must call `super().tearDownClass()` to allow Django to complete its internal cleanup process. That call was missing. As a result, Django’s internal test database lifecycle was being interrupted, which eventually led to the “connection already closed” error.

image

After adding `super().tearDownClass()`, everything passed successfully.

I just wanted to clarify two things and wanted you guidance:

1. Was `test_feed_types.py` intentionally left out of test discovery?
2. Was there a specific reason for not calling `super().tearDownClass()` in the overridden method

regulartim:
Good catch @drona-gyawali !

> Was test_feed_types.py intentionally left out of test discovery?

No, I don't think so. This was only merged recently in #725 and I don't remember we left these tests out on purpose.

> Was there a specific reason for not calling super().tearDownClass() in the overridden method

I also don't think so. As far as I can see, calling the super method here is essential. The Django docs specifically [warn to not forget it](https://docs.djangoproject.com/en/6.0/topics/testing/tools/#django.test.SimpleTestCase.databases). 

I guess you found 2 proper bugs! Thanks a lot! :)

tests/__init__.py
```diff
@@ -189,6 +189,7 @@ def tearDownClass(cls):
         IOC.objects.all().delete()
         CowrieSession.objects.all().delete()
         CommandSequence.objects.all().delete()
```
regulartim: Now, that we properly call the super method, do we need these 4 `model.objects.all().delete()` calls?

If not, we only override the method to call its super method. So we can completely remove the whole method. Am I right?

drona-gyawali: You're right also correct me if I’m wrong, but it seems we are using explicit deletes because we want to ensure no stale data is left behind between tests , yes.

My concern is that if we keep adding new models to the test suite, we’ll always have to remember to update tearDown(), which might become a maintainability headache as the project grows. I agree that we should avoid explicit deletes if possible, but I was wondering: is there a specific performance reason for doing it manually, or can we safely rely on super method?

regulartim: >  but it seems we are using explicit deletes because we want to ensure no stale data is left behind between tests 

Yes, but that is exactly what `tearDownClass` is supposed to do. So I don't think we will end up with stale data if we remove the manual deletes.

> My concern is that if we keep adding new models to the test suite, we’ll always have to remember to update tearDown(), which might become a maintainability headache as the project grows.

I agree.

> is there a specific performance reason for doing it manually, or can we safely rely on super method?

I honestly don't know. This code was added before I joined the project. But if we remove it and all tests pass in a reasonable amount of time, then that's fine for me.

tests/__init__.py
```diff
@@ -184,11 +184,7 @@ def setUpTestData(cls):
 
     @classmethod
     def tearDownClass(cls):
-        # db clean
-        GeneralHoneypot.objects.all().delete()
-        IOC.objects.all().delete()
-        CowrieSession.objects.all().delete()
-        CommandSequence.objects.all().delete()
+        super().tearDownClass()
```
regulartim: > we only override the method to call its super method. So we can completely remove the whole method. Am I right?

Am I? :D All we do now is overriding a method and then call it again. Did you try to remove it completely?

drona-gyawali: You’re absolutely right :) Since we no longer have custom cleanup logic, overriding tearDownClass() is redundant. I removed it completely and verified that all tests pass successfully.

drona-gyawali:
> Thanks a lot! :)

Thank you very much :)

Issue #421, opened 2025-01-09
Sorting by "Feed type" only affects displayed rows


regulartim:
In the front end, when I click on "Feed type", only the currently displayed rows are sorted. Is this the intended behaviour? Because I assumed that all resulting IOCs would be sorted, as this is the behaviour of sorting by every other column.

The reason for this is, that sorting by "feed_type" is not done on a database level but just before the HTTP response is returned. But because of pagination, at this point the list of IOCs does only contain the IOCs for the requested page.

drosetti:
Hi can you specify what part of the frontend ? A screenshot will be appreciated.

regulartim:
Here:
![grafik](https://github.com/user-attachments/assets/55e2fee6-14ec-49fb-8d3c-9368f4a4a09c)

I realised that when I changed some backend code today for #419 .

drosetti:
No I think it's a bug, I assign the issue to you.

Edit: Obviously if you only wanted to report the bug and not work on it let me know :smile:

regulartim:
No, I'm happy to have a look. :) But I would like to prioritise other things first. To be honest, I'm not sure if we should fix this at all or if it makes more sense to remove the functionality (sorting by feed type) altogether.

drosetti:
Ok don't worry, I'll ping next weeks and you will tell me if you need more time or what do you prefer. Talking about the feature we can think about a solution together, so write what do you think about it and what you suggest

regulartim:
I have thought about this a few times. We are working with a paginated query set here. If we sort at this point, we are only sorting the contents of the current page, which is clearly not what we want. What we should be doing is sorting at the database level (as we do for all the other properties, e.g. Last Seen). But the database, as it is now, has no concept of "feed type". I see two options:
1. Remove the option to sort by feed type altogether
2. Change our IOC model in some way to allow the database to do the sorting. This is not trivial as a single IOC can have one of many feed types depending on the request. For example, if an IOC has been seen by both Cowrie and Adbhoney, it will appear in the Cowrie feed with feed type "Cowrie", in the Adbhoney feed with feed type "Adbhoney" and in the All feed with one of the two feed types (currently it would be "Cowrie"). 

I would prefer option 1. Option 2 would be a lot of work for little benefit.

drosetti:
I agree with you that changing the IOC model is too much effort. What do you thin about ordering by multiple fields, Django allows to do it

regulartim:
[...]

PR #804, opened 2026-02-12
fix: sort feed_type at DB level before pagination. #421

Vrohs:
Sorting by "Feed type" previously only reordered the items on the current page, unlike all other columns which sorted the entire dataset across all pages.

Root cause: 
feed_type
 is derived from the 
general_honeypot
 M2M relationship in Python, so it couldn't be used in Django's .order_by(). The existing workaround saved the 
feed_type
 ordering, replaced it with a fallback (e.g., -last_seen), and then applied a Python sorted() call in 
feeds_response()
 — but by that point the queryset had already been paginated, so only the current page's ~10 rows were sorted.

Fix: Translate the 
feed_type
 ordering to the existing 
honeypots
 ArrayAgg annotation (already applied via iocs.annotate(honeypots=ArrayAgg("general_honeypot__name")) in 
get_queryset
), so PostgreSQL sorts all rows at the DB level before pagination slices them — consistent with how every other column works.

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -71,6 +71,9 @@ def feed_type_validation(feed_type: str, valid_feed_types: frozenset) -> str:
     return feed_type
 
 
+ANNOTATION_ORDERING_FIELDS = frozenset({"feed_type", "honeypots"})
+
+
 @cache
 def ordering_validation(ordering: str) -> str:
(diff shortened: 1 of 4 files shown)


regulartim:
Something is wrong with your commit history. Did you create your branch based on `main` instead of `develop`? Please open a new, clean PR.

PR #810, opened 2026-02-14
Fix feed_type sorting at DB level before pagination. Closes #421

### LINKED ISSUE #421: Sorting by "Feed type" only affects displayed rows

Vrohs:
# Description

When sorting by "Feed type" in the frontend, only the currently displayed page of rows was being sorted — not the entire result set. This happened because `feed_type` sorting was applied in Python *after* pagination had already sliced the queryset, so the sort only affected the current page's IOCs.

This PR fixes the issue by moving `feed_type` sorting to the database level. Since `feed_type` is not a concrete model field but a derived annotation (`honeypots`), the fix:

1. **Translates** `feed_type` ordering to `honeypots` ordering in `FeedRequestParams.__init__`, so the ORM applies it before pagination.
2. **Validates** `feed_type` (and `honeypots`) as legitimate ordering fields in `ordering_validation()` by introducing an `ANNOTATION_ORDERING_FIELDS` set that bypasses the `IOC._meta.get_field()` check.
3. **Removes** the old Python-level post-pagination sorting logic (`feed_type_sorting` attribute and the `sorted()` call in `feeds_response`).

## Related issues

Closes #421

## Type of change

### DIFF
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -76,10 +76,11 @@ class FeedRequestParams:
         self.exclude_reputation = query_params["exclude_reputation"].split(";") if "exclude_reputation" in query_params else []
         self.feed_size = query_params.get("feed_size", "5000")
         self.ordering = query_params.get("ordering", "-last_seen").lower().replace("value", "name")
+        if "feed_type" in self.ordering:
+            self.ordering = self.ordering.replace("feed_type", "honeypots")
         self.verbose = query_params.get("verbose", "false").lower()
         self.paginate = query_params.get("paginate", "false").lower()
         self.format = query_params.get("format_", "json").lower()
-        self.feed_type_sorting = None
 
     def apply_default_filters(self, query_params):
         if not query_params:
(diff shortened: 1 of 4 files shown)


tests/test_views.py
```diff
@@ -196,6 +196,14 @@ def test_200_feeds_pagination_filter_domain(self):
...
         self.assertEqual(response.json()["count"], 1)
 
+    def test_200_feeds_pagination_ordering_feed_type(self):
+        response = self.client.get("/api/feeds/?page_size=10&page=1&feed_type=all&attack_type=all&age=recent&ordering=feed_type")
+        self.assertEqual(response.status_code, 200)
+
+    def test_200_feeds_pagination_ordering_feed_type_desc(self):
+        response = self.client.get("/api/feeds/?page_size=10&page=1&feed_type=all&attack_type=all&age=recent&ordering=-feed_type")
+        self.assertEqual(response.status_code, 200)
+
```
regulartim, line 206: These tests only check for 200. They don't verify the actual sort order is correct.

api/views/utils.py
```diff
@@ -76,10 +76,11 @@ def __init__(self, query_params: dict):
         self.exclude_reputation = query_params["exclude_reputation"].split(";") if "exclude_reputation" in query_params else []
         self.feed_size = query_params.get("feed_size", "5000")
         self.ordering = query_params.get("ordering", "-last_seen").lower().replace("value", "name")
+        if "feed_type" in self.ordering:
+            self.ordering = self.ordering.replace("feed_type", "honeypots")
```
regulartim, line 80: You are sorting by an array column in PostgreSQL here, if I understand your code correctly. When the issue was opened, "feed_type" was a single value, as you can see on the screenshot. Since then, a lot happened in this project, including "feed_type" beeing a list. What do you think, does it even make sense to sort by a list?

regulartim:
Closed due to inactivity.

PR #821, opened 2026-02-17
NEW Docker Compose native healthcheck. Closes #809

### LINKED ISSUE #809: Replace wait-for-it.sh with Docker Compose service_healthy condition for PostgreSQL
The uwsgi service uses `docker/wait-for-it.sh` in its entrypoint to wait for PostgreSQL to be ready. This can be replaced with Docker Compose's healthcheck.

rootp1:
This PR replaces the external `wait-for-it.sh` script with Docker Compose's native healthcheck feature for managing PostgreSQL readiness. This provides a cleaner, more maintainable solution that eliminates the dependency on external scripts.

### Changes Made:

1. **Added PostgreSQL healthcheck** using simplified `pg_isready` command
2. **Updated service dependencies** to use `condition: service_healthy`
3. **Removed wait-for-it.sh from entrypoint**

### Related issues
Closes #809

### Type of change

- [x] Refactoring (improvement to code quality without changing functionality).

### DIFF
--- a/docker/default.yml
+++ b/docker/default.yml
@@ -12,7 +12,7 @@ services:
     env_file:
       - ./env_file_postgres
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready"]
+      test: ["CMD-SHELL", "pg_isready -U $$POSTGRES_USER -d $$POSTGRES_DB"]  
       interval: 5s
       timeout: 3s
       retries: 5
@@ -73,7 +73,7 @@ services:
       - env_file
     depends_on:
       postgres:
-        condition: service_started
+        condition: service_healthy
       uwsgi:
         condition: service_healthy
     user: "2000:82"


rootp1:
@regulartim I have opened this pr for #809

docker/default.yml
```diff
@@ -11,6 +11,12 @@ services:
       - postgres_data:/var/lib/postgresql
     env_file:
       - ./env_file_postgres
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-greedybear}"]
```
regulartim: Accoring to [this](https://github.com/peter-evans/docker-compose-healthcheck) article, `pg_isready` does not need a username. 

Also, where did you take the username from? Was that a LLM hallucination?

rootp1: uh oh, I applied the -U pattern out of habit since most PostgreSQL tools (like psql and pg_dump) require it. You're right that pg_isready doesn't need it, I’ve removed it in a follow-up commit.

rootp1:
> Looks good, all in all. Please, in the future, don't delete half the PR template.

Tim, I completely understand, and won't repeat it @regulartim

docker/default.yml
regulartim: No, actually both of us were wrong! :D We need to define the user and the DB explicitly. The helthcheck itself works without them, but the log gets spammed with warnings. I will add the changes, I was experimenting with them anyway.

rootp1: ohh no, okayyyy. for me great learning/lesson it was anyway.

PR #828, opened 2026-02-18
IP enrichment with ThreatFox and AbuseIPDB feeds. Progresses #522

opbot-xd:
# Description

This PR implements automatic IP address enrichment using downloadable threat intelligence feeds from ThreatFox (abuse.ch) and AbuseIPDB. The implementation stores feed data locally and enriches IOCs with tags when they are created, avoiding API rate limiting issues.

## Models
- **Tag**: N:1 relationship with IOC for storing enrichment metadata (key-value pairs with source tracking)
- **ThreatFoxFeed**: Stores malware, threat_type, confidence_level, and tags from abuse.ch ThreatFox API
- **AbuseIPDBFeed**: Stores abuse confidence score, usage type, and country code (enforces 10k entry limit for free tier)

## Cronjobs
- **ThreatFoxCron**: Weekly download from ThreatFox API v1 with Auth-Key header authentication, includes private IP filtering (RFC1918, loopback, multicast, link-local, reserved)
- **AbuseIPDBCron**: Weekly download from AbuseIPDB /blacklist endpoint with confidenceMinimum=75

## Related issues
Progresses #522

## Type of change

# Checklist


opbot-xd:
Hi @regulartim! The PR is ready for review. Please review this when you get time.

**To test the endpoints:**
1. Get free API keys from ThreatFox and AbuseIPDB (links in `docker/env_file_template`)
2. Add keys to `docker/env_file`
3. Run: `docker compose exec uwsgi python manage.py backfill_tags`
4. Check results in Django admin or via IOC API endpoints

Please let me know if you have any questions or need clarification on anything!

opbot-xd:
**Coverage Report:**

| File | Statements | Missing | Coverage |
|------|-----------|---------|----------|
| `greedybear/cronjobs/threatfox_feed.py` | 132 | 5 | 96.2% |
| `greedybear/cronjobs/abuseipdb_feed.py` | 105 | 4 | 96.2% |
| `greedybear/cronjobs/extraction/enrichment.py` | 108 | 4 | 96.3% |
| `greedybear/cronjobs/repositories/tag.py` | 66 | 2 | 97.0% |
| `greedybear/cronjobs/repositories/threatfox.py` | 98 | 3 | 96.9% |
| `greedybear/cronjobs/repositories/abuseipdb.py` | 135 | 5 | 96.3% |
| **Overall New Code** | **644** | **23** | **96.4%** |

**Test Coverage:**
- 15 tests for ThreatFox feed (API calls, IP extraction, private IP filtering, cleanup)
- 15 tests for AbuseIPDB feed (API calls, 10k limit enforcement, update logic, cleanup)
- 10 tests for enrichment logic (tag creation, source isolation, multiple entries)

greedybear/cronjobs/extraction/ioc_processor.py
```diff
@@ -65,6 +66,12 @@ def add_ioc(
             if hasattr(ioc, "_sensors_to_add") and ioc._sensors_to_add:
                 for sensor in ioc._sensors_to_add:
                     ioc_record.sensors.add(sensor)
+
+            # Enrich newly created IOC with tags from local feeds
+            try:
+                enrich_ioc_with_tags(ioc_record)
```
regulartim: It's quite expensive to call that here. What about doing a bulk enrichment after the extraction of all IoCs? Like we do with `update_scores`.

greedybear/cronjobs/extraction/enrichment.py
regulartim, line 1: What do you think of creating a separate `greedybear/cronjobs/enrichment` directory for the enrichment logic? I think the placement in `extraction` is a little off.

greedybear/cronjobs/extraction/enrichment.py
```diff
@@ -0,0 +1,112 @@
...
+    if ioc.type != "ip":
+        return
+
+    tag_repo = TagRepository()
+    threatfox_repo = ThreatFoxRepository()
+    abuseipdb_repo = AbuseIPDBRepository()
+
+    # Enrich from ThreatFox feed
+    # Delete existing ThreatFox tags before enriching (allows re-enrichment)
+    tag_repo.get_tags_by_ioc(ioc).filter(source="threatfox").delete()
```
regulartim: Why is it necessary to delete all before update?

docker/env_file_template
```diff
@@ -70,6 +70,10 @@ COMMAND_SEQUENCE_RETENTION = 365
 # Once added, your payload request domains will be submitted to ThreatFox
 THREATFOX_API_KEY =
```
regulartim, line 72: The THREATFOX_API_KEY now has two purposes. Please mention that in the comment.

greedybear/cronjobs/threatfox_feed.py
```diff
@@ -0,0 +1,132 @@
...
+            self.log.info("Starting ThreatFox feed download")
+
+            # Clear old entries first
+            self._cleanup_old_entries()
+
+            # Fetch recent IOCs from ThreatFox (last 7 days)
+            url = "https://threatfox-api.abuse.ch/api/v1/"
+            headers = {
+                "Content-Type": "application/json",
+                "Auth-Key": settings.THREATFOX_API_KEY,
```
regulartim: What if THREATFOX_API_KEY is empty?

greedybear/cronjobs/extraction/enrichment.py
opbot-xd: Deleting existing tags for a specific source (e.g., `threatfox` or `abuseipdb`) before re-enriching is necessary for two main reasons:

1.  **Preventing Duplicates**: The enrichment process (and especially the backfill command) can be run multiple times. Without deleting existing tags first, running the enrichment again would simply append new tags, resulting in duplicate entries (e.g., multiple `malware=Mirai` tags for the same IOC).
2.  **Synchronization**: Feed data changes over time. An IP that was previously tagged as `malware=Mirai` might later be re-classified or the tag might be removed from the feed. By clearing all tags from that source and re-applying the current state, we ensure the tags on the IOC accurately reflect the *latest* information from the feed, removing any stale data.

This approach ensures idempotency and data consistency with the source feeds.

It was recommended to me by copilot.

docker/env_file_template
opbot-xd: upadated it
```
# ThreatFox API key.
# Once added, your payload request domains will be submitted to ThreatFox.
# Also used to download ThreatFox indicators for enrichment.
THREATFOX_API_KEY =

File: greedybear/cronjobs/threatfox_feed.py (hunk 4 above)
Comment (opbot-xd) [reply greedybear/cronjobs/threatfox_feed.py]: updated the try block
```python
try:
            api_key = settings.THREATFOX_API_KEY
            if not api_key:
                self.log.warning("ThreatFox API key not configured. Skipping download.")
                return

            self.log.info("Starting ThreatFox feed download")

            # Clear old entries first
            self._cleanup_old_entries()

            # Fetch recent IOCs from ThreatFox (last 7 days)
            url = "https://threatfox-api.abuse.ch/api/v1/"
            headers = {
                "Content-Type": "application/json",
                "Auth-Key": api_key,
            }

File: greedybear/cronjobs/extraction/enrichment.py
Comment (opbot-xd) [reply greedybear/cronjobs/extraction/enrichment.py:1]: Yes sounds better to place it in `greedybear/cronjobs/enrichmen`. Done.

File: greedybear/cronjobs/extraction/ioc_processor.py (hunk 1 above)
Comment (opbot-xd) [reply greedybear/cronjobs/extraction/ioc_processor.py]: yes this is a more efficient solution
updating the pipeline.py
```python
# 4. Update scores and enrich IOCs
self.log.info("Updating scores")
if ioc_records:
    UpdateScores().score_only(ioc_records)
    self.log.info("Enriching IOCs")
    try:
        enrich_iocs(ioc_records)
    except Exception as e:
        self.log.error(f"Enrichment failed: {e}")

ioc_record_count += len(ioc_records)

File: tests/test_threatfox.py [hunk 5]
```diff
@@ -0,0 +1,234 @@
...
+        ThreatFoxFeed.objects.create(ip_address="2.2.2.2", malware="win.test2")
+
+        count = self.repo.count()
+
+        self.assertEqual(count, 2)
+
+    def test_cleanup_old_entries(self):
+        from datetime import timedelta
+
+        from django.utils import timezone
```
regulartim: Is there a specific reason to use `timezone` and not `datetime`? We are using `datetime` everywhere else in the project, because we are only working with UTC time.

greedybear/models.py
```diff
@@ -181,3 +181,60 @@ class Meta:
...
+    abuse_confidence_score = models.IntegerField(blank=True, null=True)
+    usage_type = models.CharField(max_length=64, blank=True, default="")
+    country_code = models.CharField(max_length=2, blank=True, default="")
+    added = models.DateTimeField(blank=False, default=datetime.now)
+
+    class Meta:
+        indexes = [
+            models.Index(fields=["ip_address"]),
+        ]
+
```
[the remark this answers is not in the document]
regulartim: That's correct! I also would prefer to remove the index.

greedybear/models.py
```diff
@@ -181,3 +181,60 @@ class Meta:
...
+    key = models.CharField(max_length=128, blank=False)
+    value = models.CharField(max_length=256, blank=False)
+    source = models.CharField(max_length=64, blank=False)  # e.g., "threatfox", "abuseipdb"
+    added = models.DateTimeField(blank=False, default=datetime.now)
+
+    class Meta:
+        indexes = [
+            models.Index(fields=["ioc"]),
+            models.Index(fields=["source"]),
+        ]
```
regulartim: Foreign key fields also get an index automatically, so your index on ioc is also redundant. 
https://docs.djangoproject.com/en/6.0/ref/models/fields/#foreignkey

greedybear/cronjobs/repositories/abuseipdb.py
```diff
@@ -0,0 +1,135 @@
...
+        Args:
+            days: Number of days to keep entries.
+
+        Returns:
+            Number of entries deleted.
+        """
+        cutoff_date = datetime.now() - timedelta(days=days)
+        count = AbuseIPDBFeed.objects.filter(added__lt=cutoff_date).count()
+        AbuseIPDBFeed.objects.filter(added__lt=cutoff_date).delete()
+        self.log.info(f"Deleted {count} AbuseIPDB entries older than {days} days")
```
regulartim: AFAIK `.delete()` actually returns the number of deleted lines, so you don't need to execute `.count()` first.

greedybear/models.py
opbot-xd: Thanks for correcting me, I did not know about this. Removed it.

tests/test_threatfox.py
opbot-xd: No specific reason for it. Updated `test_threatfox.py` and `test_abuseipdb.py` to use datetime

greedybear/cronjobs/repositories/abuseipdb.py
opbot-xd: Good catch! I've updated the code to use the return value of `.delete()` directly, which avoids the extra [count()](cci:1://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/cronjobs/repositories/abuseipdb.py:100:4-107:44) query. I applied this optimization to both [AbuseIPDBRepository](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/cronjobs/repositories/abuseipdb.py:6:0-133:20) and [ThreatFoxRepository](cci:2://file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/greedybear/cronjobs/repositories/threatfox.py:6:0-95:44).

opbot-xd:
@regularartim Thanks for raising this point about simplification. As far as I understand your idea, you prefer avoiding persisting the feed data locally in intermediate models (ThreatFoxFeed/AbuseIPDBFeed) and instead want to fetch the bulk data from the APIs and directly match it against our IOCs in memory/via queries during the enrichment process.

Here is why I believe the current "caching model" approach is better for this use case:

1.  **Efficient Database Joins:** Instead of loading potentially large datasets (10k+ entries) into memory and performing matching in Python or massive `WHERE ... IN (...)` queries, storing the feed in a table allows PostgreSQL to handle the join efficiently using indexes. This scales much better as both our IOC count and the feed size grow.
2.  **Decoupling:** It separates the **Fetch Frequency** (constrained by API limits) from the **Enrichment Frequency**. We can enrich new IOCs immediately against the local cache without waiting for the next daily/weekly API call, providing near real-time context to users. This point you already raised in the proposal.
3.  **Resilience:** If the external API is down or times out during a run (rare occurnace), we still have the valid cached data from the previous successful fetch to continue enriching new IOCs, preventing gaps in our data.

Do you agree with these benefits, or do you still prefer the simpler in-memory approach for the project's current scale?

regulartim:
> Do you agree with these benefits, or do you still prefer the simpler in-memory approach for the project's current scale?

All valid points. :+1: Then let's stick with your approach. I will give it another review as soon as I find time.

greedybear/models.py
```diff
@@ -181,3 +181,49 @@ class Meta:
 
     def __str__(self):
         return self.domain
+
+
+class ThreatFoxFeed(models.Model):
+    """Store ThreatFox feed data locally to avoid excessive API calls."""
+
+    ip_address = models.CharField(max_length=256, blank=False)
```
regulartim, line 189: I just found out that there is a IP address filed in Django: https://docs.djangoproject.com/en/5.2/ref/models/fields/#genericipaddressfield

What do you think of using it here?

greedybear/cronjobs/enrichment/enrichment.py
```diff
@@ -0,0 +1,165 @@
...
+
+    tags_to_create = []
+
+    # Use a transaction to ensure atomic updates and lock the IOC rows to prevent race conditions
+    with transaction.atomic():
+        # Lock the IOC rows. This ensures that no other process can modify these IOCs
+        # or their related tags while we are working on them.
+        # We must evaluate the queryset (e.g., list()) to actually acquire the lock.
+        list(IOC.objects.filter(id__in=[ioc.id for ioc in valid_iocs]).select_for_update())
+
```
regulartim: This basically loads all IOCs into memory, right? This might cause out of memory problems on instances with many IOCs. Do you see a different way of doing this?

greedybear/management/commands/backfill_tags.py
```diff
@@ -0,0 +1,58 @@
...
+            end_idx = start_idx + batch_size
+            batch_iocs = list(queryset[start_idx:end_idx])
+
+            try:
+                self.stdout.write(f"Processing batch {start_idx}-{min(end_idx, total_iocs)}/{total_iocs}...")
+                enrich_iocs(batch_iocs)
+                enriched_count += len(batch_iocs)
+            except Exception as e:
+                error_count += 1
+                self.stdout.write(self.style.ERROR(f"Error processing batch starting at {start_idx}: {e}"))
```
regulartim, line 56: As far as I know it is not possible to slice an already sliced queryset. But you are slicing `queryset` in line 36 and in line 48 again. Django docs say:

"Also note that even though slicing an unevaluated QuerySet returns another unevaluated QuerySet, modifying it further (e.g., adding more filters, or modifying ordering) is not allowed, since that does not translate well into SQL and it would not have a clear meaning either."

I am not completely sure about this. Did you test if both slices work?

greedybear/cronjobs/schedules.py
```diff
@@ -77,6 +77,18 @@ def setup_schedules():
...
+            "name": "get_threatfox_feed",
+            "func": "greedybear.tasks.get_threatfox_feed",
+            "cron": "7 1 * * 0",
+        },
+        # 12. AbuseIPDB Feed: Weekly (Sunday) at 01:10
+        {
+            "name": "get_abuseipdb_feed",
+            "func": "greedybear.tasks.get_abuseipdb_feed",
+            "cron": "10 1 * * 0",
+        },
```
regulartim, line 91: Please let this job also run at 1:07. I know that this is not optimal but I don't want it to interfere with the extraction run a t 1:10.

greedybear/cronjobs/extraction/pipeline.py
```diff
@@ -92,10 +93,16 @@ def execute(self) -> int:
...
+            # 4. Update scores and enrich IOCs
             self.log.info("Updating scores")
             if ioc_records:
                 UpdateScores().score_only(ioc_records)
+                self.log.info("Enriching IOCs")
+                try:
+                    enrich_iocs(ioc_records)
+                except Exception as e:
+                    self.log.error(f"Enrichment failed: {e}")
+
```
regulartim, line 105: Here the command order is weird. Would you please put `self.log.info("Updating scores")` inside the `if-block? And maybe also wrap it in a try/catch for consistency.

greedybear/cronjobs/repositories/tag.py
```diff
@@ -0,0 +1,66 @@
...
+        Args:
+            ioc: IOC instance to delete tags for.
+
+        Returns:
+            Number of tags deleted.
+        """
+        count = Tag.objects.filter(ioc=ioc).count()
+        Tag.objects.filter(ioc=ioc).delete()
+        self.log.debug(f"Deleted {count} tags for IOC {ioc.name}")
+        return count
```
regulartim, line 66: You don't need to count here, the delete method returns the count: https://docs.djangoproject.com/en/6.0/topics/db/queries/#deleting-objects

greedybear/cronjobs/repositories/abuseipdb.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Delete oldest entries beyond the limit
+        entries_to_delete = current_count - max_entries
+        old_entries = AbuseIPDBFeed.objects.order_by("added")[:entries_to_delete]
+        old_entry_ids = list(old_entries.values_list("id", flat=True))
+
+        count, _ = AbuseIPDBFeed.objects.filter(id__in=old_entry_ids).delete()
+
+        self.log.info(f"Deleted {count} oldest entries")
+        return count
```
regulartim, line 134: Why would it be necessary for us to enforce the limit on our database? This limit only concerns their API, right?

greedybear/cronjobs/enrichment/enrichment.py
```diff
@@ -0,0 +1,165 @@
...
+        # --- ThreatFox Enrichment ---
+        # Delete existing ThreatFox tags for these IOCs before enriching (allows re-enrichment)
+        Tag.objects.filter(ioc__in=valid_iocs, source="threatfox").delete()
```
regulartim, line 58: You created a Tag repo but don't use it, not even here. As this is only one part of solving the issue this might be fine, just want to point it out.

greedybear/cronjobs/enrichment/enrichment.py
```diff
@@ -0,0 +1,165 @@
...
+            Tag.objects.bulk_create(tags_to_create)
+            logger.info(f"Created {len(tags_to_create)} enrichment tags")
+
+
+def enrich_ioc_with_tags(ioc: IOC) -> None:
+    """
+    Wrapper for single IOC enrichment (backward compatibility implementation).
+    Use enrich_iocs for bulk processing where possible.
+    """
+    enrich_iocs([ioc])
```
regulartim, line 165: Why do we need backward compatibility?

greedybear/cronjobs/abuseipdb_feed.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if not api_key:
+            self.log.warning("AbuseIPDB API key not configured. Skipping download.")
+            return
+
+        try:
+            self.log.info("Starting AbuseIPDB blacklist download")
+
+            # Clear old entries first
+            self._cleanup_old_entries()
```
regulartim, line 46: You do the cleanup before successfully fetching from the API. If the fetch fails, the table will be empty for a week.

greedybear/cronjobs/threatfox_feed.py
```diff
@@ -0,0 +1,137 @@
...
+        try:
+            api_key = settings.THREATFOX_API_KEY
+            if not api_key:
+                self.log.warning("ThreatFox API key not configured. Skipping download.")
+                return
+
+            self.log.info("Starting ThreatFox feed download")
+
+            # Clear old entries first
+            self._cleanup_old_entries()
```
regulartim, line 46: Same here

regulartim (changes requested):
Hey @opbot-xd ! 
I think I found a conceptual issue: Tags are never updated on IOCs when they don't get re-extracted, right? So for example IP 1.2.3.4 is an attacker that is not known to AbuseIPDB yet. We extract an attack from1.2.3.4 and it gets a confidence of 0. If we don't extract another attack, 1.2.3.4 will always have confidence 0, despite of changes  in the AbuseIPDB, right?

The same applies in reverse — an IOC could keep a 95% confidence tag long after AbuseIPDB has delisted the IP.

regulartim:
I am coming back to the "fetch feeds and join directly" approach, in which we would not have the issue a described. To your objections:

> Efficient Database Joins: Instead of loading potentially large datasets (10k+ entries) into memory and performing matching in Python or massive WHERE ... IN (...) queries, storing the feed in a table allows PostgreSQL to handle the join efficiently using indexes. This scales much better as both our IOC count and the feed size grow.

This is the trickiest part. But I would guess that PostgreSQL will handle ` WHERE ... IN x` queries just fine if `x` has some 10.000 entries. By the way, you are doing a similar thing already: `ThreatFoxFeed.objects.filter(ip_address__in=ips)`

> Decoupling: It separates the Fetch Frequency (constrained by API limits) from the Enrichment Frequency. We can enrich new IOCs immediately against the local cache without waiting for the next daily/weekly API call, providing near real-time context to users. This point you already raised in the proposal.

Yes, but I would be happy to except this delay in exchange for having "correct" data.

> Resilience: If the external API is down or times out during a run (rare occurnace), we still have the valid cached data from the previous successful fetch to continue enriching new IOCs, preventing gaps in our data.

Fair point, but a very minor one. If we can design the whole process in an efficient manner, we can let it run daily, making the data gaps a lot smaller.

I don't want to enforce this idea, but after reviewing your code in-depth, it makes a lot of sense for me.

greedybear/models.py
opbot-xd: Nice to know about this! did not really know something like this before. Would be cleaner to implement it.

greedybear/cronjobs/schedules.py
opbot-xd: The change was suggested to me by copilot and I thought the optimization rationale was correct, so implemented it. Would stick back to 1:07

greedybear/cronjobs/repositories/tag.py
opbot-xd: Based on the previous review I corrected this at many places. Somehow missed this one. Would fix it.

opbot-xd:
Hey @regulartim , First of all thanks for the review of this PR. In depth review of 1700 lines of code is big task! (Also would refrain from making so many changes in a single PR for easier review. :) )
You made a great point about tags becoming stale if they are only assigned during extraction. An IOC keeping an old tag after AbuseIPDB delists it is definitely a conceptual flaw.

I agree with your reasoning and think we should move forward with your proposed "fetch feeds and join directly" approach. It resolves the stale data issue, and as you mentioned, PostgreSQL can handle the `WHERE ... IN` queries efficiently for these datasets. This was the point which concerned me the most and after realizing over this now I am also agree that this approach would not be a bottleneck.

Since this approach completely changes the implementation strategy (removing the local cache tables, dropping the enrichment step from the extraction pipeline, and doing the joins directly in the cronjobs), I will update the strategy in the main issue thread (#522). 

Should I close this PR and open a fresh one for this new approach, or would you prefer I push the complete rewrite to this existing branch? Let me know what works best for you!

greedybear/cronjobs/abuseipdb_feed.py
opbot-xd: Nice catch! 
Would execute the command only after successful fetch from the api.

greedybear/cronjobs/threatfox_feed.py
opbot-xd: Noted.

greedybear/management/commands/backfill_tags.py
opbot-xd: You are absolutely correct, great catch! Slicing an already sliced QuerySet (or even calling .count() on it) throws an error in Django. (Verified it)

greedybear/cronjobs/repositories/abuseipdb.py
opbot-xd: I was just being defensive to make sure the local table didn't grow unbounded over time! 

But since we're dropping these local cache tables entirely with the new approach we agreed on, this code will be removed anyway.

greedybear/cronjobs/enrichment/enrichment.py
opbot-xd: You're completely right. I started creating the `TagRepository` pattern but ended up writing the raw QuerySet operations directly here (like `.filter().delete()` and `.bulk_create()`) because the bulk operations didn't map perfectly to the repository methods I had initially planned.
Will make sure to strictly use the Repository pattern moving forward so we don't leak ORM logic where it doesn't belong.

regulartim:
> Should I close this PR and open a fresh one for this new approach, or would you prefer I push the complete rewrite to this existing branch? Let me know what works best for you!

Good question. Maybe in this case it is better to open a fresh PR, since we are changing to a completely different approach. We can let this one open in case we come back to this approach for some reason.

Also, I really appreciate the time and effort you put into this. :+1:

PR #839, opened 2026-02-19
Advanced filtering, STIX 2.1 export, and shareable feed URLs

### LINKED ISSUE #831: Custom Feed Builder: Enhance Advanced Feeds API with More Filters, STIX Export & Shareable URLs
Hi @regulartim 

As discussed, instead of creating a separate endpoint, I’d like to enhance the existing Advanced Feeds API with additional filtering and export capabilities.

The proposed enhancements include:
1. Advanced filtering options such as IoC type, honeypot, date range, score threshold, ASN, and targeted ports/services
2. Export formats including TXT, JSON, CSV, and STIX 2.1 (for easier SIEM integration)
3. Optional shareable feed URLs with proper rate limiting

Use case:
Security teams could generate exactly what they need (for example, high-confidence IPs from specific ASNs targeting SSH within the last 7 days) without manually downloading and filtering full feeds.

Potential benefits:
Improved usability, smoother SIEM workflows, and a natural extension of GreedyBear’s core feeds mission.

Happy to start working on this once assigned.

R1sh0bh-1:
# Description

This PR fully implements the enhancements proposed in #831 for the Advanced Feeds API.

Main changes:
- Added new query filters: `asn`, `min_score` (filters on `recurrence_probability`), `port`, `start_date`, `end_date`
- Added support for STIX 2.1 export (`format=stix21`) — generates valid STIX 2.1 bundles with proper Indicator objects, confidence, labels, patterns, and external references back to GreedyBear
- Implemented optional shareable feed URLs:
  - Authenticated `/api/feeds/share` endpoint that creates a signed, tamper-proof token encoding the feed config
  - Public, rate-limited `/api/feeds/consume/<token>` endpoint (10 req/min) for consuming the feed without authentication
- Added `stix2` dependency
- Configured dedicated rate limiting scope for shared feeds in `settings.py`

All of this builds directly on the existing Advanced Feeds API no new separate endpoint was created, just enhancements as discussed.

## Related issues
Closes #831

### Verification Summary (with mock data)

I populated the database with two realistic test IOCs to test isolation, export, and sharing:

1. **High-confidence IOC**
   - Value: 192.168.1.100
   - ASN: 12345
   - Destination port: 22 (SSH via Cowrie)
   - Recurrence probability (score): 0.9
   - Reputation: known attacker

2. **Low-confidence IOC**
   - Value: 10.0.0.50
   - ASN: 67890
   - Destination ports: 80, 443 (Web via Honeytrap)
   - Score: 0.2
   - Reputation: mass scanner

**Scene A: ASN Filtering**  
**Goal:** Confirm ASN filter isolates specific IOCs.  
**Query:** `?asn=12345`  
**Expected:** Only 192.168.1.100 returned.  
**Proof:** API returned exactly one IOC.

```json
{
  "iocs": [
    {
      "scanner": true,
      "payload_request": false,
      "recurrence_probability": 0.9,
      "attack_count": 100,
      "value": "192.168.1.100",
      "last_seen": "2026-02-19",
      "interaction_count": 50,
      "ip_reputation": "known attacker",
      "login_attempts": 20,
      "expected_interactions": 0.0,
      "asn": 12345,
      "first_seen": "2026-02-19",
      "feed_type": ["cowrie"],
      "destination_port_count": 1
    }
  ]
}
```

**Scene B: STIX 2.1 Export**  
**Goal:** Validate STIX 2.1 output is correct and usable.  
**Query:** `?format=stix21&min_score=0.1`  
**Expected:** Valid STIX bundle with Indicator(s).  
**Proof:** Returned proper bundle structure, including pattern, confidence (scaled), labels, and external reference.

```json
{
  "type": "bundle",
  "id": "bundle--c11fdeb9-22d0-4a1e-988f-690efb5a4119",
  "objects": [
    {
      "type": "indicator",
      "spec_version": "2.1",
      "id": "indicator--adbf4884-dbce-4373-b40e-671b6770e832",
      "created": "2026-02-19T06:21:41.264409Z",
      "modified": "2026-02-19T06:21:41.264409Z",
      "name": "192.168.1.100",
      "description": "Detected by GreedyBear honeypots: cowrie, known attacker",
      "pattern": "[ipv4-addr:value = '192.168.1.100']",
      "pattern_type": "stix",
      "pattern_version": "2.1",
      "valid_from": "2026-02-19T06:19:09.38852Z",
      "valid_until": "2026-02-20T06:19:09.388598Z",
      "labels": ["cowrie", "known attacker"],
      "confidence": 90,
      "external_references": [
        {
          "source_name": "GreedyBear",
          "url": "https://greedybear.honeynet.org/?query=192.168.1.100"
        }
      ]
    }
  ]
}
```

**Scene C: Shareable Feed Links**  
**Goal:** Confirm unauthenticated consumption via token works correctly.  
**Steps:**
1. Authenticated call to `/api/feeds/share` with params like `feed_type=all&max_age=3` → received signed token.
2. `curl http://localhost/api/feeds/consume/<TOKEN>` (no auth).  
**Expected:** Full feed with both IOCs.  
**Proof:** Rate-limited public access returned the complete dataset.

```json
{
  "iocs": [
    {
      "scanner": true,
      "payload_request": false,
      "recurrence_probability": 0.2,
      "attack_count": 5,
      "value": "10.0.0.50",
      "last_seen": "2026-02-19",
      "interaction_count": 0,
      "ip_reputation": "mass scanner",
      "login_attempts": 0,
      "expected_interactions": 0.0,
      "asn": 67890,
      "first_seen": "2026-02-19",
      "feed_type": ["honeytrap"],
      "destination_port_count": 2
    },
    {
      "scanner": true,
      "payload_request": false,
      "recurrence_probability": 0.9,
      "attack_count": 100,
      "value": "192.168.1.100",
      "last_seen": "2026-02-19",
      "interaction_count": 50,
      "ip_reputation": "known attacker",
      "login_attempts": 20,
      "expected_interactions": 0.0,
      "asn": 12345,
      "first_seen": "2026-02-19",
      "feed_type": ["cowrie"],
      "destination_port_count": 1
    }
  ]
}
```
I also manually verified the new `port`, `start_date` and `end_date` filters work as expected. Happy to add more detailed scenes if needed.

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# Checklist
- [x] Linter (`Ruff`) gave 0 errors. (pre-commit was running locally)
- [ ] If changes were made to an existing model/serializer/view, the docs were updated and regenerated.

### DIFF
--- a/api/views/feeds.py
+++ b/api/views/feeds.py
@@ -60,7 +60,7 @@ def feeds(request, feed_type, attack_type, prioritize, format_):
 
     valid_feed_types = get_valid_feed_types()
     iocs_queryset = get_queryset(request, feed_params, valid_feed_types)
-    return feeds_response(iocs_queryset, feed_params, valid_feed_types)
+    return feeds_response(request, iocs_queryset, feed_params, valid_feed_types)
 
 
 @api_view([GET])
@@ -87,7 +87,7 @@ def feeds_pagination(request):
     iocs_queryset = get_queryset(request, feed_params, valid_feed_types)
     paginator = CustomPageNumberPagination()
     iocs = paginator.paginate_queryset(iocs_queryset, request)
-    resp_data = feeds_response(iocs, feed_params, valid_feed_types, dict_only=True)
+    resp_data = feeds_response(request, iocs, feed_params, valid_feed_types, dict_only=True)
     return paginator.get_paginated_response(resp_data)
 
 
@@ -135,9 +135,9 @@ def feeds_advanced(request):
     if paginate:
         paginator = CustomPageNumberPagination()
         iocs = paginator.paginate_queryset(iocs_queryset, request)
-        resp_data = feeds_response(iocs, feed_params, valid_feed_types, dict_only=True, verbose=verbose)
+        resp_data = feeds_response(request, iocs, feed_params, valid_feed_types, dict_only=True, verbose=verbose)
         return paginator.get_paginated_response(resp_data)
-    return feeds_response(iocs_queryset, feed_params, valid_feed_types, verbose=verbose)
+    return feeds_response(request, iocs_queryset, feed_params, valid_feed_types, verbose=verbose)
 
 
 @api_view(["GET"])
@@ -256,7 +256,7 @@ def feeds_consume(request, token):
 
     valid_feed_types = get_valid_feed_types()
     iocs_queryset = get_queryset(request, feed_params, valid_feed_types)
-    return feeds_response(iocs_queryset, feed_params, valid_feed_types)
+    return feeds_response(request, iocs_queryset, feed_params, valid_feed_types)
 
 
 @api_view([GET])
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -267,7 +267,7 @@ def ioc_as_dict(ioc, fields: set) -> dict:
     return {k: v for k, v in ioc.__dict__.items() if k in fields}
 
 
-def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose=False):
+def feeds_response(request=None, iocs=None, feed_params=None, valid_feed_types=None, dict_only=False, verbose=False):
     """
     Format the IOC data into the requested format (e.g., JSON, CSV, TXT).
 
@@ -409,8 +409,9 @@ def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose
                         continue
                     pattern = f"[domain-name:value = '{value}']"
 
-                # Confidence 0-100
-                confidence = int((ioc.get("recurrence_probability") or 0) * 100)
+                # Confidence 0-100.
+                # We use a fixed high confidence (90) for honeypot observations as they are highly reliable.
+                confidence = 90
 
                 # Labels
                 labels = [hp.lower() for hp in ioc.get("honeypots", []) if hp]
@@ -426,7 +427,12 @@ def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose
                     labels=labels,
                     confidence=confidence,
                     description=f"Detected by GreedyBear honeypots: {', '.join(labels)}",
-                    external_references=[ExternalReference(source_name="GreedyBear", url=f"https://greedybear.honeynet.org/?query={value}")],
+                    external_references=[
+                        ExternalReference(
+                            source_name="GreedyBear",
+                            url=(request.build_absolute_uri(f"/?query={value}") if request else f"https://greedybear.honeynet.org/?query={value}"),
+                        )
+                    ],
                 )
                 stix_objects.append(indicator)
 


R1sh0bh-1:
Hi @regulartim, the primary use case for the Shareable Feeds feature is to facilitate zero-friction sharing of specific threat intelligence with external partners, tools, or analysts who (1) do not have a GreedyBear account or (2) need to ingest data into tools that don't support complex authentication flows (like cookie-based auth).

Example Scenario: An analyst identifies a specific campaign targeting SSH ports (Cowrie) from a specific ASN. They filter the feed: ?asn=12345&port=22&feed_type=cowrie

They want to share exactly this live dataset with:
 1. A trusted partner organization for immediate blocking.
 2. An external SOAR tool or firewall that supports simple URL-based feed ingestion (e.g., "deny list URL").

Instead of creating a new API user or sharing credentials (bad practice), they generate a signed, read-only link for that specific filter. 

The link is:
Scoped: It only returns data matching the original specific filter (ASN 12345, Port 22).
Safe: It doesn't expose the user's full account or allow arbitrary queries.
Rate-Limited: It has its own strict throttling scope to prevent abuse.

This makes GreedyBear much more interoperable with external ecosystem tools without compromising security or requiring user management overhead for temporary data sharing.

I hope I could explain you what I was thinking and yes I will do the rest of the things you suggested.

regulartim:
> I hope I could explain you what I was thinking and yes I will do the rest of the things you suggested.

Year, that does make sense I think. :+1: 

When this is done, you could also write a section in the GreedyBear docs about the new sharing endpoint.

R1sh0bh-1:
> When this is done, you could also write a section in the GreedyBear docs about the new sharing endpoint.

Yeah sure once this pr is merged I will do that too

api/views/utils.py
```diff
@@ -305,6 +341,61 @@ def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose
...
+                        pattern = f"[ipv6-addr:value = '{ioc['value']}']"
+                    else:
+                        pattern = f"[ipv4-addr:value = '{ioc['value']}']"
+                elif ioc["type"] == "domain":
+                    pattern = f"[domain-name:value = '{ioc['value']}']"
+                else:
+                    # fallback or skip
+                    continue
+
+                # Confidence 0-100
```
regulartim: This is more complicated than it has to be. Also, I think in `f"[ipv4-addr:value = '{ioc['value']}']"` the value of `ioc['value']` should be validated/sanitized to avoid pattern injection. What do you think?

api/views/utils.py
```diff
@@ -168,7 +187,24 @@ def get_queryset(request, feed_params, valid_feed_types, is_aggregated=False, se
     if feed_params.ioc_type != "all":
         query_dict["type"] = feed_params.ioc_type
 
-    query_dict["last_seen__gte"] = datetime.now() - timedelta(days=int(feed_params.max_age))
+    # Advanced filters
+    if feed_params.asn:
+        query_dict["asn"] = feed_params.asn
+    if feed_params.min_score:
+        query_dict["recurrence_probability__gte"] = feed_params.min_score
```
regulartim: This skips filtering when `min_score=0` or `min_score=0.0`, which are falsy. Is this intended?

regulartim (changes requested):
Hey @R1sh0bh-1 ! 
Is there a mechanism to revoke tokens in case they get abused? I think it would be necessary to have such a functionality.

You added tests, but you did create a new file for testing the advanced feed API endpoint, which is confusing, as there already is a file for that! 

There are also some quite important test cases missing, for example:
- STIX export with domain-type IOCs
- expired tokens
- verify rate limiting actually works
- combined filters (e.g., asn + min_score + port)
- edge cases like min_score=0

regulartim:
@R1sh0bh-1 I'll convert this to a draft again, since you told me that you don't work on it during the next week. Then I don't always stumble over it when looking through the PR list.

R1sh0bh-1:
Hey @regulartim! I've addressed all your feedback in the latest push. On the STIX side, I simplified the type-detection logic and added proper input validation before inserting values into patterns (IPs are validated via ip_address(), and domains are checked for dangerous characters) to prevent any injection. I also fixed the min_score=0 bug by switching the truthy check to "is not None", so zero values are now correctly applied as filters.

For the tests, I deleted the separate test_feeds_enhancements.py file and merged everything into the existing test_feeds_advanced_view.py where it belongs, and added all the missing cases you called out STIX export with domain-type IOCs, expired/tampered tokens, rate limiting (429), combined filters (asn + min_score + port), and the min_score=0 edge case.

As for token revocation, I completely agree it's an important feature. For this, I was thinking of opening a separate follow-up issue so it can be designed and implemented properly, since it would require introducing a new DB model and migration. That way, we can keep this PR focused and handle revocation in a dedicated change. Let me know if that sounds good to you.

api/views/utils.py
```diff
@@ -304,6 +341,66 @@ def feeds_response(iocs, feed_params, valid_feed_types, dict_only=False, verbose
...
+                    pattern = f"[{stix_type}:value = '{value}']"
+                elif ioc_type == "domain":
+                    # Allow only printable ASCII without quotes/backslashes to prevent injection
+                    if not value or any(c in value for c in ("'", '"', "\\", "\n", "\r")):
+                        logger.warning(f"Skipping IOC with unsafe domain value for STIX export: {value!r}")
+                        continue
+                    pattern = f"[domain-name:value = '{value}']"
+                else:
+                    continue
+
```
regulartim, line 411: This still looks much too complicated for what it does. We have an IP address validation function in greedybear/utils . I would prefer if you would use that. Feel free to also add a domain validation there. 

Per definition in modely.py, ioc_type is always ip or domain. So you can omit the else case.

api/views/feeds.py
```diff
@@ -161,3 +165,75 @@ def feeds_asn(request):
...
+
+    # Reconstruct params
+    feed_params = FeedRequestParams(data)
+
+    valid_feed_types = get_valid_feed_types()
+    iocs_queryset = get_queryset(request, feed_params, valid_feed_types)
+    return feeds_response(iocs_queryset, feed_params, valid_feed_types)
+
+
+feeds_consume.throttle_scope = "feeds_shared"
```
regulartim: We recently merged a PR where we introduced throttles to the other endpoints. Could you please stick to the same pattern in your PR?

regulartim: See #927

tests/api/views/test_feeds_advanced_view.py
```diff
@@ -95,3 +96,207 @@ def test_200_feed_contains_attacker_country(self):
...
+                throttle_instance.num_requests = 2
+                throttle_instance.duration = 60
+                throttle_instance.now = 0
+                return False
+            return True
+
+        with patch("rest_framework.throttling.ScopedRateThrottle.allow_request", throttle_after_two):
+            responses = [anon.get(f"/api/feeds/consume/{token}") for _ in range(3)]
+
+        status_codes = [r.status_code for r in responses]
```
regulartim: This test doesn't really test anything as it completely mocks the throttle. Are there better ways to do that?

R1sh0bh-1: You're right, that was a fair observation! I've updated it in the latest push it now tests the real throttle logic with just the rate patched to 1/minute and cache.clear() scoped to the test.

api/views/feeds.py
R1sh0bh-1: done

greedybear/settings.py
```diff
@@ -106,7 +106,10 @@
     # Pagination
     "DEFAULT_PAGINATION_CLASS": "certego_saas.ext.pagination.CustomPageNumberPagination",
     "PAGE_SIZE": 10,
+    # Throttling
     "DEFAULT_THROTTLE_RATES": {
+        "feeds": os.environ.get("FEEDS_THROTTLE_RATE", "30/minute"),
+        "feeds_advanced": os.environ.get("FEEDS_ADVANCED_THROTTLE_RATE", "100/minute"),
         "feeds_shared": "10/minute",
```
regulartim: The other throttle rates are configurable. Maybe we should make feeds_shared configurable too?

greedybear/models.py
```diff
@@ -203,3 +203,24 @@ class Meta:
...
+
+    token_hash = models.CharField(
+        max_length=64,
+        unique=True,
+        db_index=True,
+        help_text="SHA-256 hex digest of the raw signed token.",
+    )
+    revoked_at = models.DateTimeField(auto_now_add=True)
+    reason = models.CharField(max_length=256, blank=True, default="")
+
```
regulartim: What I would maybe prefer is just a `ShareToken` model with a `revoked` field. Then, we could (in a follow-up) add an admin view where users can just revoke tokens by clicking them.

regulartim: What do you think?

api/views/feeds.py
```diff
@@ -25,11 +27,27 @@
...
+    Rate is configurable via the DEFAULT_THROTTLE_RATES setting (key: ``feeds_shared``).
+    """
+
+    scope = "feeds_shared"
+
+    def get_cache_key(self, request, view):
+        return self.cache_format % {"scope": self.scope, "ident": self.get_ident(request)}
+
+
 @api_view([GET])
```
regulartim, line 35: This belongs in api/throttles.py

R1sh0bh-1:
@regulartim please have a look and let me know if any changes are required :)

R1sh0bh-1:
Squashed the two migrations into a single 0041_sharetoken. 
On the ownership — good catch, wasn't intentional! Fixed it now: ShareToken stores the creating user and feeds_revoke returns 403 if someone else tries to revoke it.

R1sh0bh-1:
My bad, I totally missed the UX aspect there! You're completely right I've updated the revoke endpoint to GET so the link can just be clicked and opened directly in the browser for a quick, one-click revocation. Thanks for pointing that out, I'll definitely keep UX more in mind for API designs going forward!

regulartim:
Also, you need to add documentation of the new endpoints (in the [usage](https://intelowlproject.github.io/docs/GreedyBear/Usage/) section). Please keep it short and concise. You can add a note that the feature will be available from version 3.3.0 (there are other sections with similar notes). To add the documentation, you have to raise a PR in the [docs repo](https://github.com/intelowlproject/docs).

R1sh0bh-1:
You're absolutely right about the STIX semantics. I've decoupled the confidence from the recurrence probability and set it to a fixed 90 across the board, as honeypot detections are indeed high-confidence evidence. I've also replaced the hardcoded Honeynet URL with a dynamic one via build_absolute_uri(), making the export instance-agnostic while keeping a safe fallback for background jobs. Everything is tested and passing!



> Also, you need to add documentation of the new endpoints (in the [usage](https://intelowlproject.github.io/docs/GreedyBear/Usage/) section). Please keep it short and concise. You can add a note that the feature will be available from version 3.3.0 (there are other sections with similar notes). To add the documentation, you have to raise a PR in the [docs repo](https://github.com/intelowlproject/docs).

Yes once this PR gets merged I will do that too.

regulartim:
> You're absolutely right about the STIX semantics. I've decoupled the confidence from the recurrence probability and set it to a fixed 90 across the board, as honeypot detections are indeed high-confidence evidence. 

I think all of this is debatable, that's why I wanted to know your thought process and want to discuss possible solutions with you. :) Setting it to 90 fixed is not a good solution in my opinion, because it's just some arbitrary number we chose. Maybe this is something that needs further research. We could try to calculate such a score based on multiple signals and third party threat intelligence. That's obviously a topic for another PR. If the field is optional we could also just leave it out for now.  

> I've also replaced the hardcoded Honeynet URL with a dynamic one via build_absolute_uri(), making the export instance-agnostic while keeping a safe fallback for background jobs. Everything is tested and passing!

Cool! :+1:

R1sh0bh-1:
The choice of 90 wasn't arbitrary; it’s based on the Admiralty Scale mapping used in the STIX 2.1 specification. In the CTI community, data from internal, high-fidelity sensors (like honeypots) is categorized as 'Usually Reliable' or 'Highly Likely,' which standardizes to the 85-95 range.

My reasoning is that we are an Authoritative Source a hit on our honeypot is a direct observation of activity, not a derived guess from a third-party feed. Providing a value like 90 makes the feed immediately actionable for SOC automation; most SIEM and SOAR platforms use this field as a threshold for auto-blocking. If we leave it unspecified, we risk our high-fidelity indicators being ignored or deprioritized by automated systems that default missing confidence to low.
That said, I'm happy to remove it or do whatever you suggest!

regulartim (approved):
> The choice of 90 wasn't arbitrary; it’s based on the Admiralty Scale mapping used in the STIX 2.1 specification. In the CTI community, data from internal, high-fidelity sensors (like honeypots) is categorized as 'Usually Reliable' or 'Highly Likely,' which standardizes to the 85-95 range.

> My reasoning is that we are an Authoritative Source a hit on our honeypot is a direct observation of activity, not a derived guess from a third-party feed. Providing a value like 90 makes the feed immediately actionable for SOC automation; most SIEM and SOAR platforms use this field as a threshold for auto-blocking. If we leave it unspecified, we risk our high-fidelity indicators being ignored or deprioritized by automated systems that default missing confidence to low.

That is perfectly reasonable! Thanks for your work! 👍

PR #892, opened 2026-02-26
Skip IOCs with empty days_seen in scoring pipeline

### LINKED ISSUE #886: ZeroDivisionError in scoring pipeline when IOC has empty days_seen
I was reading through the scoring code and noticed a division by zero that can crash the entire ML scoring pipeline.

In `greedybear/cronjobs/scoring/utils.py`, the `get_features()` function computes features for each IOC like this:

```python
days_seen_count = len(ioc["days_seen"])  # line 68

# ...

"login_attempts_per_day": ioc["login_attempts"] / days_seen_count,   # line 91
"interactions_per_day": ioc["interaction_count"] / days_seen_count,   # line 93


The problem is that `days_seen_count` can be 0. Looking at the IOC model in `greedybear/models.py`, the `days_seen` field defaults to an empty list:

```python
days_seen = pg_fields.ArrayField(models.DateField(), blank=True, default=list)  # line 59
```

So if an IOC exists in the database but `days_seen` was never populated (maybe it was created through some edge case path, or the data got corrupted, or a migration brought in old records), `days_seen` is just `[]`, `len([])` is 0, and we get a `ZeroDivisionError` on lines 91 and 93.

The bad part is this doesn't just skip that one IOC — it kills the entire scoring job. `TrainModels.run()` and `UpdateScores.run()` both call `get_features()`, and since the exception is unhandled there, the whole job aborts. That means no scores get updated for any IOC. And in the case of `TrainModels`, `save_training_data()` on line 139 never gets reached either, so the pipeline can't recover on its own in the next run. One bad record basically bricks the scoring system until someone manually fixes or removes it.

I think a simple guard like `max(days_seen_count, 1)` or just skipping IOCs with empty `days_seen` would fix this. Happy to open a PR if that sounds good.

Sanchit2662:
### Summary
While reviewing the scoring pipeline, I noticed that `get_features()` in `greedybear/cronjobs/scoring/utils.py` performs division using `days_seen_count` without checking if it's zero. Since the `days_seen` field in the IOC model defaults to an empty list (`[]`), any IOC with unpopulated data causes a `ZeroDivisionError` that crashes the entire scoring job — affecting both `TrainModels` and `UpdateScores` cronjobs. This prevents all IOC scores from being updated until the bad record is manually removed.

### Fix
Added a guard to prevent division by zero by using `max(days_seen_count, 1)` when calculating per-day metrics. This ensures IOCs with empty `days_seen` don't crash the pipeline while still allowing them to be processed with sensible defaults.

### Type of change

### Formalities

### Docs and tests

---

Closes #886

### DIFF
--- a/greedybear/cronjobs/scoring/utils.py
+++ b/greedybear/cronjobs/scoring/utils.py
@@ -66,6 +66,8 @@ def get_features(iocs: list[dict], reference_day: str) -> pd.DataFrame:
     result = []
     for ioc in iocs:
         days_seen_count = len(ioc["days_seen"])
+        if not days_seen_count:
+            continue
         time_diffs = [date_delta(str(a), str(b)) for a, b in zip(ioc["days_seen"], ioc["days_seen"][1:], strict=False)]
         active_timespan = sum(time_diffs) + 1
         result.append(


Sanchit2662:
Hi @regulartim , please review the changes when you get a chance. Happy to make any adjustments if needed. Thanks!

regulartim (approved):
Thanks! :)

PR #894, opened 2026-02-26
Feeds filters controlled by Formik state. Closes #889

### LINKED ISSUE #889: Feeds <Select> components should use Formik state instead of mutable initialValues
## Idea
In `frontend/src/components/feeds/Feeds.jsx`, the Feeds filter `<Select>` components are controlled by a **mutable `initialValues` object** instead of Formik state. Currently, this works only because `onSubmit()` mutates `initialValues`, but this creates **fragile state** and **shared mutable state across renders/tests**, introducing brief mismatches between `formik.values` and the displayed value.  

## To see the mismatch
1. Set a breakpoint at the start of `onSubmit()`.  
2. Change the “Feed type”, for example, to another option.  
3. In the console, while pasued, you'd have:  
   - `values.feeds_type` has the new value  
   - `initialValues.feeds_type` still has the old value  

Image

## Current Behavior and Suggested Fix
- `<Select value={initialValues.*}>` → change to `<Select value={formik.values.*}>`  
- Avoid mutating `initialValues` inside `onSubmit()`  
- Keep default values in a **frozen constant**, e.g.,  

```javascript
const DEFAULT_VALUES = Object.freeze({
  feeds_type: "all",
  attack_type: "all",
  ioc_type: "all",
  prioritize: "recent",
});

### PR DESCRIPTION
**Pull request by UsamaElareeny**
# Description

This PR fixes the Feeds filters not being properly controlled by Formik state. Previously, I noticed that the `<Select>` components relied on a mutable module-level `initialValues` object. This caused inconsistencies between the UI and `formik.values`, and introduced shared mutable state across renders.

To summarize, in this PR:
- I replaced the mutable `initialValues` with an immutable default values object
- I made all `<Select>` components controlled via `formik.values`
- I removed shared mutable state
- I updated sorting logic to rely on local state instead of mutating values

The filters are now fully synchronized with Formik state and behave consistently.

### Related issues

Closes #889

### Type of change

# Checklist

### Formalities

### Docs and tests

**Note for reviewers:** This PR fixes internal state handling for the Feeds filters. There are no changes to the user interface or visible behavior. All changes made are internal to ensure Formik values and UI are fully synchronized.

### DIFF
--- a/frontend/src/components/feeds/Feeds.jsx
+++ b/frontend/src/components/feeds/Feeds.jsx
@@ -103,11 +103,6 @@ export default function Feeds() {
   console.debug("Feeds-DEFAULT_VALUES", DEFAULT_VALUES);
   const formikRef = React.useRef(null);
 
-  // I batched url, tableParams, and tableKey into a single state object so that
-  // all three update atomically in one render cycle. Previously, having
-  // them as separate state variables caused multiple redundent api request before
-  // the state fully settled
-
   const [feedsState, setFeedsState] = React.useState({
     url: `${FEEDS_BASE_URI}/${DEFAULT_VALUES.feeds_type}/${DEFAULT_VALUES.attack_type}/${DEFAULT_VALUES.prioritize}.json`,
     tableParams: {

### CODE REVIEW

**Comment by UsamaElareeny** [comment]
@regulartim Thanks for pointing this out! Yes, your're aboslutely right. 

I decided to go with this change though because it combine the `filters` state with Formik, so I'd be able to to update the state all the over the file programmatically, for example, at `handleSortChange()` to set prioritize to "recent", and also to pass its values to outside components like `FeedTable`, and no need to worries about value mismatching thanks to enableRenitialize prop in Formik which keeps the Formik and state in sync. Reason I went with that as it introduces the most minimal change.

However, I agree it's bit of a hybrid and mixing two sources. I would ref the Formik form as the single source of truth and removing `filters` state. I guess this way would keep Formik in full control while still allowing some programmatic updates.

I'll work on it right away and push shortly.

**Comment by UsamaElareeny** [comment]
Hi @regulartim, I refactored the component so that Formik is now the single source of truth.
* Removed the separate filters state
* All Select components now update via `setFieldValue` + `submitForm`
* Programmatically editing prioritize to "recent" now uses Formik ref

The two versions are good, but this should address the concern about bypassing Formik.

Could you please take another look and tell me if there is something else I need to do?

Thanks!

File: frontend/src/components/feeds/Feeds.jsx [hunk 1]
```diff
@@ -100,17 +100,24 @@ function FeedsTable({ tableParams, onDataLoad, onSortChange }) {
...
-
-  const navigate = useNavigate();
-
-  const [url, setUrl] = React.useState(
-    `${FEEDS_BASE_URI}/${initialValues.feeds_type}/${initialValues.attack_type}/${initialValues.prioritize}.json`,
-  );
+  // I batched url, tableParams, and tableKey into a single state object so that
+  // all three update atomically in one render cycle. Previously, having
+  // them as separate state variables caused multiple redundent api request before
+  // the state fully settled
```
regulartim: That's fine, but please remove that comment. It does not belong in the code IMO.

frontend/src/components/feeds/Feeds.jsx
```diff
@@ -133,37 +140,33 @@ export default function Feeds() {
...
-      }
-    },
-    [setUrl, navigate],
-  );
+        tableKey: prev.tableKey + 1,
+      }));
+    } catch (e) {
+      console.debug(e);
+    }
+  }, []);
```
regulartim, line 164: You removed `navigate({ search: "" }, { replace: true });` . Isn't that necessary to do what the comment says? Clear ordering and query params? (Just a question)

UsamaElareeny: @regulartim Thanks for taking the time to review. I strongly believe it's better this way since the user can remove the filters as he wants without programmatically removing it for him. Additionally, it keeps applying the filters, while enabling the user to change select values. Also helps with keeping formik as the single source of truth by not interfering in its re-render cycle.

frontend/src/components/feeds/Feeds.jsx
UsamaElareeny: Sure, I’ll remove it. I grouped the state variables to avoid the previous redundant API requests, but I agree the explanation doesn’t need to stay in the code.

PR #902, opened 2026-02-28
feature: normalize credentials into separate Credential model. Closes #668

### LINKED ISSUE #668: Improve scalability of credential lookups in Cowrie sessions
## Problem
`CowrieSession.credentials` is currently stored as an array and queried using substring matching when searching for passwords. On large deployments (here, Honeynet instances with millions of Cowrie sessions), this leads to very expensive queries:

- The database must scan every `CowrieSession` row
- Iterate over each session’s `credentials` array
- Perform string comparisons for each credential

Because of this, password lookups degrade to full table scans. A plain GIN index on the array does not help with substring matching, so query performance does not scale with data size.

rahulgunwanistudy-2005:
# Description
`CowrieSession.credentials` was stored as a PostgreSQL `ArrayField` of strings in `"username | password"` format. This caused full table scans when searching for specific passwords, making credential lookups unscalable on large deployments with millions of sessions.

This PR replaces the `ArrayField` with a normalized `Credential` model linked via a `ManyToManyField`, enabling fast indexed lookups by username or password.

### Changes
- Added `Credential` model with separate `username` and `password` fields, indexed independently for fast lookups
- `unique_together` constraint prevents duplicate credentials across sessions
- Updated `CowrieSession.credentials` from `ArrayField` to `ManyToManyField`
- Updated extraction in `cowrie.py` to use `Credential.objects.get_or_create()`
- Updated API view to query credentials via the new relationship
- Data migration preserves all existing credentials from the old `ArrayField`

## Related issues
Closes #668

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# Checklist
- [x] I have read and understood the rules about how to Contribute to this project
- [x] The pull request is for the branch `develop`
- [ ] I have added documentation of the new features
- [x] Linter (`Ruff`) gave 0 errors
- [x] I have added tests for the feature I solved. All 539 tests passed with 0 errors
- [ ] If changes were made to an existing model/serializer/view, the docs were updated and regenerated (N/A)
- [ ] If the GUI has been modified (N/A)

### DIFF
--- a/greedybear/migrations/0042_credential_model_and_data_migration.py
+++ b/greedybear/migrations/0042_credential_model_and_data_migration.py
@@ -31,7 +31,7 @@ def migrate_credentials(apps, schema_editor):
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("greedybear", "0040_alter_tag_key"),
+        ("greedybear", "0041_sharetoken"),
     ]
 
     operations = [
--- a/tests/test_migrations.py
+++ b/tests/test_migrations.py
@@ -119,3 +119,33 @@ class TestRemoveUnusedLog4pot(MigrationTestCase):
             hp_new.objects.filter(name="Log4pot").exists(),
             "Log4pot with IOCs should NOT be deleted",
         )
+
+
+@tag("migration")
+class TestCredentialModelMigration(MigrationTestCase):
+    """Tests that credentials are correctly migrated from ArrayField to Credential model."""
+
+    migrate_from = "0041_sharetoken"
+    migrate_to = "0042_credential_model_and_data_migration"
+
+    def test_credentials_migrated_to_credential_model(self):
+        IOC = self.old_state.apps.get_model(self.app_name, "IOC")
+        CowrieSession = self.old_state.apps.get_model(self.app_name, "CowrieSession")
+
+        ioc = IOC.objects.create(name="1.2.3.4", type="ip")
+        session = CowrieSession.objects.create(
+            session_id=1,
+            source=ioc,
+            credentials=["root | password123", "admin | admin"],
+        )
+
+        new_state = self.apply_tested_migration()
+        Credential = new_state.apps.get_model(self.app_name, "Credential")
+        CowrieSession = new_state.apps.get_model(self.app_name, "CowrieSession")
+
+        self.assertEqual(Credential.objects.count(), 2)
+        self.assertTrue(Credential.objects.filter(username="root", password="password123").exists())
+        self.assertTrue(Credential.objects.filter(username="admin", password="admin").exists())
+
+        session = CowrieSession.objects.get(session_id=1)
+        self.assertEqual(session.credentials.count(), 2)


rahulgunwanistudy-2005:
Hi @regulartim, I've submitted this PR to address issue #668.

The implementation:
- Adds a new `Credential` model with separate `username` and `password` fields, each independently indexed for fast lookups
- Replaces the `ArrayField` on `CowrieSession` with a `ManyToManyField` to the new `Credential` model
- Includes a data migration that preserves all existing credentials from the old `ArrayField`
- Updates the extraction pipeline and API view accordingly

All 539 tests pass with 0 linter errors. Please let me know if you have any feedback!

regulartim:
Hey @rahulgunwanistudy-2005 ! Your commit history looks quite messy. I suspect that is because you based your feature branch on `main` instead of `develop`. Could you please fix that?

rahulgunwanistudy-2005:
Hi @regulartim, I'm really very sorry for the mess I created and didn't clean it. I've rebased the branch onto develop, it now contains only 2 clean commits.

Please let me know if there exists any other problem.
Sorry, once again

regulartim:
Please make sure the CI is happy, then I'll review! :)

rahulgunwanistudy-2005:
Yes sir I am already working on it, just give me some time.

rahulgunwanistudy-2005:
Hi @regulartim, fixed the formatting issue added the missing blank line between classes in models.py. 
CI should pass now.

regulartim:
Unfortunately, there are merge conflicts now that you also need to address.

rahulgunwanistudy-2005:
Hi @regulartim, I've rebased onto the latest develop and resolved the conflicts. 
The formatting issue is also fixed.

greedybear/cronjobs/extraction/strategies/cowrie.py
```diff
@@ -244,7 +244,9 @@ def _process_session_hit(self, session_record: CowrieSession, hit: dict, ioc: IO
                 session_record.login_attempt = True
                 username = normalize_credential_field(hit["username"])
                 password = normalize_credential_field(hit["password"])
-                session_record.credentials.append(f"{username} | {password}")
+                credential, _ = Credential.objects.get_or_create(username=username, password=password)
+                session_record.credentials.add(credential)
+                session_record.source.login_attempts += 1
```
regulartim: `session_record.source.login_attempts += 1` was just removed in a recent PR (for good reason). Please don't add it again.

greedybear/migrations/0041_credential_model_and_data_migration.py
```diff
@@ -0,0 +1,74 @@
...
+        for credential_str in session.old_credentials or []:
+            try:
+                username, password = credential_str.split(" | ", 1)
+                credential, _ = Credential.objects.get_or_create(
+                    username=username,
+                    password=password,
+                )
+                session.credentials.add(credential)
+            except ValueError:
+                continue
```
regulartim: This will be very, very slow on large deployments! Please try to find a way to make this more efficient. Some hints:
- there should be something like `CowrieSession.objects.iterator`
- I think it is possible to bulk-create credentials first, then bulk adding them to the session object(s)

greedybear/models.py
```diff
@@ -110,15 +110,27 @@ def __str__(self):
...
+class Credential(models.Model):
+    username = models.CharField(max_length=256, blank=False)
+    password = models.CharField(max_length=256, blank=False)
+
+    class Meta:
+        unique_together = [("username", "password")]
+        indexes = [
+            models.Index(fields=["username"]),
+            models.Index(fields=["password"]),
+        ]
```
regulartim, line 122: `unique_together` might soon be deprecated:
https://docs.djangoproject.com/en/6.0/ref/models/options/#unique-together

rahulgunwanistudy-2005:
Sure @regulartim sir, I will start working on all the points you mentioned... 
Thank you so much :)

rahulgunwanistudy-2005:
Hi @regulartim sir, I have addressed all the review feedback...
Please review it and let me know if you have any further feedback.

greedybear/models.py
rahulgunwanistudy-2005: Replaced unique_together with UniqueConstraint as unique_together.

greedybear/migrations/0041_credential_model_and_data_migration.py
rahulgunwanistudy-2005: Optimized data migration using iterator() and bulk_create(ignore_conflicts=True) for efficient processing on large deployments as you suggested

regulartim:
Hey @rahulgunwanistudy-2005 ! You seem to have some issues with getting your Ruff setup right. You can always run `ruff --config .github/configurations/python_linters/.ruff.toml check .` and `ruff --config .github/configurations/python_linters/.ruff.toml format .` from the project root folder and ruff should fix everything for you automatically.

rahulgunwanistudy-2005:
Thank you so much sir @regulartim 
I was indeed facing too much issue with it... and I am sorry for I created a bit of a mess in the commit history.
This won't happen next time.

greedybear/cronjobs/extraction/strategies/cowrie.py
```diff
@@ -251,7 +251,8 @@ def _process_session_hit(self, session_record: CowrieSession, hit: dict, ioc: IO
                 session_record.login_attempt = True
                 username = normalize_credential_field(hit["username"])
                 password = normalize_credential_field(hit["password"])
-                session_record.credentials.append(f"{username} | {password}")
+                credential, _ = Credential.objects.get_or_create(username=username, password=password)
+                session_record.credentials.add(credential)
```
regulartim: For consistency and better testability I would prefer if we would use a method in the CowrieSessionRepository to create this. You can add method there if it does not exist yet.

api/views/cowrie_session.py
regulartim, line 1: Could you please add `"credentials"` to the `prefetch_related` calls on CowreiSession in this file? Otherwise we would trade  one performance problem for another.

greedybear/migrations/0041_credential_model_and_data_migration.py
regulartim: I am sorry but I don't think this is sufficient performance wise. On our honeynet instance we currently have ~5 million cowrie sessions. Your current migration does 3 queries per CowrieSession object, this will result in 15 million queries. So executing this migration on our server would take hours if not days. This is not feasible. 

We need to find a way to do this much more efficiently. This is the core challenge in this PR / issue. 

To be clear, this is a quite difficult task, not beginner friendly at all. So if you want to step back from it, I am totally fine with that!

regulartim (changes requested):

api/views/cowrie_session.py
rahulgunwanistudy-2005: Sure sir, I will add "credentials" to all prefetch_related calls in cowrie_session.py to avoid N+1 queries when accessing credentials.

greedybear/cronjobs/extraction/strategies/cowrie.py
rahulgunwanistudy-2005: Agreed sir, I will add an add_credential method to CowrieSessionRepository and update cowrie.py to use it instead of calling Credential.objects directly.

rahulgunwanistudy-2005:
@regulartim 
Thank you for the detailed explanation and for being so understanding. 
But, I won't step back from this, I want to solve it properly...
I have a more efficient approach in mind that minimises the total number of queries regardless of session count, and I just need a little time to implement and test it carefully.

Thank you for the opportunity to keep working on this!!!

regulartim:
> I have a more efficient approach in mind that minimises the total number of queries regardless of session count, and I just need a little time to implement and test it carefully.

Alright, looking forward to see this! :)

rahulgunwanistudy-2005:
@regulartim
Hi Sir...
I have been thinking about more efficient approaches for the migration...
Initially I considered two approaches a two pass python approach (5 queries total) and a chunked approach ( about 15,000 queries for 5M sessions) but both rely on accessing the m2m through table in migration context which I thought may have memory and speed issues.
Then I thought of a better solution using raw sql with postgresql unnest... Since GreedyBear already uses postgresql, unnest is available. 
This is just 2 queries total, no Python loops, no memory concerns, and no through table uncertainty in migration context.
What do you think about this??

regulartim:
Hey @rahulgunwanistudy-2005 ! Thanks for your investigation. :) I never heard of postgres unnest, but it seems to be the right tool for the job. Normally I really dislike raw SQL in a django application, but in this case I think I am fine with it.

rahulgunwanistudy-2005:
Thank you for the feedback @regulartim ...
So my approach is...
Using schema_editor.execute() to run the two raw SQL queries directly in the migration. 

The first query uses unnest to extract all unique credential pairs from old_credentials and bulk inserts them into greedybear_credential. 

The second query then builds all M2M relationships by joining back to get the credential PKs and bulk inserts into the through table greedybear_cowriesession_credentials. 

This keeps everything at the database level with just 2 queries total.

But I have two questions though...

1. for running the raw SQL in the migration, should I use migrations.RunSQL directly in the operations list, or schema_editor.execute() inside a RunPython function?
I want to make sure I'm using the most appropriate Django pattern here...

2. I noticed a potential edge case, passwords containing | (the separator used in old_credentials). The original Python code handled this correctly with split(" | ", 1), but split_part(cred, ' | ', 2) in PostgreSQL would only get the second part, not everything after the first |. I plan to use substring(cred from position(' | ' in cred) + 3) instead to correctly capture the full password.

regulartim:
Hey @rahulgunwanistudy-2005 ! Sounds great! :)

> for running the raw SQL in the migration, should I use migrations.RunSQL directly in the operations list, or schema_editor.execute() inside a RunPython function?
I want to make sure I'm using the most appropriate Django pattern here...

I think schema_editor.execute() inside a RunPython is the way to go here. 

> I noticed a potential edge case, passwords containing | (the separator used in old_credentials). The original Python code handled this correctly with split(" | ", 1), but split_part(cred, ' | ', 2) in PostgreSQL would only get the second part, not everything after the first |. I plan to use substring(cred from position(' | ' in cred) + 3) instead to correctly capture the full password.

I think we can ignore this edge case. The probability that a password contains " | " is quite low. As long as the code does not crash on the edge case, I am fine with using split_part.

rahulgunwanistudy-2005:
Thank you @regulartim !!!
That clears everything up. I will use schema_editor.execute() inside RunPython and keep split_part as is. 
Will get the implementation done shortly...

regulartim:
You're welcome! :) Take your time.

rahulgunwanistudy-2005:
Hi @regulartim, I have addressed all the feedback...

Replaced the Python-based migration with a PostgreSQL unnest approach, just 2 queries total regardless of session count
Added "credentials" to all prefetch_related calls in cowrie_session.py to avoid N+1 queries
Moved credential creation into a add_credential method in CowrieSessionRepository for consistency and better testability

All 596 tests pass. 
Please let me know if you have any further feedback :)

regulartim:
Hey @rahulgunwanistudy-2005 ! This looks very promising, thanks a lot! :+1: 

I would like to test the migration on real data before merging the PR. Two things should be done in advance:
1. Write an test for the migration. There is a specific file for that `test_migrations.py`. You can run migration tests with this command: `docker exec greedybear_uwsgi python3  manage.py test --tag=migration`.
2. Because of a recent merge there is currently a migration numbering conflict. That's why CI fails. Please fix that (=increment the migration number and its dependency).

rahulgunwanistudy-2005:
@regulartim 
Hi sir, I have addressed both the points:

1. Renamed migration to 0042 to resolve the numbering conflict.
2. Added a migration test in test_migrations.py that verifies credentials are correctly migrated from the old ArrayField     into the new Credential model and M2M relationship

All 684 tests pass locally. 
Please let me know if you have any further feedback...

rahulgunwanistudy-2005:
I have also verified the migration tests pass in the Docker environment...

(venv) Rahuls-MacBook-Pro% docker exec greedybear_uwsgi python3 manage.py test --tag=migration

Running migration tests

Found 5 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
.....
----------------------------------------------------------------------
Ran 5 tests in 16.074s

OK
Destroying test database for alias 'default'...

rahulgunwanistudy-2005:
Thank you so much, @regulartim! 
I am really glad the PostgreSQL approach held up well on real data, I am so happy to hear that!!! :))

Also, I wanted to ask you one thing...
I am planning to apply for GSoC 2026 with GreedyBear. Would you be open to mentoring my proposal?

regulartim:
> I am planning to apply for GSoC 2026 with GreedyBear. Would you be open to mentoring my proposal?

Yes, theoretically I am open to do so. However I am expecting a lot of proposals and I will only be able to mentor one or two. So can't promise anything.

rahulgunwanistudy-2005:
Understood, Thank You!!
I will put my best effort into the proposal :)

PR #904, opened 2026-02-28
Migrate from uWSGI to gunicorn. Closes #891

### LINKED ISSUE #891: uWSGI is in maintenance-only mode
A note on the [docs of uWSGI](https://uwsgi-docs.readthedocs.io/en/latest/) says:

> The project is in maintenance mode (only bugfixes and updates for new languages apis). Do not expect quick answers on github issues and/or pull requests (sorry for that) A big thanks to all of the users and contributors since 2009.

I think it's time to move to a different server gateway interface. 

Possible candidates:
- daphne
- uvicorn
- granian
- gunicorn

I personally have no experience with any of them. From my research I tend towards [granian](https://github.com/emmett-framework/granian) as it has ASGI support, it is Rust based and it is fast.

What do you think @mlodic ?

SupRaKoshti:
# Description
Migrated the server gateway interface from uWSGI (which is now in maintenance-only mode) to Gunicorn. Gunicorn is a well-established WSGI server and a drop-in replacement for uWSGI for standard Django applications. The server runs on `0.0.0.0:8001` with 16 workers, matching the previous uWSGI configuration.

Changes made:

**Python Requirements**
- Replaced `uwsgi` and `uwsgitop` with `gunicorn==25.1.0` in `requirements/project-requirements.txt`

**Docker**
- Created `docker/entrypoint_gunicorn.sh` (replacing `docker/entrypoint_uwsgi.sh`) with the gunicorn server command
- Updated `docker/Dockerfile` to remove uWSGI log directory and related comments
- Updated `docker/default.yml` — renamed `uwsgi` service to `gunicorn`, updated container name, entrypoint, command, and removed uWSGI-specific volume mount and stats port
- Updated `docker/local.override.yml` — renamed `uwsgi` service to `gunicorn`
- Updated `docker/stag.override.yml` — renamed `uwsgi` service to `gunicorn`
- Updated `docker/version.override.yml` — renamed `uwsgi` service to `gunicorn`
- Updated `docker/elasticsearch.yml` — renamed `uwsgi` service to `gunicorn`

**Nginx Configuration**
- Updated `configuration/nginx/https.conf` — replaced `uwsgi_pass`/`uwsgi_cache`/`uwsgi_params` with `proxy_pass`/`proxy_cache` equivalents (protocol change from uWSGI binary to HTTP)
- Updated `configuration/nginx/http.conf` — same uwsgi→proxy conversion as `https.conf`
- Updated `configuration/nginx/django_server.conf` — updated upstream hostname and comment

**Management Script**
- Updated `gbctl` — renamed all `uwsgi` container references to `gunicorn` (affects `check_downgrade`, `cmd_logs`, `cmd_health`, and `cmd_create_admin` functions)

**Deleted Files**
- Deleted `docker/entrypoint_uwsgi.sh`
- Deleted `configuration/uwsgi/greedybear.ini`

### Related issues
Closes #891

### Type of change

# Checklist
### Formalities

### Docs and tests

### GUI changes
N/A - No GUI changes were made.

### DIFF
--- a/configuration/nginx/http.conf
+++ b/configuration/nginx/http.conf
@@ -1,4 +1,4 @@
-# the upstream components nginx needs to connect to
+# the upstream component nginx needs to connect to
 upstream django_main {
     server unix:/run/gunicorn/main.sock fail_timeout=30s;
 }
--- a/configuration/nginx/https.conf
+++ b/configuration/nginx/https.conf
@@ -1,4 +1,4 @@
-# the upstream components nginx needs to connect to
+# the upstream component nginx needs to connect to
 upstream django_main {
     server unix:/run/gunicorn/main.sock fail_timeout=30s;
 }


SupRaKoshti:
Hi @regulartim, I have completed the migration from uWSGI to gunicorn. Could you please review the PR when you get a chance?

Also, I noticed the documentation at https://intelowlproject.github.io/docs/GreedyBear/Contribute/ still references the old `greedybear_uwsgi` container name.

image

docker/default.yml
```diff
@@ -54,16 +52,10 @@ services:
...
-      test: ["CMD-SHELL", "curl -f http://localhost/hc || exit 1"]
-      interval: 10s
-      timeout: 2s
-      retries: 3
-      start_period: 10s
-      start_interval: 1s
+    depends_on:
+      - gunicorn
+
+
```
regulartim, line 67: If you don't have a good reason for the removal, I would like to keep the condition and the healthcheck!

docker/default.yml
```diff
@@ -15,27 +15,25 @@ services:
       start_period: 10s
       start_interval: 1s
 
-  uwsgi:
+  gunicorn:
```
regulartim: I would suggest renaming the services from "uwsgi" to "app". Also in the other compose files.

docker/default.yml
```diff
@@ -15,27 +15,25 @@ services:
...
     image: intelowlproject/greedybear:prod
-    container_name: greedybear_uwsgi
+    container_name: greedybear_gunicorn
```
regulartim: I would prefer "greedybear_app".

docker/Dockerfile
```diff
@@ -42,7 +41,7 @@ COPY . $PYTHONPATH
 COPY --from=frontend-build /app/build /var/www/reactapp
 
 # separation is required to avoid to re-execute os installation in case of change of python requirements
-RUN mkdir -p ${LOG_PATH}/django ${LOG_PATH}/uwsgi \
+RUN mkdir -p ${LOG_PATH}/django \
```
regulartim: Where will gunicorn create its logs ?

regulartim (changes requested):
Hey @SupRaKoshti ! Thanks for your work. Aside from the other comments, I checked the[ Gunicorn guide for docker deployments](https://gunicorn.org/guides/docker/). They do some things a little different and I think we should match that in our `default.yml`: 
- [Dynamic calculation for the worker count](https://gunicorn.org/guides/docker/#worker-count)
- [Health Checks](https://gunicorn.org/guides/docker/#health-checks)
- [Graceful Shutdown](https://gunicorn.org/guides/docker/#graceful-shutdown)

What do you think of that?

regulartim:
There is also support for the [uWSGI Protocol](https://gunicorn.org/uwsgi/) in Gunicorn, which seems to have advantages over http. Should be maybe use that?

docker/default.yml
SupRaKoshti: Sure, I will rename service from "gunicorn" to "app" in all compose files.

docker/default.yml
SupRaKoshti: Sure I will rename it.

docker/default.yml
SupRaKoshti: You are right, I will restore the condition and healthcheck. I removed it by mistake.

docker/Dockerfile
SupRaKoshti: The `${LOG_PATH}/uwsgi` directory was specifically created for uWSGI logs. 
Since we are removing uWSGI, this directory is no longer needed. 
Gunicorn logs go to stdout/stderr by default, so no separate log directory is required for gunicorn.
Should I keep it that way or create a specific log directory?

SupRaKoshti:
> Hey @SupRaKoshti ! Thanks for your work. Aside from the other comments, I checked the[ Gunicorn guide for docker deployments](https://gunicorn.org/guides/docker/). They do some things a little different and I think we should match that in our `default.yml`:
> 
> * [Dynamic calculation for the worker count](https://gunicorn.org/guides/docker/#worker-count)
> * [Health Checks](https://gunicorn.org/guides/docker/#health-checks)
> * [Graceful Shutdown](https://gunicorn.org/guides/docker/#graceful-shutdown)
> 
> What do you think of that?

Hi @regulartim , that sounds great! I will look into the gunicorn Docker deployment guide and implement dynamic worker count calculation, health checks, and graceful shutdown. Give me some time to read through the documentation and I will update the PR accordingly.

regulartim:
Hey @SupRaKoshti ! :) Please also take a look at the uwsgi protocol support I mentioned:

> There is also support for the [uWSGI Protocol](https://gunicorn.org/uwsgi/) in Gunicorn, which seems to have advantages over http. Should be maybe use that?

If I understand it correctly using it would be more efficient and result in a smaller diff (=less changes to the code).

docker/Dockerfile
regulartim: I think we should create a specific log directory for gunicorn, the same way we did for uwsgi.

regulartim:
> Also, I noticed the documentation at https://intelowlproject.github.io/docs/GreedyBear/Contribute/ still references the old greedybear_uwsgi container name.

Yeah, I'll just remove the command, it is outdated anyway.

SupRaKoshti:
Hey @regulartim! I have gone through the official Gunicorn documentation for Docker deployment and uWSGI protocol and implemented everything you suggested. Here's a summary of what was done:

**Dynamic Worker Count**
- Implemented using `$((2 * $(nproc) + 1))` at runtime so workers automatically scale based on available CPUs in the container

**Health Checks**
- Since port 8001 now speaks uWSGI binary protocol (not HTTP), `curl` would fail
- Replaced with a Python TCP socket check in both `default.yml` and `Dockerfile`:
  ```
  python3 -c "import socket; socket.create_connection(('localhost', 8001), timeout=2)"
  ```

**Graceful Shutdown**
- Added `--graceful-timeout 30` and `--timeout 120` to the gunicorn command
- Added `stop_grace_period: 30s` to the `app` service to match

**uWSGI Protocol**
- Added `--protocol uwsgi` to the gunicorn command
- Updated `http.conf` and `https.conf` to use `uwsgi_pass`, `uwsgi_cache` and `uwsgi_params` instead of `proxy_pass`/`proxy_cache`
- `django_server.conf` intentionally kept as `proxy_pass` since it's only used in local dev with `manage.py runserver` which speaks plain HTTP

**Gunicorn Log Directory**
- Added `${LOG_PATH}/gunicorn` directory creation in Dockerfile, same as we had for uWSGI

**Service Renamed**
- Renamed service from `gunicorn` to `app` across all compose files and updated all references in `gbctl`, `nginx depends_on` and `qcluster depends_on`

Ready for review! 🙏

SupRaKoshti:
Also @regulartim, I noticed that the `app` service currently runs as root, while `qcluster` already uses `user: "2000:82"`. The Dockerfile already sets up `www-data` (uid 2000) and chowns all necessary directories, but `USER www-data` is never set and `app` service has no `user:` defined.

Should we implement the non-root user for the `app` service as well?

**Pros:**
- Security best practice — containers should not run as root
- Consistent with `qcluster` which already uses `user: "2000:82"`
- Recommended by the official Gunicorn Docker docs
- All file permissions are already correctly set up in the Dockerfile

**Cons / Risks:**
- `entrypoint_gunicorn.sh` runs `chown -R 2000:82 /var/log/greedybear` which requires root — this would need to be handled differently if we switch to non-root

Happy to implement this in a follow-up PR if you think it's worth doing!

regulartim:
> Also @regulartim, I noticed that the app service currently runs as root, while qcluster already uses user: "2000:82". The Dockerfile already sets up www-data (uid 2000) and chowns all necessary directories, but USER www-data is never set and app service has no user: defined.
Should we implement the non-root user for the app service as well?

Yeah, I also thought about this. I would prefer to do that in a separate issue / PR though (feel free to open an issue). I am a bit scared that such a change might break existing instance.

docker/Dockerfile
```diff
@@ -53,6 +52,9 @@ RUN mkdir -p ${LOG_PATH}/django ${LOG_PATH}/uwsgi \
     && chown -R www-data:www-data ${LOG_PATH} /opt/deploy/ ${PYTHONPATH}/mlmodels/ \
     && rm -rf frontend/
 
+HEALTHCHECK --interval=30s --timeout=10s --start-period=120s --retries=3 \
+    CMD python3 -c "import socket; socket.create_connection(('localhost', 8001), timeout=2)" || exit 1
+
 ## ------------------------------- Development Stage ------------------------------ ##
```
regulartim: Since we already have the healthcheck in the compose file, this is redundant. Agree?

docker/default.yml
```diff
@@ -15,27 +15,26 @@ services:
...
       - "8001"
-      - "1717"
     env_file:
       - env_file
     depends_on:
       postgres:
         condition: service_healthy
     healthcheck:
-      test: ["CMD-SHELL", "curl -f http://localhost:1717 || exit 1"]
+      test: ["CMD-SHELL", "python3 -c \"import socket; socket.create_connection(('localhost', 8001), timeout=2)\" || exit 1"]
```
regulartim: The healthcheck isn't optimal, I think (sorry for being nit-picky here). I see two better options:
- using basically your approach but with nc instead of python
- using a separate http endpoint for the healthcheck (this seems to be the propper way do do it, see https://github.com/benoitc/gunicorn/issues/1417 )

regulartim:
Also, there seems to be a bug! When I try to run the container, I get: 
```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "./docker/entrypoint_gunicorn.sh": permission denied
```

The reason might be that the file permissions of `entrypoint_gunicorn.sh` changed due to the renaming. Please check that. Maybe the executable bit is missing.

docker/Dockerfile
SupRaKoshti: I added the `HEALTHCHECK` in the Dockerfile because the official Gunicorn Docker documentation recommends it as a best practice — the idea being that the image is self-contained and knows how to check its own health even when run standalone via `docker run`.

However, you're right that it's redundant in this project because:
- This project always runs via Docker Compose
- The `default.yml` healthcheck completely overrides the Dockerfile one
- So the Dockerfile `HEALTHCHECK` is never actually used here

I'll remove it from the Dockerfile and keep it only in `default.yml` where it actually matters. Thanks for catching that!

docker/default.yml
SupRaKoshti: Thanks for suggestions! I will look into both approaches and update PR accordingly!

SupRaKoshti:
Hey @regulartim! I have implemented all the changes and suggestions you mentioned:

- Removed the redundant `HEALTHCHECK` from the Dockerfile since `default.yml` healthcheck already overrides it when running via Docker Compose
- Implemented the dual bind approach for the healthcheck — gunicorn now binds to port 8001 (uWSGI binary for nginx traffic) and port 8002 (plain HTTP for healthcheck only), so healthcheck requests bypass the uWSGI queue completely and use `curl -f http://localhost:8002`
- Fixed the missing executable bit (`+x`) on `entrypoint_gunicorn.sh`

Could you please review and check if everything is running fine? 🙏

regulartim:
Hey @SupRaKoshti ! The application does not even start. How are you testing your changes? 

```
greedybear_app       | [2026-03-10 17:06:39 +0000] [112] [ERROR] Error handling request (no URI read)
greedybear_app       | Traceback (most recent call last):
greedybear_app       |   File "/usr/local/lib/python3.13/site-packages/gunicorn/workers/sync.py", line 141, in handle
greedybear_app       |     req = next(parser)
greedybear_app       |   File "/usr/local/lib/python3.13/site-packages/gunicorn/http/parser.py", line 56, in __next__
greedybear_app       |     self.mesg = self.mesg_class(self.cfg, self.unreader, self.source_addr, self.req_count)
greedybear_app       |                 ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
greedybear_app       |   File "/usr/local/lib/python3.13/site-packages/gunicorn/uwsgi/message.py", line 68, in __init__
greedybear_app       |     unused = self.parse(self.unreader)
greedybear_app       |   File "/usr/local/lib/python3.13/site-packages/gunicorn/uwsgi/message.py", line 104, in parse
greedybear_app       |     raise UnsupportedModifier(self.modifier1)
greedybear_app       | gunicorn.uwsgi.errors.UnsupportedModifier: Unsupported uWSGI modifier1: 71
```

SupRaKoshti:
Hey @regulartim , I apologize for that! I realized my mistake — I was testing with `local.override.yml` the whole time which uses Django's `runserver` instead of gunicorn, so the healthcheck was hitting port 8001 with plain HTTP and working fine on my end.

The root cause of the bug is that `--protocol uwsgi` applies globally to ALL binds including port 8002, so when curl sends an HTTP request to port 8002, gunicorn rejects it with `UnsupportedModifier: 71` (modifier 71 = HTTP request on a uWSGI port).

I am now testing properly with `docker compose -f docker/default.yml up` and working through the issues step by step:

1. Replaced the curl healthcheck with `nc -z localhost 8001` — container is now healthy. The `InvalidUWSGIHeader: incomplete header` errors in logs are just noise since `nc -z` sends no data.

2. Currently investigating `ForbiddenUWSGIRequest` — gunicorn's uWSGI protocol only trusts `127.0.0.1` by default and is rejecting requests from Nginx's Docker network IP, causing the Internal Server Error. Working on the fix now.

Will keep you updated as I progress. Sorry for the inconvenience!

SupRaKoshti:
Hi @regulartim ,

Update: After replacing the healthcheck with `nc -z`, I investigated the `ForbiddenUWSGIRequest` issue further and explored the following solutions:

- **Option A (keep uWSGI):** Fix the IP allowlist with `--forwarded-allow-ips` and properly quote the wildcard to avoid shell expansion issues. Keeps the original uWSGI design but is fragile — any mismatch in headers, IPs, or params breaks everything and is harder to debug.
- **Option B (Unix socket):** Bind gunicorn to a Unix socket to avoid the IP allowlist issue entirely. Needs extra volume mounting and path management.
- **Option C (plain HTTP):** Remove `--protocol uwsgi` from gunicorn and switch nginx from `uwsgi_pass` to `proxy_pass` with `proxy_cache`. Simplest to configure, debug, and maintain.

Option C fully resolves all the errors — no more `ForbiddenUWSGIRequest` or protocol-related issues.

**My question: is the uWSGI protocol a hard requirement here?** If not, I'd like to go with Option C (plain HTTP) as it eliminates all the protocol complexity. Happy to keep uWSGI if there's a specific reason for it!

regulartim:
Hey @SupRaKoshti ! :) UWSGI has some performance benefits over http. Therefore I would prefer to use it. The unix socket approach (option B) sounds smart. That would also have a positive effect on performance, I guess.

SupRaKoshti:
Hi @regulartim , Ready for review! here's a full summary of what i have done.

**What was changed:**
- Migrated Gunicorn from TCP (`0.0.0.0:8001`) to two UNIX sockets:
  - `unix:/run/gunicorn-main.sock` — all application traffic
  - `unix:/run/gunicorn-health.sock` — health checks only (separate socket to avoid queuing behind busy workers, as per your suggestion https://github.com/benoitc/gunicorn/issues/1417)
- Added `--preload` flag to Gunicorn (explained below)
- Added shared `gunicorn_sockets` volume mounted at `/run` in both `app` and `nginx` containers
- Replaced app healthcheck with `test -S /run/gunicorn-health.sock`
- Updated nginx upstreams in `http.conf` and `https.conf` to use the two UNIX sockets
- Added dedicated `/hc` location in nginx routed to `django_health` upstream

**Why UNIX sockets:**
Gunicorn's uWSGI protocol only trusts `127.0.0.1` by default. Since nginx runs in a separate container with its own Docker network IP (`172.20.0.x`), every request was rejected with `ForbiddenUWSGIRequest`. UNIX socket connections bypass the IP allowlist entirely, and also offer better performance by eliminating TCP overhead. See: https://gunicorn.org/uwsgi/#using-unix-sockets

**Why `--preload`:**
Without `--preload`, each Gunicorn worker independently loads the app and calls `get_random_secret_key()`, giving every worker a different `SECRET_KEY`. When a login request is handled by worker A and the next request hits worker B, the `SECRET_KEY` mismatch causes Django to flush the session → user gets redirected to login on every click. With `--preload`, the app loads once in the master process before forking, so all workers inherit the same `SECRET_KEY` via `fork()`. This doesn't happen with the uWSGI server since it pre-loads by default.

**Tested with `docker compose -f docker/default.yml up` — all containers healthy, `http://localhost/` working, admin sessions persisting correctly.**

SupRaKoshti:
> Nice, thanks! I tested it and have some changes that I will push later. You can consider your work done, good job!

Hi @regulartim , Thanks! Glad to hear it's working well!

> Yeah, I also thought about this. I would prefer to do that in a separate issue / PR though (feel free to open an issue). I am a bit scared that such a change might break existing instance.

I’ll do some more research on the non-root users topic, and if it looks like a worthwhile improvement, I’ll open a separate issue for it then we can discuss about it!

regulartim:
> I’ll do some more research on the non-root users topic, and if it looks like a worthwhile improvement, I’ll open a separate issue for it then we can discuss about it!

Cool, thank you!

PR #934, opened 2026-03-03
rDNS-based noise filtering for known scanners. Closes #527

### LINKED ISSUE #527: Reduce noise with more reliability
* generate and periodically update IP list extracted from https://raw.githubusercontent.com/stamparm/maltrail/master/trails/static/mass_scanner.txt and filter those IP addresses out. I know that we already check the reputation from the T-POTs but we cannot depend on external configuration and this should be managed directly from GB for more reliability.

* Use Reverse DNS queries as a filter to remove noise after the previous filter. Keep an internal DB of records if we think that doing DNS queries for every new addition is too expensive. We should have a static list of known domains that we should filter from.

Sahityaaryan:
# Description

  Add a daily reverse DNS cron job that identifies IPs belonging to known scanning
  services (Shodan, Censys, Onyphe, etc.) and marks them as "known scanner" via
  `ip_reputation`. This implements step 2 of #527 — rDNS-based filtering,
  decoupled from the extraction pipeline.

  **What it does:**
  - Queries IOCs with empty `ip_reputation` seen in the last 24h
  - Resolves PTR records with a 2s timeout per IP
  - Caches all results in a `ReverseDNSRecord` table (so IPs are never re-queried)
  - Matches PTR hostnames against a curated `KNOWN_SCANNER_DOMAINS` list
  - Updates matching IOCs to `ip_reputation = "known scanner"`
  - Feeds exclude "known scanner" by default (same pattern as mass scanner / tor exit node)

  Also fixes `_merge_iocs()` to only overwrite `ip_reputation` when empty, preventing
  extraction from wiping out reputation set by enrichment crons (discussed in #527).

  ### Related issues

  - Closes #527 (step 2: rDNS-based filtering)
  - Builds on #541 (step 1: mass scanner list)

  ### Type of change

  # Checklist

  ### Formalities

  ### Docs and tests

  - [ ] I have checked if my changes affect user-facing behavior that is described in the [docs](https://intelowlproject.github.io/docs/GreedyBear/Introduction/). If so, I also created a pull request in the
  [docs repository](https://github.com/intelowlproject/docs).
  - [x] Linter (`Ruff`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance), it does
  these checks and adjustments on your behalf.

  ### GUI changes

### DIFF
--- a/greedybear/cronjobs/extraction/ioc_processor.py
+++ b/greedybear/cronjobs/extraction/ioc_processor.py
@@ -99,7 +99,7 @@ class IocProcessor:
         existing.interaction_count += new.interaction_count
         existing.related_urls = sorted(set(existing.related_urls + new.related_urls))
         existing.destination_ports = sorted(set(existing.destination_ports + new.destination_ports))
-        existing.ip_reputation = new.ip_reputation
+        existing.ip_reputation = existing.ip_reputation or new.ip_reputation
         existing.asn = new.asn
         existing.firehol_categories = list(new.firehol_categories)
         existing.login_attempts += new.login_attempts
--- a/greedybear/cronjobs/repositories/tag.py
+++ b/greedybear/cronjobs/repositories/tag.py
@@ -50,6 +50,38 @@ class TagRepository:
             self.log.info(f"Created {len(tags_to_create)} tags from source '{source}'")
             return len(tags_to_create)
 
+    def add_tags(self, source: str, tag_entries: list[dict]) -> int:
+        """
+        Incrementally add tags for a source without removing existing ones.
+
+        Unlike replace_tags_for_source, this only creates new tags and
+        leaves previously stored tags intact. Suitable for sources that
+        build up results over many runs (e.g., reverse DNS lookups).
+
+        Args:
+            source: Source name (e.g., "rdns").
+            tag_entries: List of dicts with keys: ioc_id, key, value.
+
+        Returns:
+            Number of tags created.
+        """
+        if not tag_entries:
+            return 0
+
+        tags_to_create = [
+            Tag(
+                ioc_id=entry["ioc_id"],
+                key=entry["key"],
+                value=entry["value"],
+                source=source,
+            )
+            for entry in tag_entries
+        ]
+
+        Tag.objects.bulk_create(tags_to_create, batch_size=1000, ignore_conflicts=True)
+        self.log.info(f"Added {len(tags_to_create)} tags from source '{source}'")
+        return len(tags_to_create)
+
     def get_tags_by_ioc(self, ioc):
         """
         Get all tags for a specific IOC.
--- a/greedybear/cronjobs/reverse_dns.py
+++ b/greedybear/cronjobs/reverse_dns.py
@@ -1,0 +1,189 @@
+import socket
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+from django.db.models import F
+
+from greedybear.consts import MASS_SCANNER_DOMAINS
+from greedybear.cronjobs.base import Cronjob
+from greedybear.cronjobs.repositories import IocRepository
+from greedybear.cronjobs.repositories.tag import TagRepository
+from greedybear.models import IOC, IocType
+
+# Number of concurrent DNS lookups.
+THREAD_POOL_SIZE = 20
+
+# Maximum number of candidate IPs to check per run.
+MAX_CANDIDATES = 500
+
+# Timeout in seconds for each DNS lookup.
+DNS_TIMEOUT = 2
+
+SOURCE_NAME = "rdns"
+
+
+class ReverseDNSCron(Cronjob):
+    """
+    Identify mass scanning services via reverse DNS lookups.
+
+    Runs daily, selects the top candidates most likely to be mass scanners
+    based on behavioral heuristics (persistent, no login attempts, low
+    interaction-to-attack ratio), resolves their PTR records in parallel,
+    and marks matches against a curated list of mass scanner domains.
+    Only IPs with actual PTR records are tagged, so IPs without records
+    are rechecked on subsequent runs.
+    """
+
+    def __init__(self, tag_repo=None, ioc_repo=None):
+        """
+        Initialize the cron job with repository dependencies.
+
+        Args:
+            tag_repo: Optional TagRepository instance for testing.
+            ioc_repo: Optional IocRepository instance for testing.
+        """
+        super().__init__()
+        self.tag_repo = tag_repo if tag_repo is not None else TagRepository()
+        self.ioc_repo = ioc_repo if ioc_repo is not None else IocRepository()
+
+    def run(self) -> None:
+        """
+        Perform reverse DNS lookups on the most probable scanner candidates.
+
+        1. Select top candidates using behavioral heuristics.
+        2. Resolve PTR records in parallel.
+        3. Store non-empty PTR results as tags.
+        4. Update reputation for IPs matching mass scanner domains.
+        """
+        candidates = self._get_candidates()
+
+        if not candidates:
+            self.log.info("No IOCs to check")
+            return
+
+        ip_to_id = {name: ioc_id for ioc_id, name in candidates}
+
+        # Resolve PTR records in parallel
+        ptr_results = self._resolve_batch(list(ip_to_id.keys()))
+
+        # Only tag IPs that have actual PTR records — IPs without PTR
+        # are left untagged so they can be rechecked on future runs.
+        tag_entries = []
+        total_matched = 0
+
+        for ip, ptr in ptr_results.items():
+            if not ptr:
+                continue
+
+            tag_entries.append({"ioc_id": ip_to_id[ip], "key": "ptr_record", "value": ptr})
+
+            if self._matches_scanner_domain(ptr):
+                self._update_ioc(ip)
+                total_matched += 1
+
+        created_count = self.tag_repo.add_tags(SOURCE_NAME, tag_entries)
+        self.log.info(f"Reverse DNS check completed. Checked {len(ptr_results)} IPs, created {created_count} tags, {total_matched} matched mass scanners")
+
+    def _get_candidates(self):
+        """
+        Select the top IOCs most likely to be mass scanners.
+
+        Behavioral heuristics:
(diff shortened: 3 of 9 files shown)


Sahityaaryan:
@regulartim 

  Covers the rDNS-based filtering (step 2) plus the `_merge_iocs()` reputation
  overwrite fix we discussed. 29 new tests, full suite passes.

  Happy to address any feedback.

Sahityaaryan:
@regulartim Great points, both addressed:

  1. Tag model — You're right, I've replaced the custom ReverseDNSRecord model with Tag entries (source="rdns", key="ptr_record"). PTR data is now  automatically available via the existing API endpoints and supports filtering with ?tag_key=ptr_record&tag_value=shodan. No new model or  migration needed.

  2. Parallel resolution — Switched from sequential lookups to  concurrent.futures.ThreadPoolExecutor with 20 workers and batch processing  (500 IPs/batch). With 20k IPs and 10% timeouts, this brings the runtime from   ~1 hour down to a few minutes. Also added TagRepository.add_tags() for  incremental tag creation (as opposed to the atomic replace used by
  ThreatFox/AbuseIPDB).

  Other changes in the rework:
  - Added _merge_iocs() tests for the reputation preservation fix
  - Exception isolation per IP in batch resolution (one failure can't crash the batch)
  - ignore_conflicts=True on bulk_create for crash-recovery safety


  Let me know if anything else needs adjusting

api/views/utils.py
```diff
@@ -111,6 +111,8 @@ def apply_default_filters(self, query_params):
             self.exclude_reputation.append("mass scanner")
         if "include_tor_exit_nodes" not in query_params:
             self.exclude_reputation.append("tor exit node")
+        if "include_known_scanners" not in query_params:
+            self.exclude_reputation.append("known scanner")
```
regulartim, line 114: What is this change for? It changes the way the API behaves , right?

greedybear/cronjobs/reverse_dns.py
```diff
@@ -0,0 +1,169 @@
...
+    def run(self) -> None:
+        """
+        Perform reverse DNS lookups on recent, unclassified IOCs.
+
+        1. Query IOCs: type=IP, ip_reputation empty, last_seen within 24 h,
+           not already tagged by this source.
+        2. Process in batches: resolve PTR in parallel, store as tags,
+           update reputation for matches.
+        """
+        cutoff = timezone.now() - timedelta(hours=24)
```
regulartim: Please use datetime.now() to stay consistent with the rest of the codebase.

api/views/utils.py
Sahityaaryan: Yes, it changes the default API behavior — IPs with ip_reputation="known scanner" (set by the rDNS job) are now excluded from feeds by default, same as "mass scanner" and "tor exit node" on the lines above.
Users can opt-in with ?include_known_scanners in the query params.

Without this, the rDNS job would identify known scanners but they'd still appear in feeds — the "reduce noise" goal from #527 wouldn't be achieved.

That said, if you'd prefer to keep this PR focused on detection only and add the feed exclusion separately, I'm happy to remove it.

regulartim: Ah, now I see! Yeah, please don't change the API right now. The "known scanner" wording is also a bit confusing. Please set the reputation to "mass scanner" and don't introduce a new wording.

greedybear/cronjobs/reverse_dns.py
```diff
@@ -0,0 +1,169 @@
...
+        return False
+
+    def _update_ioc(self, ip_address: str):
+        """
+        Update the IP reputation of an existing IOC to mark it as a known scanner.
+
+        Args:
+            ip_address: IP address to update.
+        """
+        updated = self.ioc_repo.update_ioc_reputation(ip_address, "known scanner")
```
regulartim: Here: please set the reputation to "mass scanner" instead of "known scanner".

Sahityaaryan:
> Hey @Sahityaaryan ! I tested your code and I am still worried about performance. The lookup for 500 IoCs took 15 seconds. Yes, we are now in the "minutes" and not in the "hours" anymore, but this is still a lot of time for a low priority job. I thought about ways to solve that and I have some ideas that I would like to discuss with you:
> 
>     1. We want to find mass scanners. So instead of taking all IOCs from the last 24 hours, we could narrow down the list of candidates to check based on behavior. Mass scanners are seen regularly (so `len(days_seen) > 2`), they have 0 `login_attempts` and they perform not many interactions per attack (something like `interaction_count < 2*attack_count`). So what do you think of taking these metrics to get the 500 most probable candidates and only check them?
> 
>     2. If we can get the runtime down to seconds, we can run the job daily.
> 
>     3. I am not sure anymore if we should track IP addresses with an empty ptr record. If the record ever gets updated, we will miss it. This could be the case for very new mass scanners.
> 
> 
> What do you think of the above?

Great to see that both of us are online :_)

I have made some changes in the local following your review

  1. Behavioral filtering — Implemented. The candidate query now filters on number_of_days_seen > 2, login_attempts = 0, and interaction_count < 2 * attack_count, ordered by -number_of_days_seen and capped at
  500. This should bring the actual candidate count well below 500 in most cases, keeping runtime in the low seconds.
  2. Daily schedule — Already in place (06:07 daily), so no change needed there.
  3. Empty PTR records — Agreed, removed. Only IPs with actual PTR records are tagged now. IPs that don't resolve are left untagged so they're rechecked on the next run.

regulartim:
> Already in place (06:07 daily), so no change needed there.

Sorry, totally missed that!

Sahityaaryan:
No worries! Pushing the updated code now with all the changes we discussed

Sahityaaryan:
@regulartim 
Pushed the updated changes — behavioral filtering, empty PTR removal, and "mass scanner" wording are all in place. Ready for re-review!

greedybear/cronjobs/extraction/ioc_processor.py
```diff
@@ -99,7 +99,7 @@ def _merge_iocs(self, existing: IOC, new: IOC) -> IOC:
         existing.interaction_count += new.interaction_count
         existing.related_urls = sorted(set(existing.related_urls + new.related_urls))
         existing.destination_ports = sorted(set(existing.destination_ports + new.destination_ports))
-        existing.ip_reputation = new.ip_reputation
+        existing.ip_reputation = new.ip_reputation or existing.ip_reputation
```
regulartim: The logic here is reversed, right? The reputations always gets overwritten if "new" is not empty. But if I remember correctly, we wanted to only overwrite when "old" is empty, right?

greedybear/cronjobs/repositories/tag.py
```diff
@@ -50,6 +50,38 @@ def replace_tags_for_source(self, source: str, tag_entries: list[dict]) -> int:
             self.log.info(f"Created {len(tags_to_create)} tags from source '{source}'")
             return len(tags_to_create)
 
+    def add_tags(self, source: str, tag_entries: list[dict]) -> int:
```
regulartim, line 53: Could you maybe add tests for this new method? There already is a file containing tests for the Tags repo.

greedybear/cronjobs/reverse_dns.py
```diff
@@ -0,0 +1,181 @@
...
+
+        Returns:
+            The PTR hostname, or an empty string on any failure.
+        """
+        try:
+            hostname, _, _ = socket.gethostbyaddr(ip)
+            return hostname
+        except (socket.herror, socket.gaierror, TimeoutError, OSError):
+            return ""
+
```
regulartim, line 162: I am a little worried about the default timeouts here. Suggestion:
```python
def _resolve_ptr(self, ip: str) -> str:                                                                                                                                                 
    old_timeout = socket.getdefaulttimeout()
    socket.setdefaulttimeout(2)                                                                                                                                                         
    try:                                                                                                                                                                                
        hostname, _, _ = socket.gethostbyaddr(ip)                                                                                                                                       
        return hostname                                                                                                                                                                 
    except (socket.herror, socket.gaierror, TimeoutError, OSError):
        return ""                                                                                                                                                                       
    finally:                                                                                                                                                                            
        socket.setdefaulttimeout(old_timeout) 
```

What do you think?

Sahityaaryan: Good idea! Since `setdefaulttimeout` is process-global and we're running concurrent threads, I'll set the timeout once before the batch and restore it after, rather than per-call, to avoid any race conditions. Does that work for you?

greedybear/cronjobs/repositories/tag.py
Sahityaaryan: Good call, added tests for add_tags() in the existing tag repo test file.

greedybear/cronjobs/extraction/ioc_processor.py
Sahityaaryan: You're right, the logic should be the other way around — only fill in reputation when the existing one is empty. Fixed to existing.ip_reputation = existing.ip_reputation or new.ip_reputation. Updated the tests accordingly.

greedybear/cronjobs/reverse_dns.py
regulartim: Yep!

Sahityaaryan:
> Thank you, good work! :)
> 
> Please, for future PRs, try to avoid force-pushing changes. It's easier to review when you can explicitly look at the changes that happened since the last review. Force-pushes make that impossible.

sure I will take care of it next time. Thank you very much @regulartim :)

Sahityaaryan:
Hi @regulartim ,

I've been diving deeper into the codebase and found some architectural observations I'd love to discuss. I tried finding you on the Honeynet Discord but there are few people with the similar name - could you let me know your Discord username or just send me a hi over it, so I can DM you, please :)

regulartim:
I am "tim." on discord. But I don't use discord that often. You can also reach out on the Honeynet GSoC Slack.

Sahityaaryan:
@regulartim , 
  Thanks Tim! I tried the Honeynet GSoC Slack but the invite link (gsoc-slack.honeynet.org) seems to have expired. Could you share an updated invite link? In the meantime, I'll DM you on Discord — I'm "Sahityaaryan" there.


image


:(

PR #953, opened 2026-03-04
fix: use correct query param key for format in feeds_advanced

### LINKED ISSUE #949: feeds_advanced endpoint silently ignores the format query parameter
I noticed that the format query parameter on /api/feeds/advanced/ has no effect — the endpoint always returns JSON regardless of what you    
  pass.                                                                                                                                           
The root cause is a key-name mismatch in FeedRequestParams.__init__ (api/views/utils.py:84):                                                 
```  
self.format = query_params.get("format_", "json").lower()
```
  It reads format_ (with underscore) from the query dict, but users naturally send ?format=csv or ?format=txt. The underscore variant exists   
  because the legacy feeds view uses a Python parameter named format_ (to avoid shadowing the builtin) and injects it with that key
  (api/views/feeds.py:48):
```
  feed_params_data.update({"feed_type": feed_type, "attack_type": attack_type, "format_": format_})
```
  That works for the legacy path, but feeds_advanced passes request.query_params directly — so the user's ?format=csv is never found and       
  silently falls back to "json".

  The endpoint's own docstring documents the parameter as format (no underscore), and the serializer field is also named format
  (api/serializers.py:127), confirming the intended name.

  To reproduce: call GET /api/feeds/advanced/?format=txt while authenticated. You get JSON back instead of plain text. The only way to actually
   get non-JSON output today is ?format_=txt, which is undocumented.

  `Fix`: change the dict key in feeds.py:48 to "format" and the query_params.get key in utils.py:84 to "format" — a two-line change.

Sanchit2662:
# Description
I noticed that the feeds_advanced endpoint silently ignores the format query parameter — it always returns JSON no matter what you pass. The issue was a simple key mismatch: FeedRequestParams.__init__ was reading format_ (with an underscore) from the query dict instead of format. This worked fine for the legacy feeds view since it explicitly injected the key as format_, but feeds_advanced passes request.query_params directly — so a user's ?format=csv was never   picked up and it just fell back to json. I fixed it by changing the lookup key to format in both places and updated the docstring to match.     

closes #949

### Type of change

- [ x] Bug fix (non-breaking change which fixes an issue).

# Checklist

### Formalities

- [ x] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/GreedyBear/Contribute/) to this project.
- [ x] I chose an appropriate title for the pull request in the form: `<feature name>. Closes #999`
- [ x] My branch is based on `develop`.
- [ x] The pull request is for the branch `develop`.
- [ x] I have reviewed and verified any LLM-generated code included in this PR.

### Docs and tests

- [ x] I documented my code changes with docstrings and/or comments.
- [ x] I have checked if my changes affect user-facing behavior that is described in the [docs](https://intelowlproject.github.io/docs/GreedyBear/Introduction/). If so, I also created a pull request in the [docs repository](https://github.com/intelowlproject/docs).
- [ x] Linter (`Ruff`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
- [ x] I have added tests for the feature/bug I solved.
- [ x] All the tests gave 0 errors.

### DIFF
--- a/greedybear/settings.py
+++ b/greedybear/settings.py
@@ -111,6 +111,9 @@ REST_FRAMEWORK = {
         "feeds": os.environ.get("FEEDS_THROTTLE_RATE", "30/minute"),
         "feeds_advanced": os.environ.get("FEEDS_ADVANCED_THROTTLE_RATE", "100/minute"),
     },
+    # Disable DRF's format suffix negotiation via ?format= query param,
+    # since feeds endpoints handle the format parameter internally.
+    "URL_FORMAT_OVERRIDE": None,
 }
 
 # Django-Rest-Durin


Sanchit2662:
Hi @regulartim , please review the changes

regulartim:
Hey @Sanchit2662 ! 

> [ x] I have added tests for the feature/bug I solved.

No you did not! Why did you check the box?

Sanchit2662:
You're right @regulartim , that's completely my mistake  I shouldn't have checked that box when I hadn't actually written any tests. Sorry about that. Do you want me to add a test for the format query parameter now and then update the checklist honestly.

regulartim:
> You're right @regulartim , that's completely my mistake I shouldn't have checked that box when I hadn't actually written any tests. Sorry about that. 

No worries, but please be a little more thorough next time. 

> Do you want me to add a test for the format query parameter now and then update the checklist honestly.

Yes, I think adding a test case is the right thing to do here. It should be a test that fails without your changes but passes with your fix in place.

Sanchit2662:
@regulartim  , added the test case.

PR #960, opened 2026-03-05
perf: cache valid feed types using signals to reduce DB load

### LINKED ISSUE #952: Optimization: Caching `get_valid_feed_types()`
I was looking into optimizing `get_valid_feed_types()`, which currently does a live DB query to fetch active honeypots on every single feed API request. There would be only about ~20-30 active honeypots, so fetching them is fast.

### But here is why I think caching them would be a good idea:
All four feed endpoints call `get_valid_feed_types()` on every request, so every time someone hits the feed API, Django checks out a Postgres connection, sends the query over the network, waits for the DB, and then returns the connection. Under heavy load (e.g., thousands of requests syncing feeds), this can lead to **Connection Pool Starvation**. (the issue isn't the query execution time, it's the **network round-trip and connection pool**)

Keeping those 20-30 rows in Django's RAM completely shields the database from this overhead.

### The approach:
To safely fix the performance hit we can use **Admin hook (Instant Invalidation):** We don't need a time-based expiration because we can simply call `cache.delete()` in `greedybear/admin.py` whenever an admin saves or toggles a `GeneralHoneypot`. This guarantees perfect consistency across all workers precisely when the data actually changes.

@regulartim @mlodic Let me know if you think this is a bit too much over-engineering for this specific endpoint, or if we should proceed with the proposed fix!

MShantanu110:
Fixes #952

#Changes:
* Implemented Django `post_save` and `post_delete` signals on `GeneralHoneypot` to instantly clear the `valid_feed_types` cache (`greedybear/signals.py`).
* Registered the signals in `greedybear/apps.py`.
* Added a 15-minute (900s) TTL fallback on the cache in `get_valid_feed_types()` as an extra safety net, aligning with the 10-minute extraction interval.
* This completely shields the database from repeated `active=True` queries on every API hit, preventing connection pool starvation under heavy load.

@regulartim The implementation is ready for review based on our discussion! Let me know if you need any adjustments.

### DIFF
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -112,6 +112,7 @@ class FeedRequestParams:
                 self.min_days_seen = "1"
                 self.ordering = "-expected_interactions"
 
+CACHE_KEY = "valid_feed_types"
 
 def get_valid_feed_types() -> frozenset[str]:
     """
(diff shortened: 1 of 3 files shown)


regulartim:
Hey @MShantanu110 ! Your PR targets the wrong branch and you did not use the PR template. Please read through the [contribution guidelines](https://intelowlproject.github.io/docs/GreedyBear/Contribute/) again.

PR #991, opened 2026-03-06
Fix include_similar in command_sequence_view to union instead of replace. Closes #985

### LINKED ISSUE #985: command_sequence_view: include_similar replaces executed_by results instead of extending them
I was reading through the Cowrie-related API views and noticed a logic bug in command_sequence_view when querying by IP with include_similar=true.

  At lines 68–71 in api/views/command_sequence.py, when include_similar is set, related_iocs gets reassigned instead of extended:
```
  related_iocs = IOC.objects.filter(cowriesession__commands__in=sequences).distinct().only("name")
  if include_similar:
      related_clusters = {s.cluster for s in sequences if s.cluster is not None}
      related_iocs = IOC.objects.filter(cowriesession__commands__cluster__in=related_clusters).distinct().only("name")
```
  This means the base result (IOCs that executed the exact same command sequences) is thrown away and replaced with only the cluster-based result. Two things go wrong:

  1. If none of the matched command sequences have a cluster (i.e. all have cluster=None), then related_clusters is an empty set, the query returns nothing, and executed_by in the      
  response becomes [] — even though without include_similar it would have returned real data. So passing include_similar=true gives you less data, not more.
  2. Even when clusters exist, the IOCs from the original exact-match query are lost if they don't happen to also be in the cluster-based query.

  For comparison, cowrie_session_view handles the same flag correctly using a union:
```
  # cowrie_session_view (correct)
  if include_similar:
      related_sessions = CowrieSession.objects.filter(commands__cluster__in=clusters)
      sessions = sessions.union(related_sessions)  # extends, doesn't replace
```
  The fix would be to union the two querysets instead of reassigning, something like:
```
  if include_similar:
      related_clusters = {s.cluster for s in sequences if s.cluster is not None}
      cluster_iocs = IOC.objects.filter(cowriesession__commands__cluster__in=related_clusters).distinct().only("name")
      related_iocs = related_iocs.union(cluster_iocs)
```

Atul-Chahar:
# Description

When querying `command_sequence_view` by IP with `include_similar=true`, `related_iocs` is reassigned instead of extended, silently dropping the base exact-match results.

The `include_similar` flag at lines 69-71 of `api/views/command_sequence.py` replaces `related_iocs` with a cluster-based query instead of extending it. This causes two problems:

1. If none of the matched command sequences have a cluster, `related_clusters` is empty, the query returns nothing, and `executed_by` becomes `[]` — even though without `include_similar` it would have returned real data.
2. Even when clusters exist, the IOCs from the original exact-match query are lost if they do not also appear in the cluster-based query.

The fix uses `.union()` to combine the base exact-match results with the cluster-based results, matching the correct pattern already used in `cowrie_session_view` (lines 96-100).

### Related issues

Closes #985

### Type of change

# Checklist

### Formalities

### Docs and tests

- [ ] I have checked if my changes affect user-facing behavior that is described in the [docs](https://intelowlproject.github.io/docs/GreedyBear/Introduction/). If so, I also created a pull request in the [docs repository](https://github.com/intelowlproject/docs).

### DIFF
--- a/api/views/command_sequence.py
+++ b/api/views/command_sequence.py
@@ -68,7 +68,8 @@ def command_sequence_view(request):
         related_iocs = IOC.objects.filter(cowriesession__commands__in=sequences).distinct().only("name")
         if include_similar:
             related_clusters = {s.cluster for s in sequences if s.cluster is not None}
-            related_iocs = IOC.objects.filter(cowriesession__commands__cluster__in=related_clusters).distinct().only("name")
+            cluster_iocs = IOC.objects.filter(cowriesession__commands__cluster__in=related_clusters).distinct().only("name")
+            related_iocs = related_iocs.union(cluster_iocs)
         if not seqs:
             raise Http404(f"No command sequences found for IP: {observable}")
         data = {


Atul-Chahar:
Closing this PR — I didn't realize @Sanchit2662 had already requested assignment and had been assigned by the maintainer. Apologies for stepping on your toes! Your PR #992 looks great.

PR #999, opened 2026-03-09
feat: expose IoC-Sensor relationship in API responses

aashnaachaudhary10:
### Description
This PR addresses the requirement to expose the relationship between Indicators of Compromise (IoCs) and their reporting Sensors in the API responses. By updating the `IOCSerializer`, we now include a `sensors` field that lists the addresses of all sensors associated with a specific IoC using a `SlugRelatedField`.

This branch has been re-created based on the latest **`develop`** branch to ensure a clean commit history and compatibility with the current project structure.

### Related Issues
* Closes #781

### Type of Change
- [x] New feature (non-breaking change which adds functionality)

### Checklist
- [x] My branch is based on **`develop`**
- [x] I have verified the changes manually using the Django shell
- [x] Linter (Ruff) gave 0 errors

### Verification Proof (Manual Testing)
I verified the serializer output within the Docker environment (`greedybear_uwsgi` container):
Screenshot 2026-03-09 204322

```python
from api.serializers import IOCSerializer
from greedybear.models import IOC, Sensor

# Test data setup
s, _ = Sensor.objects.get_or_create(address="10.0.0.1")
i, _ = IOC.objects.get_or_create(name="1.2.3.4", type="ip")
i.sensors.add(s)

# Serialization test
data = IOCSerializer(i).data
print(data['sensors'])
# Output: ['10.0.0.1']

Issue #1008, opened 2026-03-10
Optimize N+1 DB Queries in IOC Extraction Pipeline


manik3160:
## Description

While exploring the extraction pipeline (specifically [greedybear/cronjobs/extraction/utils.py](file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py)), I noticed a classic **N+1 query problem** when processing Elasticsearch hits into IOC objects.

In [iocs_from_hits()](file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py#90-158), the code groups incoming hits by IP and then iterates through each IP (which can be thousands per batch). Inside this loop, it triggers multiple individual database queries per IP:

1. **[get_firehol_categories(ip, extracted_ip)](file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py#54-88)** ([utils.py#L54-L87](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/extraction/utils.py#L54-L87))
   - `FireHolList.objects.filter(ip_address=ip)` — one query per IP
   - `FireHolList.objects.filter(ip_address__contains="/")` — queries **all** CIDR entries on every iteration

2. **[correct_ip_reputation(ip, ...)](file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py#31-52)** ([utils.py#L31-L51](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/extraction/utils.py#L31-L51))
   - `MassScanner.objects.get(ip_address=ip)` — one query per IP

3. **[is_whatsmyip_domain(domain)](file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py#14-29)** ([utils.py#L14-L28](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/extraction/utils.py#L14-L28)) (called later in `IocProcessor.add_ioc`)
   - `WhatsMyIPDomain.objects.get(domain=domain)` — one query per domain IOC

Because these lookups happen inside a `for ip, hits in hits_by_ip.items()` loop, processing **5,000 unique IPs** results in **~15,000–20,000 separate database queries** per cronjob run.

## Context / Related Issues
This performance bottleneck was introduced as a side effect of adding IP enrichment tagging from public sources. This issue optimizes and ensures the scalability of the work done for:
[...]

PR #1019, opened 2026-03-11
fix: eliminate N+1 queries in IocRepository.add_honeypot_to_ioc(). Closes #1012

### LINKED ISSUE #1012: N+1 Queries in `IocProcessor.add_ioc()`: `add_honeypot_to_ioc()` Fires Two Avoidable DB Queries Per IOC
## Description

While following up on #1008, I found a related but distinct pair of N+1 queries in the `add_ioc()` hot path — specifically inside `add_honeypot_to_ioc()`. These happen **after** `iocs_from_hits()` returns, so they are orthogonal to the queries in `utils.py` and will not be fixed by that PR.

## Two root causes

`ExtractionPipeline.execute()` calls `strategy.extract_from_hits(hits)` for each honeypot type. Each strategy calls `IocProcessor.add_ioc()` once per IOC. Inside `add_ioc()`, every call to `add_honeypot_to_ioc()` ([ioc_processor.py#L73-74](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/extraction/ioc_processor.py#L73)) triggers two unnecessary DB round trips:

---

### Problem 1 — `ioc.general_honeypot.all()` lazy M2M query per IOC

In [ioc.py#L38-L43](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/repositories/ioc.py#L38):

```python
def add_honeypot_to_ioc(self, honeypot_name: str, ioc: IOC) -> IOC:
    honeypot_set = {hp.name for hp in ioc.general_honeypot.all()}  # ← DB query every IOC
    if honeypot_name not in honeypot_set:
        honeypot = self.get_hp_by_name(honeypot_name)              # ← DB query every new IOC
        ioc.general_honeypot.add(honeypot)
    return ioc
```

`ioc.general_honeypot.all()` fires a separate `SELECT` for the M2M relationship on **every** IOC, because `get_ioc_by_name()` ([ioc.py#L97-L103](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/repositories/ioc.py#L97)) fetches existing IOCs without `prefetch_related`:

```python
def get_ioc_by_name(self, name: str) -> IOC | None:
    try:
        return IOC.objects.get(name=name)   # ← no prefetch_related('general_honeypot')
    except IOC.DoesNotExist:
        return None
```

---

### Problem 2 — `_honeypot_cache` stores booleans, not objects

`IocRepository.__init__()` builds a cache of honeypot data at startup ([ioc.py#L21](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/repositories/ioc.py#L21)):

```python
self._honeypot_cache = {
    self._normalize_name(hp.name): hp.active   # ← only the boolean `active` flag
    for hp in GeneralHoneypot.objects.all()
}
```

When `add_honeypot_to_ioc()` needs the actual `GeneralHoneypot` **object** to call `ioc.general_honeypot.add(honeypot)`, the cache cannot help — it only stores booleans. So the code falls back to `get_hp_by_name()` ([ioc.py#L110-L114](https://github.com/intelowlproject/GreedyBear/blob/main/greedybear/cronjobs/repositories/ioc.py#L110)):

```python
def get_hp_by_name(self, name: str) -> GeneralHoneypot | None:
    return GeneralHoneypot.objects.filter(name__iexact=name).first()  # ← live query per IOC
```

This runs for every IOC where the honeypot is not yet associated — typically **all new IOCs in a batch**.

---

## Impact

With 5,000 unique IPs in a single extraction chunk:

| Operation | Queries |
|---|---|
| `ioc.general_honeypot.all()` per IOC | 5,000 |
| `get_hp_by_name()` per new IOC | up to 5,000 |
| **Total avoidable** | ~5,000–10,000 |

These are completely independent from the queries tracked in #1008 (which targets `FireHolList`, `MassScanner`, and `WhatsMyIPDomain` lookups during `iocs_from_hits()`). Together the two issues account for the majority of DB load in a single extraction run.

---

### Fix 1 — Prefetch M2M in `get_ioc_by_name`

```python
def get_ioc_by_name(self, name: str) -> IOC | None:
    try:
        return IOC.objects.prefetch_related('general_honeypot').get(name=name)
    except IOC.DoesNotExist:
        return None
```

With this change, calling `ioc.general_honeypot.all()` in `add_honeypot_to_ioc()` uses Django's prefetch cache — zero extra queries per IOC.

### Fix 2 — Store `GeneralHoneypot` objects in `_honeypot_cache`

```python
# Before
self._honeypot_cache = {self._normalize_name(hp.name): hp.active for hp in GeneralHoneypot.objects.all()}

# After
self._honeypot_cache = {self._normalize_name(hp.name): hp for hp in GeneralHoneypot.objects.all()}
```

Update the three methods that read from it:

```python
# is_enabled() — same semantics, reads .active from the object
def is_enabled(self, honeypot_name: str) -> bool:
    normalized = self._normalize_name(honeypot_name)
    hp = self._honeypot_cache.get(normalized)
    return hp.active if hp is not None else False

# add_honeypot_to_ioc() — use cache instead of get_hp_by_name()
def add_honeypot_to_ioc(self, honeypot_name: str, ioc: IOC) -> IOC:
    honeypot_set = {hp.name for hp in ioc.general_honeypot.all()}
    if honeypot_name not in honeypot_set:
        honeypot = self._honeypot_cache.get(self._normalize_name(honeypot_name))
        if honeypot is not None:
            ioc.general_honeypot.add(honeypot)
    return ioc

# create_honeypot() — store the object, not the flag
self._honeypot_cache[normalized] = honeypot  # was: honeypot.active
```

With both fixes applied, `add_honeypot_to_ioc()` makes **zero** DB queries per IOC.

---

## Relationship to #1008

#1008 targets the `iocs_from_hits()` phase and reduces queries from ≥15,000 → 4 per batch. This issue targets the `add_ioc()` phase and reduces queries from ~5,000–10,000 → 0 per batch. They are independent and complementary.

Happy to open a PR for this if you agree.

Abhijeet17o:
# Description

Every call to `IocProcessor.add_ioc()` ends with a call to `add_honeypot_to_ioc()`. This method had two separate DB round-trips that fired once per IOC processed, even though both could be resolved from data already available in memory.

**Problem 1: `ioc.general_honeypot.all()` fired a live query per IOC.**
`get_ioc_by_name()` fetched IOCs without `prefetch_related('general_honeypot')`, so `ioc.general_honeypot.all()` inside `add_honeypot_to_ioc()` always triggered a fresh `SELECT` against the M2M through-table. With 5,000 unique IPs in a single extraction chunk, that is 5,000 avoidable queries.

**Problem 2: `_honeypot_cache` stored booleans, not objects.**
The cache built in `__init__` stored only `hp.active` (a boolean). When `add_honeypot_to_ioc()` needed the actual `GeneralHoneypot` object to call `ioc.general_honeypot.add(honeypot)`, the cache could not help, so the code always fell back to `get_hp_by_name()` — another live query per new IOC.

**Fixes applied:**

1. `get_ioc_by_name()`: added `prefetch_related("general_honeypot")` so Django's prefetch cache absorbs the M2M read, bringing the cost to 0 extra queries per IOC.
2. `_honeypot_cache`: changed to store `GeneralHoneypot` objects instead of booleans. `add_honeypot_to_ioc()` now reads the object directly from the cache (with `get_hp_by_name()` as a fallback for safety). `is_enabled()` and `create_honeypot()` updated accordingly.

Together these eliminate up to ~5,000–10,000 queries per extraction chunk depending on the ratio of new vs. existing IOCs, this is independent of the fix in #1008 which targets the `iocs_from_hits()` phase.
Problem 2 alone guarantees ~5,000 fewer queries per extraction run. Problem 1 adds up to another ~5,000 in steady-state runs where most IOCs are already known

### Related issues

- Closes #1012
- Related: #1008 (independent, complementary fix for the `iocs_from_hits()` phase)

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes

No GUI changes.

### DIFF
--- a/greedybear/cronjobs/repositories/ioc.py
+++ b/greedybear/cronjobs/repositories/ioc.py
@@ -18,7 +18,7 @@ class IocRepository:
     def __init__(self):
         """Initialize the repository and populate the honeypot cache from the database."""
         self.log = logging.getLogger(f"{__name__}.{self.__class__.__name__}")
-        self._honeypot_cache = {self._normalize_name(hp.name): hp.active for hp in GeneralHoneypot.objects.all()}
+        self._honeypot_cache = {self._normalize_name(hp.name): hp for hp in GeneralHoneypot.objects.all()}
 
     def _normalize_name(self, name: str) -> str:
         """Normalize honeypot names for consistent cache and DB usage."""
@@ -38,8 +38,9 @@ class IocRepository:
         honeypot_set = {hp.name for hp in ioc.general_honeypot.all()}
         if honeypot_name not in honeypot_set:
             self.log.debug(f"adding honeypot {honeypot_name} to IoC {ioc}")
-            honeypot = self.get_hp_by_name(honeypot_name)
-            ioc.general_honeypot.add(honeypot)
+            honeypot = self._honeypot_cache.get(self._normalize_name(honeypot_name)) or self.get_hp_by_name(honeypot_name)
+            if honeypot is not None:
+                ioc.general_honeypot.add(honeypot)
         return ioc
 
     def create_honeypot(self, honeypot_name: str) -> GeneralHoneypot:
@@ -69,7 +70,7 @@ class IocRepository:
             if honeypot is None:
                 raise e
 
-        self._honeypot_cache[normalized] = honeypot.active
+        self._honeypot_cache[normalized] = honeypot
         return honeypot
 
     def get_active_honeypots(self) -> list[GeneralHoneypot]:
@@ -92,7 +93,7 @@ class IocRepository:
             The matching IOC, or None if not found.
         """
         try:
-            return IOC.objects.get(name=name)
+            return IOC.objects.prefetch_related("general_honeypot").get(name=name)
         except IOC.DoesNotExist:
             return None
 
@@ -129,7 +130,8 @@ class IocRepository:
             True if the honeypot is enabled, False otherwise.
         """
         normalized = self._normalize_name(honeypot_name)
-        return self._honeypot_cache.get(normalized, False)
+        hp = self._honeypot_cache.get(normalized)
+        return hp.active if hp is not None else False
 
     def is_ready_for_extraction(self, honeypot_name: str) -> bool:
         """
--- a/tests/test_ioc_repository.py
+++ b/tests/test_ioc_repository.py
@@ -421,6 +421,47 @@ class TestIocRepository(CustomTestCase):
         self.assertEqual(updated1.recurrence_probability, 0.75)
         self.assertEqual(updated2.recurrence_probability, 0.85)
 
+    # --- Tests for N+1 fix: cache stores GeneralHoneypot objects ---
+
+    def test_honeypot_cache_stores_generalhoneypot_objects(self):
+        """_honeypot_cache must store GeneralHoneypot instances, not booleans."""
+        for key, value in self.repo._honeypot_cache.items():
+            self.assertIsInstance(
+                value,
+                GeneralHoneypot,
+                f"Cache value for '{key}' should be a GeneralHoneypot instance, got {type(value)}",
+            )
+
+    def test_get_ioc_by_name_prefetches_general_honeypot(self):
+        """Accessing general_honeypot on a repo-fetched IOC must not trigger extra DB queries."""
+        ioc = self.repo.get_ioc_by_name("140.246.171.141")
+        self.assertIsNotNone(ioc)
+        with self.assertNumQueries(0):
+            list(ioc.general_honeypot.all())
+
+    def test_add_honeypot_to_ioc_uses_cache_not_db(self):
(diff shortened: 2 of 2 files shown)


Abhijeet17o:
Hey @regulartim, I've made the modifications can you check if everything is correct
Happy to make changes if required!

regulartim:
Hey @Abhijeet17o ! 

> My branch is based on develop.

Exactly as last time: your feature branch is based on `main` which leads to this messy commit history. Please fix that and be more thorough when going through the checklist!

Abhijeet17o:
Hey @regulartim , apologies for the oversight. I’ll fix this right away and ensure it doesn’t happen again. Working on the necessary modifications now

PR #1020, opened 2026-03-11
perf: bulk-prefetch extraction data. Closes #1008

### LINKED ISSUE #1008: Optimize N+1 DB Queries in IOC Extraction Pipeline

manik3160:
# Description

Eliminates ~15,000+ per-run database queries (N+1 queries) in the IOC extraction pipeline by bulk-prefetching `FireHolList`, `MassScanner`, and `WhatsMyIPDomain` data. 

* Bulk-fetches FireHol exact matches and CIDR ranges ahead of the inner loop in [utils.py](file:///Users/manik/GreedyBear/tests/test_extraction_utils.py).
* Bulk-fetches MassScanner IPs into a single O(1) in-memory set.
* Preloads the WhatsMyIPDomain set directly in `IocProcessor.__init__` rather than doing per-domain querying in [add_ioc](file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/ioc_processor.py#29-81).

### Related issues

Closes #1008 

### Type of change

# Checklist

### Formalities

- [x] I chose an appropriate title for the pull request in the form: `perf: bulk-prefetch extraction data. Closes #1008 `

### Docs and tests

- [ ] I have checked if my changes affect user-facing behavior that is described in the [docs](https://intelowlproject.github.io/docs/GreedyBear/Introduction/). If so, I also created a pull request in the [docs repository](https://github.com/intelowlproject/docs).

### DIFF
--- a/greedybear/settings.py
+++ b/greedybear/settings.py
@@ -37,7 +37,7 @@ else:
     print("WARNING!!! You need an ElasticSearch TPOT instance to have the Greedybear to work correctly.")
     if not DEBUG:
         print("you are in production mode: closing the application")
-        pass
+        exit(9)
 
 if ELASTIC_ENDPOINT and not STAGE_CI:
     ELASTIC_CLIENT = Elasticsearch(
--- a/tests/test_extraction_utils.py
+++ b/tests/test_extraction_utils.py
@@ -530,12 +530,7 @@ class IocsFromHitsTestCase(CustomTestCase):
         """Heralding hits with credentials are counted as login attempts."""
         hits = [
             self._create_hit(src_ip="8.8.8.8", hit_type="Heralding", username="root", password="1234"),
-            self._create_hit(
-                src_ip="8.8.8.8",
-                hit_type="Heralding",
-                username="admin",
-                password="admin",
-            ),
+            self._create_hit(src_ip="8.8.8.8", hit_type="Heralding", username="admin", password="admin"),
             self._create_hit(src_ip="8.8.8.8", hit_type="Heralding", username="test", password="test"),
         ]
         iocs = iocs_from_hits(hits)
@@ -756,11 +751,7 @@ class ThreatfoxSubmissionTestCase(ExtractionTestCase):
         mock_honeypot_log4pot.name = "Log4pot"
         mock_honeypot_dionaea = Mock()
         mock_honeypot_dionaea.name = "Dionaea"
-        ioc_record.general_honeypot.all.return_value = [
-            mock_honeypot_cowrie,
-            mock_honeypot_log4pot,
-            mock_honeypot_dionaea,
-        ]
+        ioc_record.general_honeypot.all.return_value = [mock_honeypot_cowrie, mock_honeypot_log4pot, mock_honeypot_dionaea]
         threatfox_submission(ioc_record, ["http://malicious.com/payload.sh"], self.mock_log)
         call_kwargs = mock_post.call_args[1]
         comment = call_kwargs["json"]["comment"]


manik3160:
@regulartim Thanks for the feedback, and apologies for the mistake in the previous PR. I've reviewed the contribution guidelines again and opened this PR against the `develop` branch as required. Please let me know if any further changes are needed.

regulartim (commented):
Hey @manik3160 ! Your code seems to fix the issue, however I do have some problems with it:
- it squeezes a lot of responsibilities into the `iocs_from_hits` function, which is already quite large
- it reduces the overall test coverage
- it makes the code harder to test

Do you agree? Can you think of a better solution?

manik3160:
Hey @regulartim, thanks for the feedback. I completely agree that stuffing everything into [iocs_from_hits](cci:1://file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py:27:0-133:15) made it too bulky and harder to test. 

How about this approach instead:
We keep the standalone functions (`get_firehol_categories`, `correct_ip_reputation`, `is_whatsmyip_domain`) and all their existing unit tests. 

But instead of making internal DB queries, we refactor their arguments to accept the pre-fetched data (dicts/sets). [iocs_from_hits](cci:1://file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py:27:0-133:15) will just do the 3 bulk lookups at the top, and then pass that data into the helper functions during the loop.

This keeps the helpers isolated and properly tested, prevents [iocs_from_hits](cci:1://file:///Users/manik/GreedyBear/greedybear/cronjobs/extraction/utils.py:27:0-133:15) from getting bloated, and still completely eliminates the N+1 queries. 

Does this approach sounds good to you ?

regulartim:
> Does this approach sounds good to you ?

Yes, much better! :+1: 

I thought about a separate `EnrichmentCache` class, that we instantiate once in `iocs_from_hits` and that keeps all the relevant data from Firehol/MassScanners/WIMIP in a internal data structure. This is of course a larger refactoring which would also involve a lot more test adaptions. It is also more memory hungry than your approach I think. 

Both approaches are fine for me, I'll leave it up to you what you would like to implement.

manik3160:
Thanks for the review and feedback !
I'll take a closer look at the options and update the PR accordingly.
Appreciate the suggestions.

manik3160:
@regulartim  I've just pushed the updated approach.Let me know if you need any change !

greedybear/settings.py
```diff
@@ -37,7 +37,7 @@
     print("WARNING!!! You need an ElasticSearch TPOT instance to have the Greedybear to work correctly.")
     if not DEBUG:
         print("you are in production mode: closing the application")
-        exit(9)
+        pass
```
regulartim: What is this?

tests/test_extraction_utils.py
```diff
@@ -534,7 +530,12 @@ def test_heralding_counts_login_attempts(self):
         """Heralding hits with credentials are counted as login attempts."""
         hits = [
             self._create_hit(src_ip="8.8.8.8", hit_type="Heralding", username="root", password="1234"),
-            self._create_hit(src_ip="8.8.8.8", hit_type="Heralding", username="admin", password="admin"),
+            self._create_hit(
+                src_ip="8.8.8.8",
+                hit_type="Heralding",
+                username="admin",
+                password="admin",
+            ),
```
regulartim: Please don't push unrelated formatting-only changes.

greedybear/settings.py
regulartim: What is this?

manik3160: Oh apologies! The pass in settings.py was an accidental leftover from when I was trying to bypass the Elasticsearch check so I could run the test suite locally.

tests/test_extraction_utils.py
manik3160: The wrapped lines in test_extraction_utils.py were just an artifact of my local auto-formatter catching them.
I didn't mean to include either of those in this PR.
Next time I will make sure not to include.

manik3160:
Hey @regulartim,I have just pushed a commit reverting them both to keep the PR focused and clean!

PR #1033, opened 2026-03-12
fix: normalize_credential_field missing truncation.

### LINKED ISSUE #1029: normalize_credential_field missing truncation causes extraction batch failure on oversized credentials
While reading through the Cowrie extraction pipeline I noticed that normalize_credential_field doesn't truncate to match the Credential model's max_length=256:                        
```
  # cowrie.py:66
  def normalize_credential_field(field: str) -> str:
      return field.replace("\x00", "[NUL]")
```
  Compare with normalize_command right above it, which correctly truncates to match its model's max_length:
```
  # cowrie.py:52-53
  def normalize_command(message: str) -> str:
      return message.removeprefix("CMD: ").replace("\x00", "[NUL]")[:1024]
```
  The Credential model enforces max_length=256 at the database level:
```
  # models.py:114-115
  username = models.CharField(max_length=256, blank=False)
  password = models.CharField(max_length=256, blank=False)
```
  And add_credential passes the value straight to PostgreSQL without any truncation:
```
  # cowrie_session.py:137
  credential, _ = Credential.objects.get_or_create(username=username, password=password)
```
  When an attacker sends a credential longer than 256 characters (common in buffer overflow / fuzzing attacks against SSH), this triggers a DataError from PostgreSQL. The exception doesn't just skip that one credential  it propagates up through _process_session_hit → _get_sessions → _get_scanners → extract_from_hits, where the pipeline catches it and discards the entire honeypot batch:
```
  # pipeline.py:98-99
  except Exception as exc:
      self.log.error(f"Extraction failed for honeypot {honeypot}: {exc}")
```
  All IOCs from that batch are lost  not just the session with the oversized credential.

  This gap was introduced by the Credential model refactoring  the old ArrayField was lenient on element length at the DB level, but CharField(max_length=256) on the new model strictly enforces it in PostgreSQL. The data migration in b0d85d1 fixed existing data, but the extraction code path for new incoming data was never patched.

  Suggested fix: Truncate credential fields to 256 chars in normalize_credential_field, consistent with how normalize_command already handles this for commands.

Sanchit2662:
# Description
I found that normalize_credential_field was missing truncation to 256 characters, while the Credential model fields are capped at 256. This caused a DataError   when Cowrie logs contained oversized credentials during extraction, failing entire batches.                                                                    
The fix adds [:256] truncation matching the pattern already used in normalize_command. Added a test verifying the truncation works correctly. This prevents the   DataError and ensures extraction continues even with malformed/oversized credentials.   

### Type of change

# Checklist

### Formalities

### Docs and tests

### DIFF
--- a/greedybear/cronjobs/extraction/strategies/cowrie.py
+++ b/greedybear/cronjobs/extraction/strategies/cowrie.py
@@ -62,14 +62,9 @@ def normalize_credential_field(field: str) -> str:
 
     Returns:
         Normalized credential field, truncated to 256 characters.
-        Any partial [NUL] replacement at the boundary is removed.
     """
-    # Truncate to 256 chars to match Credential model field max_length.
-    # Strip trailing partial [NUL] that may result from truncation mid-replacement.
-    result = field.replace("\x00", "[NUL]")[:256]
-    if result.endswith(("[N", "[NU", "[NUL")):
-        result = result[: result.rfind("[")]
-    return result
+    # Truncate to 256 chars to match Credential model field max_length
+    return field.replace("\x00", "[NUL]")[:256]
 
 
 class CowrieExtractionStrategy(BaseExtractionStrategy):
--- a/tests/test_cowrie_extraction.py
+++ b/tests/test_cowrie_extraction.py
@@ -81,13 +81,11 @@ class TestHelperFunctions(ExtractionTestCase):
         self.assertEqual(len(result), 256)
         self.assertTrue(result.startswith("A"))
 
-    def test_normalize_credential_field_partial_nul_at_boundary(self):
-        """Test that partial [NUL] at the 256-char boundary is removed."""
-        # Place a null byte so that [NUL] replacement straddles position 256
-        field = "A" * 254 + "\x00" + "B" * 100
-        result = normalize_credential_field(field)
-        self.assertFalse(result.endswith(("[N", "[NU", "[NUL")))
-        self.assertTrue(len(result) <= 256)
+    def test_normalize_credential_field_short_not_truncated(self):
+        """Test that short strings are not truncated."""
+        short_field = "password123"
+        result = normalize_credential_field(short_field)
+        self.assertEqual(result, short_field)
 
 
 class TestCowrieExtractionStrategy(ExtractionTestCase):


Sanchit2662:
Hi @regulartim , please review the changes when you get a chance. Happy to make any adjustments if needed. Thanks!

greedybear/cronjobs/extraction/strategies/cowrie.py
```diff
@@ -63,7 +63,8 @@ def normalize_credential_field(field: str) -> str:
     Returns:
         Normalized credential field
```
armoredvortex: The docstring should be updated to mention truncation

greedybear/cronjobs/extraction/strategies/cowrie.py
```diff
@@ -63,7 +63,8 @@ def normalize_credential_field(field: str) -> str:
...
     """
-    return field.replace("\x00", "[NUL]")
+    # Truncate to 256 chars to match Credential model field max_length
+    return field.replace("\x00", "[NUL]")[:256]
```
armoredvortex: This can result in the field ending with a partial "[NUL]" string if it gets cut off.
consider removing it after truncation

Sanchit2662:
Hey @regulartim , made the required changes based on the review by @armoredvortex :
                                                                                                                                                                                                           - Updated the docstring to mention truncation   
                                                                                                                                                                                                                                                                                                                                                                   - Added handling to strip any partial [NUL] that may appear at the 256-char boundary (e.g. if a null byte falls near position 252-255, the replacement [NUL] would get cut mid-string)

Let me know if anything else needs to be adjusted.

greedybear/cronjobs/extraction/strategies/cowrie.py
```diff
@@ -55,15 +55,21 @@ def normalize_command(message: str) -> str:
...
-        Normalized credential field
+        Normalized credential field, truncated to 256 characters.
+        Any partial [NUL] replacement at the boundary is removed.
     """
-    return field.replace("\x00", "[NUL]")
+    # Truncate to 256 chars to match Credential model field max_length.
+    # Strip trailing partial [NUL] that may result from truncation mid-replacement.
+    result = field.replace("\x00", "[NUL]")[:256]
+    if result.endswith(("[N", "[NU", "[NUL")):
+        result = result[: result.rfind("[")]
```
regulartim: We can safely ignore that case. I have yet only seen NULL characters in very short passwords. So I would prefer if you would just return `field.replace("\x00", "[NUL]")[:256]` for the reduced complexity.

Sanchit2662:
Done @regulartim ,simplified it as suggested and also added a check to ensure short strings aren’t truncated.

PR #947, opened 2026-03-04
feat(pipeline/api): replace IOC.asn with AutonomousSystem FK. Closes #770

drona-gyawali:
# Description

Changes include:
- Created `AutonomousSystem` model with `asn` (primary key) and optional `name`.
- Migrated existing `IOC.asn` data to `IOC.autonomous_system` FK.
- Updated extraction utilities, scoring, and feeds to use `autonomous_system`.
- Updated admin display to show ASN and AS name neatly.
- Created `ASNRepository` for consistent get-or-create logic.
- Added comprehensive tests covering:
    - ASNRepository edge cases (create, update, logging)
    - iocs_from_hits enrichment with AS info
    - Migration from ASN field to AutonomousSystem FK
    - Feeds API and scoring utilities integration

### Related issues
closes: #770 

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue).
- [x] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Chore (refactoring, dependency updates, CI/CD changes, code cleanup, docs-only changes).

# Checklist

Please complete this checklist carefully. It helps guide your contribution and lets maintainers verify that all requirements are met.

### Formalities

- [x] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/GreedyBear/Contribute/) to this project.
- [x] I chose an appropriate title for the pull request in the form: `<feature name>. Closes #999`
- [x] My branch is based on `develop`.
- [x] The pull request is for the branch `develop`.
- [x] I have reviewed and verified any LLM-generated code included in this PR.

### Docs and tests

- [x] I documented my code changes with docstrings and/or comments.
- [x] I have checked if my changes affect user-facing behavior that is described in the [docs](https://intelowlproject.github.io/docs/GreedyBear/Introduction/). If so, I also created a pull request in the [docs repository](https://github.com/intelowlproject/docs).
- [x] Linter (`Ruff`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
- [x] I have added tests for the feature/bug I solved.
- [x] All the tests gave 0 errors.

### GUI changes

Ignore this section if you did not make any changes to the GUI.

- [ ] I have provided a screenshot of the result in the PR.
[...]

PR #1062, opened 2026-03-16
Cache ASN aggregation results with version-based invalidation. Closes #852

### LINKED ISSUE #852: Optimization: Pre-calculate complex API aggregations using PostgreSQL Materialized Views to reduce DB compute load
**Prerequisite:** This issue depends on **#770** (and #854), which introduces the base `AutonomousSystem` model.

Right now, whenever someone hits the `/api/feeds/asn/` endpoint, we dynamically compute a bunch of heavy aggregates (`Sum`, `Count`, `Min`, `Max`) via `asn_aggregated_queryset()` in `api/views/utils.py`. 

This is totally fine with a small database, but as GreedyBear collects millions of IOC logs and honeypot events over time, doing aggregate `GROUP BY`s across all rows synchronously on every API request is going to cause some severe database strain and memory spikes.

I am not really sure with approach:
1. Setup an unmanaged Django Model and `RunSQL` migration to define a `MATERIALIZED VIEW` that holds all the calculated ASN groupings and counts.
2. During `greedybear/cronjobs/extraction/pipeline.py`, after new logs are enriched and scored, trigger a background task to `REFRESH MATERIALIZED VIEW CONCURRENTLY view_name`.
3. Swap out the expensive `annotate()` logic in `feeds.py` to simply query the materialized view.

opbot-xd:
# Description

Cache the heavy ASN aggregation query in [asn_aggregated_queryset()](file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/api/views/utils.py#448-548) using Django's cache framework, so identical `/api/feeds/asn/` requests between extraction runs are served from cache instead of re-querying the database.

**How it works:**
- Cache key = version counter + hash of all query params, so different `max_age`/[feed_type](file:///home/opbotxd/Desktop/vscode/dev/gsoc/honeynet/GreedyBear/api/views/utils.py#141-151)/`attack_type` combinations each get their own cache entry
- After each extraction run, the pipeline bumps the version counter, which effectively invalidates all stale ASN cache entries
- 24-hour TTL as safety net for cache expiry
- This complements the existing nginx-level caching by giving application-level control over invalidation (e.g. flushing after extraction)

### Related issues

Closes #852
Depends on #770 and #947 (already merged)

### Type of change

### Formalities

### Docs and tests

### GUI changes

_No GUI changes._

### DIFF
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -1,14 +1,16 @@
 # This file is a part of GreedyBear https://github.com/honeynet/GreedyBear
 # See the file 'LICENSE' for copying permission.
 import csv
+import hashlib
 import logging
+import urllib.parse
 from datetime import datetime, timedelta
 
 import feedparser
 import requests
 from django.conf import settings
 from django.contrib.postgres.aggregates import ArrayAgg
-from django.core.cache import cache
+from django.core.cache import cache, caches
 from django.db.models import Count, F, Max, Min, Q, Sum, Value
 from django.db.models.functions import JSONObject
 from django.http import HttpResponse, HttpResponseBadRequest, StreamingHttpResponse
@@ -446,15 +448,33 @@ def feeds_response(request=None, iocs=None, feed_params=None, valid_feed_types=N
 
 def asn_aggregated_queryset(iocs_qs, request, feed_params):
     """
-    Perform DB-level aggregation grouped by ASN.
+    Retrieve ASN aggregation data. Caches the heavy aggregation query
+    since the data only updates during the extraction cronjob.
 
     Args
         iocs_qs (QuerySet): Filtered IOC queryset from get_queryset;
         request (Request): The API request object;
         feed_params (FeedRequestParams): Validated parameter object
 
-    Returns: A values-grouped queryset with annotated  metrics and honeypot arrays.
+    Returns: A list of dicts with aggregated metrics and honeypot arrays per ASN.
     """
+
+    # Build reliable cache key from query params
+    sorted_params = sorted(request.query_params.lists())
+    params_string = urllib.parse.urlencode(sorted_params, doseq=True)
+    param_hash = hashlib.sha256(params_string.encode("utf-8")).hexdigest()
+
+    # To prevent per-worker continuous RAM bloat, use the shared DB-backed cache
+    # instead of the default LocMemCache, since the JSON response size can be large.
+    # The extraction pipeline invalidates this cache by bumping the version counter.
+    shared_cache = caches["django-q"]
+    version = shared_cache.get("asn_feeds_version", 1)
+    cache_key = f"asn_feeds_v{version}_{param_hash}"
+
+    cached_result = shared_cache.get(cache_key)
+    if cached_result is not None:
+        return cached_result
+
     asn_filter = request.query_params.get("asn")
     if asn_filter:
         iocs_qs = iocs_qs.filter(autonomous_system__asn=asn_filter)
@@ -480,9 +500,11 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
             first_seen=Min("first_seen"),
             last_seen=Max("last_seen"),
         )
-        .order_by(ordering)
     )
+    numeric_agg = numeric_agg.order_by(ordering)
 
+    # Honeypot names still require a lightweight aggregation because
+    # they depend on the active flag which can change independently.
     honeypot_agg = (
         iocs_qs.exclude(autonomous_system__isnull=True)
         .filter(general_honeypot__active=True)
@@ -497,13 +519,15 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
 
     hp_lookup = {row["asn"]: row["honeypots"] or [] for row in honeypot_agg}
 
-    # merging numeric aggregate with honeypot names for each asn
     result = []
     for row in numeric_agg:
         asn = row["asn"]
         row_dict = dict(row)
         row_dict["honeypots"] = sorted(hp_lookup.get(asn, []))
         result.append(row_dict)
+
+    # Set cache with a 60-minute timeout (max extraction interval length) to prevent memory bloat
+    shared_cache.set(cache_key, result, timeout=3600)
 
     return result
 
--- a/greedybear/cronjobs/extraction/pipeline.py
+++ b/greedybear/cronjobs/extraction/pipeline.py
@@ -1,6 +1,8 @@
 import logging
 from collections import defaultdict
 
+from django.core.cache import caches
+
 from greedybear.cronjobs.extraction.strategies.factory import ExtractionStrategyFactory
 from greedybear.cronjobs.repositories import (
     ElasticRepository,
@@ -103,5 +105,16 @@ class ExtractionPipeline:
             if ioc_records:
                 UpdateScores().score_only(ioc_records)
             ioc_record_count += len(ioc_records)
+
+        # 5. Invalidate API caches only if any IOC records were processed
+        if ioc_record_count > 0:
+            # Use the shared DB-backed cache so the version bump is visible to
+            # gunicorn API workers (LocMemCache is per-process).
+            self.log.info("Invalidating feeds ASN cache")
+            shared_cache = caches["django-q"]
+            try:
+                shared_cache.incr("asn_feeds_version")
+            except ValueError:
+                shared_cache.set("asn_feeds_version", 2, timeout=None)
 
         return ioc_record_count
--- a/tests/api/views/test_feeds_asn_view.py
+++ b/tests/api/views/test_feeds_asn_view.py
@@ -1,3 +1,4 @@
+from django.core.cache import cache, caches
 from django.utils import timezone
 from rest_framework.test import APIClient
 
@@ -63,6 +64,8 @@ class FeedsASNViewTestCase(CustomTestCase):
 
     def setUp(self):
         super().setUp()
+        cache.clear()
+        caches["django-q"].clear()
         self.client = APIClient()
         self.client.force_authenticate(user=self.superuser)
         self.url = "/api/feeds/asn/"
@@ -221,3 +224,45 @@ class FeedsASNViewTestCase(CustomTestCase):
             # Restore original name
             self.as_low.name = original_name
             self.as_low.save(update_fields=["name"])
+
+    def test_asn_feed_caching_behavior(self):
+        """
+        Verify that identical requests to the ASN feed return cached results,
+        and that invalidating the 'asn_feeds_version' forces a re-computation.
+        """
+        # 1. First request computes and caches the result
+        response1 = self.client.get(self.url)
+        self.assertEqual(response1.status_code, 200)
+        results1 = self._get_results(response1)
+        high_item1 = next((item for item in results1 if str(item["asn"]) == self.high_asn), None)
+        self.assertIsNotNone(high_item1)
+
+        # 2. Modify DB (e.g. change an IOC's attack count)
+        original_attack_count = self.ioc_high1.attack_count
+        self.ioc_high1.attack_count += 100
+        self.ioc_high1.save(update_fields=["attack_count"])
+
+        # 3. Second request should hit the cache and return SAME old value
+        response2 = self.client.get(self.url)
+        results2 = self._get_results(response2)
+        high_item2 = next((item for item in results2 if str(item["asn"]) == self.high_asn), None)
+
+        self.assertEqual(high_item1["total_attack_count"], high_item2["total_attack_count"])
+
+        # 4. Invalidate the cache (simulate extraction cronjob behavior)
+        shared_cache = caches["django-q"]
+        try:
+            shared_cache.incr("asn_feeds_version")
+        except ValueError:
+            shared_cache.set("asn_feeds_version", 2, timeout=None)
+
+        # 5. Third request should re-compute and show UPDATED DB value
+        response3 = self.client.get(self.url)
+        results3 = self._get_results(response3)
+        high_item3 = next((item for item in results3 if str(item["asn"]) == self.high_asn), None)
+
+        self.assertEqual(high_item3["total_attack_count"], high_item1["total_attack_count"] + 100)
+
+        # Cleanup DB
+        self.ioc_high1.attack_count = original_attack_count
+        self.ioc_high1.save(update_fields=["attack_count"])


opbot-xd:
Hi @regulartim , the PR is ready for review.
Although I assumed the Extraction Cronjob would be the hardest part, I was actually a bit stuck on the utils.py file. Refactoring the dynamic aggregation conditions turned out to be the most challenging part of this implementation, even Copilot raised a lot of genuine flags over it...

Review Request: Could you please take a close look at the updated filtering logic in `api/views/utils.py`
Thanks!

regulartim:
Hey @opbot-xd ! I think I found a major conceptional issue:
The AS API takes a `max_age` parameter with a default of `3`. This was a deliberate choice, because most of the time we are interested in current activity and not in accumulated data going back months or years. 
The problem is this fact completely defeats the purpose of your changes. Because you are only pre-computing value for a very specific API request, these values may rarely be used. One solution might be to always pre-compute the values for the default path, but that also falls apart when users change a single query parameter. 

Sorry for not catching this problem earlier. What do you think about that?

opbot-xd:
Hi @regulartim, you're absolutely right. the pre-computed fields would only be correct for one specific parameter combination, which defeats the purpose since max_age scopes the aggregation.

I've reworked the approach: instead of storing aggregates on the model, I now cache the actual aggregation query result using Django's cache framework. Each unique set of query params gets its own cache entry (keyed by a hash of the params), and the cache is invalidated at the end of every extraction run via a version bump.

 - Any `max_age`/ `feed_type`/ `attack_type` combination works correctly.
 - Identical requests between extraction runs are served from cache (O(1))
 
 I agree we are still running the heavy computation query for a new request but it would still provide us some performance edge.
 
 
 P.S. The implementation is also much cleaner with this approach.

regulartim:
Hey @opbot-xd ! Thank you for the suggestion. We already have a very similar caching in place for the /api/feeds endpoint. But the caching is not done by django, but via ngnix (see `http.conf`):

```
    location ~^/api/feeds {
        ...
        uwsgi_cache feeds_cache;
        uwsgi_cache_key $scheme$host$uri$is_args$args;
        uwsgi_cache_valid 200 10m;
        add_header X-Cache-Status $upstream_cache_status;
    }
```

What do you think of this? Is it equivalent / better / worse than your caching approach?

opbot-xd:
I wasn't aware of the nginx-level caching! Looking at it: The nginx cache is arguably better for this use case because it operates at the HTTP layer (zero Python overhead on cache hits) and already keys on the full URL + query params, so different `max_age/feed_type` combinations naturally get separate cache entries.
Given Thatthe nginx cache already solves the core performance issue from #852. Would you agree, or is there still value in pursuing application-level optimization here?

regulartim:
>  nginx cache already solves the core performance issue from #852

Yeah, I think it does. 

One advantage of the application level caching is that we have more control over it. We could, for example, flush all caches after each extraction run. That's a level of control that we do not have with nginx. So I am generally interested in your caching solution. Can you point me to the branch? Or open a draft PR?

opbot-xd:
Hi @regulartim, I've rebased and squashed the branch into a single commit, the PR now only contains the application-level caching for the ASN endpoint. (Updated desc and Title also)

**How it works:**
 - asn_aggregated_queryset() caches the aggregation result per unique set of query params (cache key = version + hash of all params)
 - After each extraction run, the pipeline bumps a version counter, which invalidates all stale cache entries
 - 24-hour TTL as a safety net
 - Different `max_age/feed_type/attack_type` combinations each get their own cache entry, so correctness is preserved regardless of query parameters

regulartim:
Hey @opbot-xd ! I took a quick look and this is definitely the way forward. Hope to find time for a full review soon.

api/views/utils.py
```diff
@@ -497,14 +529,27 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
...
+            "ioc_count": row["ioc_count"],
+            "total_attack_count": row["total_attack_count"],
+            "total_interaction_count": row["total_interaction_count"],
+            "total_login_attempts": row["total_login_attempts"],
+            "expected_ioc_count": row["expected_ioc_count"],
+            "expected_interactions": row["expected_interactions"],
+            "first_seen": row["first_seen"],
+            "last_seen": row["last_seen"],
+            "honeypots": sorted(hp_lookup.get(asn, [])),
+        }
```
regulartim: Whats the advantage of the explicit dict construction here? As far as I can see, the logic hasn't really changed.

api/views/utils.py
```diff
@@ -480,9 +505,16 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
...
     )
+    # Map API-facing 'as_name' ordering to the underlying 'as_name' alias
+    if ordering in {"name", "-name"}:
+        prefix = "-" if ordering.startswith("-") else ""
+        ordering = f"{prefix}as_name"
+
+    numeric_agg = numeric_agg.order_by(ordering)
 
+    # Honeypot names still require a lightweight aggregation because
+    # they depend on the active flag which can change independently.
```
regulartim, line 507: I think we can ignore that. Honeypot activation/deactivation does not happen too often. I think it is fine if the feed contains deactivated honeypots until the next extraction run.

greedybear/cronjobs/repositories/autonomous_system.py
```diff
@@ -6,10 +6,12 @@
 class ASRepository:
     """Repository to handle AutonomousSystem objects with caching."""
 
-    def __init__(self):
+    def __init__(self, preload_cache: bool = True):
```
regulartim: This new argument is not used anywhere, right?

api/views/utils.py
```diff
@@ -446,23 +447,47 @@ def feeds_response(request=None, iocs=None, feed_params=None, valid_feed_types=N
...
+
+    # Build reliable cache key from query params
+    sorted_params = sorted(request.query_params.items())
+    params_string = "&".join(f"{k}={v}" for k, v in sorted_params)
+    param_hash = hashlib.sha256(params_string.encode("utf-8")).hexdigest()
+
+    # Read version from the shared DB-backed cache so that the extraction
+    # pipeline (running in a separate Django-Q worker process) can invalidate
+    # caches visible to the gunicorn API workers.
+    from django.core.cache import caches
```
regulartim: Why no top level import?

api/views/utils.py
```diff
@@ -480,9 +505,16 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
             first_seen=Min("first_seen"),
             last_seen=Max("last_seen"),
         )
-        .order_by(ordering)
     )
+    # Map API-facing 'as_name' ordering to the underlying 'as_name' alias
+    if ordering in {"name", "-name"}:
+        prefix = "-" if ordering.startswith("-") else ""
+        ordering = f"{prefix}as_name"
+
```
regulartim: I also don't understand why you check for "name" here. The ordering value should be "as_name", no?

api/views/utils.py
```diff
@@ -446,23 +447,47 @@ def feeds_response(request=None, iocs=None, feed_params=None, valid_feed_types=N
...
+        return cached_result
+
     asn_filter = request.query_params.get("asn")
     if asn_filter:
         iocs_qs = iocs_qs.filter(autonomous_system__asn=asn_filter)
 
     # default ordering is overridden here because of serializer default(-last-seen) behaviour
     ordering = feed_params.ordering
-    if not ordering or ordering.strip() in {"", "-last_seen", "last_seen"}:
+    ordering_param_present = "ordering" in request.query_params
```
regulartim: Why is ordering_param_present necessary here?

greedybear/cronjobs/extraction/pipeline.py
```diff
@@ -104,4 +104,16 @@ def execute(self) -> int:
...
+        # Use the shared DB-backed cache so the version bump is visible to
+        # gunicorn API workers (LocMemCache is per-process).
+        self.log.info("Invalidating feeds ASN cache")
+        from django.core.cache import caches
+
+        shared_cache = caches["django-q"]
+        try:
+            shared_cache.incr("asn_feeds_version")
+        except ValueError:
+            shared_cache.set("asn_feeds_version", 2)
```
regulartim: Why do we need the version-based invalidataion? Can't we just run`cache.clear()`?

api/views/utils.py
opbot-xd: > This function comment says honeypot names are still aggregated live because they depend on the active flag, but the cached payload includes the merged honeypot list. If a GeneralHoneypot is toggled active/inactive, callers will keep seeing stale honeypot names until the cache TTL/version bump. Consider caching only the numeric aggregates (or adding honeypot-active state into the cache key / invalidation strategy) so honeypots truly stay live.

Exactly this is why I ignored this suggestion of copilot

greedybear/cronjobs/repositories/autonomous_system.py
opbot-xd: You're right, this was a leftover from the previous iteration. I'll remove it.

api/views/utils.py
opbot-xd: Nice catch! Moved all imports to module level. The local imports were a leftover from an earlier iteration where I was trying to avoid circular imports, but that's not an issue here.

api/views/utils.py
opbot-xd: I originally used the explicit dict to have a clear, documented contract of exactly which fields the cached response contains, especially since the result gets serialized into the cache. But you're right, `dict(row)`
 is cleaner here since the .values().annotate() already defines the exact fields. Reverted!

greedybear/cronjobs/extraction/pipeline.py
opbot-xd: I beleive `cache.clear()` only clears the current process's LocMemCache. Since the extraction pipeline runs in the Django Q worker (a separate process from gunicorn), calling cache.clear() there won't clear the API workers' caches. The version counter uses the shared DatabaseCache (django-q alias), which IS visible across processes, bumping it makes all workers naturally miss on the next request.

Do you have any cleaner approach for it?

api/views/utils.py
opbot-xd: You're right, was overthinking it. Reverted to the original logic, the serializer default already handles this correctly.

api/views/utils.py
opbot-xd: Ah, you're right. I had `name` there because I originally mapped the AutonomousSystem model field name, but the API response returns it as `as_name`. Removed the mapping so we just use `as_name`

opbot-xd:
> My plan is to replace the nginx caching with this logic. So for me this is kind of a prototypical implementation. Do you see a problem doing that? It would be possible to adapt this to the other feeds endpoits, right?

Hi @regulartim That’s a great idea! I don't see any problems doing that. It would be very straightforward to adapt this to the other `/api/feeds/` and `/api/feeds/advanced/` endpoints since they also take configurable query parameters.

Should I expand this logic to cover the other endpoints in this current PR, or should we merge this ASN endpoint as a prototype and do the rest in a new follow-up PR?

On a completely separate note, looking at `api/views/utils.py`, it has grown to about 600 lines and needs refactoring in my opinion... should we open issue for it? maybe we can pull out some code and put it in separate modules?

api/views/utils.py
regulartim: Thanks! My main concern here is this: if we add something to the row we also have to explicitly add it here. With `dict(row)` we don't have to.

greedybear/cronjobs/extraction/pipeline.py
regulartim: No, then you found a good solution for this, i guess! :+1:

regulartim:
> Should I expand this logic to cover the other endpoints in this current PR, or should we merge this ASN endpoint as a prototype and do the rest in a new follow-up PR?

I would keep the PR focused on AS and see how it works and if there are any rough edges. Then we can add it to the other endpoints in a follow up.

> On a completely separate note, looking at api/views/utils.py, it has grown to about 600 lines and needs refactoring in my opinion... should we open issue for it? maybe we can pull out some code and put it in separate modules?

Yes, I have the idea of refactoring the FeedsAPI in an object oriented way. Will open an issue soon. In this refactoring we can then also add the new caching to the other endpoints, I think.

api/views/utils.py
```diff
@@ -497,14 +518,16 @@ def asn_aggregated_queryset(iocs_qs, request, feed_params):
...
 
-    # merging numeric aggregate with honeypot names for each asn
     result = []
     for row in numeric_agg:
         asn = row["asn"]
         row_dict = dict(row)
         row_dict["honeypots"] = sorted(hp_lookup.get(asn, []))
         result.append(row_dict)
 
+    # Set cache with a 24-hour timeout (will be invalidated explicitly by extraction pipeline if needed)
```
regulartim: I think the 24h timeout is to large. If I understand the code correctly, cache entries are never explicitly deleted. So they remain in memory until the timeout. So I would suggest 60 minutes here as this is our maximum extraction interval. Just to remove pressure from the memory. Do you agree?

opbot-xd:
Hi @regulartim , You made a fantastic point about reducing RAM usage. Storing the heavy payload in the shared `django-q` cache directly prevents keeping N isolated copies of the payload loaded in each gunicorn worker's memory.

However, we **still need the version counter** even if we use the DB cache directly. Here's why:

- The endpoint accepts parameters like `max_age`, `attack_type`, and `ordering`. Because of this, extraction creates dozens of unique cache keys for the ASN endpoint, not just one.
- Django's native cache framework does not support wildcard deletion (e.g. `cache.delete("asn_feeds_*")`). Calling `cache.clear()` would wipe the *entire* database cache, potentially destroying legitimate background tasks or session data.
- Bumping the version counter instantly orphans all those dynamic keys in one single DB operation. Old keys simply TTL out, preventing any cross-deletion of unrelated cached data.

I agree we should update the actual payload caching to use `caches["django-q"]` to fix the RAM bloat. Should I push an update to switch the cache backend entirely over to `django-q`?

opbot-xd:
I misclicked on close with comment.

regulartim:
> However, we still need the version counter even if we use the DB cache directly

I see, thanks for clarifying. :+1: 

> I agree we should update the actual payload caching to use caches["django-q"] to fix the RAM bloat. Should I push an update to switch the cache backend entirely over to django-q?

Yes, I think that's the right move. Thanks for all the work and thought you put into this!

PR #1100, opened 2026-03-22
fix: reject promise on error in requestPasswordReset

### LINKED ISSUE #1095: Bug : requestPasswordReset swallows errors, causing premature popover closing
## Description:
In f`rontend/src/components/auth/api.js`, the `requestPasswordReset` function currently catches API errors and returns null instead of returning a rejected promise.

Because the promise resolves successfully with null, the underlying UI component `(EmailForm inside ForgotPasswordButton)` treats the request as a success. It executes await `apiCallback(values)` without throwing an error, which causes it to immediately trigger `onFormSubmit()` and close the popover, even though the email failed to send.

---

### File Location:
`frontend/src/components/auth/api.js`

---

### Steps to Reproduce:

- Open the application and click the `"Forgot Password?"` button to open the email input popover.
- Open Browser Developer `Tools -> Network tab`.
- Set network throttling to `"Offline"` (to force an API failure).
- Enter an email address and click `Submit.`
- Observe: A red `"Failed to send email!"` toast appears, but the popover automatically closes, preventing the user from correcting their input or trying again.

---

### Expected Behavior:
When the API request fails, the promise should be rejected. The UI should display the error toast, but the `EmailForm` popover should stay open and the loading spinner should stop, allowing the user to try again.
 
---

### Actual Behavior:
The promise resolves with `null,` causing the UI to execute its `"success"` logic (closing the popover) despite the underlying failure.

Fikri-20:
Fixes #1095

`requestPasswordReset` was returning `null` on error while every other function in `api.js` returns `Promise.reject(err)`. This caused the UI to treat failures as success, closing the popover before the user could retry.

Changed `return null` to `return Promise.reject(err)` to match the existing pattern.

### DIFF
--- a/frontend/src/components/auth/api.js
+++ b/frontend/src/components/auth/api.js
@@ -51,7 +51,7 @@ export async function requestPasswordReset(body) {
     return resp;
   } catch (err) {
     addToast("Failed to send email!", err.parsedMsg, "danger", true);
-    return null;
+    return Promise.reject(err);
   }
 }
 


regulartim:
Hey @Fikri-20 ! Lots of issues with your PR:
1. you are not assigned to the corresponding issue
2. you obviously did not read our contribution guidelines (or you ignored them)
3. you deleted the PR template
4. it targets the wrong branch

I suggest you read and follow the [contribution guidelines](https://intelowlproject.github.io/docs/GreedyBear/Contribute/) if you want to contribute to this project.

PR #1106, opened 2026-03-23
Non-root user using gosu in app container. closes #1102

### LINKED ISSUE #1102: Run 'app' service as non-root user
### Current behavior
The `app` service runs as root in `default.yml`. The `qcluster` service already uses `user: "2000:82"`, making `app` inconsistent.

### What needs to change
- Install `gosu` in the Dockerfile
- Use `gosu` in `entrypoint_gunicorn.sh` to drop privileges from root to `www-data` before gunicorn starts

The Dockerfile already sets up `www-data` (uid 2000) and chowns all necessary directories — so the infrastructure is already in place.

### Why
- Security best practice — if an attacker exploits a vulnerability in Django or Gunicorn, running as root means they get root on the host
- Consistent with `qcluster` which already uses `user: "2000:82"`
- Recommended by the [official Gunicorn Docker docs](https://gunicorn.org/guides/docker/#non-root-user)

### Risk for existing instances
The `chown` line in `entrypoint_gunicorn.sh` currently fixes log file ownership on every startup. With `gosu`, the entrypoint still starts as root so `chown` still runs — existing deployments with log files owned by root from previous versions are automatically fixed on restart. No manual migration needed.

I would like to work on this issue!

SupRaKoshti:
# Description

- Install `gosu` in Dockerfile alongside existing runtime dependencies (`libgomp1`, `curl`)
- Add `mkdir -p /run/gunicorn` and `chown -R 2000:82 /var/log/greedybear /run` in `entrypoint_gunicorn.sh` to pre-create the gunicorn socket directory and fix ownership of any root-owned files from previous deployments
- Replace `exec "$@"` with `exec gosu www-data "$@"` at the end of `entrypoint_gunicorn.sh` so gunicorn runs as `www-data` (uid 2000) instead of root

The entrypoint still starts as root so `chown`, migrations and `collectstatic` continue to work correctly. `gosu` drops privileges to `www-data` right before gunicorn starts, ensuring gunicorn and all its workers run as non-root for their entire lifetime.

Existing deployments with log files owned by root are automatically fixed by the `chown` step on every restart — no manual migration needed.

Tested with both `default.yml` (production) and `local_override.yml` (development) — all working correctly.

### Related issues
Closes #1102 

### Type of change

# Checklist

### Formalities

### Docs and tests
- [ ] I have checked if my changes affect user-facing behavior that is described in the docs. If so, I also created a pull request in the docs repository.

### DIFF
--- a/docker/entrypoint_gunicorn.sh
+++ b/docker/entrypoint_gunicorn.sh
@@ -21,7 +21,7 @@ mkdir -p /var/log/greedybear/gunicorn
 mkdir -p /run/gunicorn
 
 # Fix log file ownership (manage.py commands above run as root and may create new log files)
-chown -R 2000:82 /var/log/greedybear /run
+chown -R 2000:82 /var/log/greedybear /run/gunicorn
 
 # Obtain the current GreedyBear version number
 . /opt/deploy/greedybear/docker/.version
@@ -33,4 +33,10 @@ echo "DEBUG: $DEBUG"
 echo "DJANGO_TEST_SERVER: $DJANGO_TEST_SERVER"
 echo "------------------------------"
 
-exec gosu www-data "$@"
+if [ "$DJANGO_TEST_SERVER" = "True" ]; then
+    # Dev mode: run as root (needed for hot-reload on volume-mounted source)
+    exec "$@"
+else
+    # Production mode: drop privileges to www-data before starting Gunicorn
+    exec gosu www-data "$@"
+fi


SupRaKoshti:
Hi @regulartim , I have implemented non-root user! please take look when you get a chance. happy to make any changes based on your review!

docker/entrypoint_gunicorn.sh
```diff
@@ -18,9 +18,10 @@ python manage.py collectstatic --noinput --clear --verbosity 0
 
 # Ensure log directories exist (volumes may persist from older builds)
 mkdir -p /var/log/greedybear/gunicorn
+mkdir -p /run/gunicorn
 
 # Fix log file ownership (manage.py commands above run as root and may create new log files)
-chown -R 2000:82 /var/log/greedybear
+chown -R 2000:82 /var/log/greedybear /run
```
regulartim: Why not "/run/guincorn" ? There isn't really much else in "/run" but that might change...

docker/entrypoint_gunicorn.sh
```diff
@@ -32,4 +33,4 @@ echo "DEBUG: $DEBUG"
 echo "DJANGO_TEST_SERVER: $DJANGO_TEST_SERVER"
 echo "------------------------------"
 
-exec "$@"
+exec gosu www-data "$@"
```
regulartim: I had some problems when running the django dev server: watchfiles does not have the permission to read some of my folders. Maybe we should skip gosu in dev mode?
```
  if [ "$DJANGO_TEST_SERVER" = "True" ]; then
      exec "$@"
  else
      exec gosu www-data "$@"
  fi
```

mlodic: yeah definitively not needed in dev mode

docker/entrypoint_gunicorn.sh
SupRaKoshti: Good catch! you are 100% correct. In `default.yml` file, docker volume(gunicorn_sockets) is mounted at `/run/gunicorn` specifically. so `www-data` needs ownership only `run/gunicorn` not whole 'run' folder.

docker/entrypoint_gunicorn.sh
SupRaKoshti: Yes, I will write the code to skip gosu in dev mode!

SupRaKoshti:
Hi @regulartim , I have made changes! Could you please verify?

PR #1141, opened 2026-03-27
Fix/cowrie session str nonetype

### LINKED ISSUE #1083: fix: handle None session_id in CowrieSession.__str__
## Description
While setting up GreedyBear locally, I found a bug where visiting `/admin/greedybear/ioc/add/` throws a `TypeError: 'NoneType' object cannot be interpreted as an integer`.

## Bug
When Django admin renders the IOC add form, it calls `__str__` on related `CowrieSession` objects. Since `session_id` is the primary key and is `None` on unsaved objects, calling `hex(self.session_id)` raised:

```
TypeError: 'NoneType' object cannot be interpreted as an integer
```

## Fix
Added a `None` check for `session_id` in `CowrieSession.__str__` before calling `hex()`:

```python
# Before
def __str__(self):
    return f"Session {hex(self.session_id)[2:]} from {self.source.name}"

# After
def __str__(self):
    if self.session_id is None:
        return "New Session (unsaved)"
    return f"Session {hex(self.session_id)[2:]} from {self.source.name}"
```

## Steps to Reproduce
1. Run GreedyBear locally
2. Go to `/admin/greedybear/ioc/add/`
3. You will see `TypeError: 'NoneType' object cannot be interpreted as an integer`

I already have the fix ready in my local repository. Could you please guide me on whether I should open a new issue first and then submit a PR, or directly submit the PR?

ayushagarwalhere:
## Description
Closes #1083

Fixes a `TypeError` that occurred when visiting `/admin/greedybear/ioc/add/` in the Django admin panel.

## Bug
When Django admin renders the IOC add form, it calls `__str__` on related `CowrieSession` objects. Since `session_id` is the primary key and is `None` on unsaved objects, calling `hex(self.session_id)` raised:

```
TypeError: 'NoneType' object cannot be interpreted as an integer
Exception Location: /opt/deploy/greedybear/greedybear/models.py, line 161, in __str__
```

## Root Cause
`CowrieSession` has a `ForeignKey` to `IOC`. When Django renders the IOC add form, it calls `__str__()` on related `CowrieSession` objects. Since `session_id` is `None` on unsaved objects, `hex(self.session_id)` crashes.

## Fix
Added a `None` check for `session_id` in `CowrieSession.__str__()` before calling `hex()`:

```python
# Before
def __str__(self):
    return f"Session {hex(self.session_id)[2:]} from {self.source.name}"

# After
def __str__(self):
    if self.session_id is None:
        return "New Session (unsaved)"
    return f"Session {hex(self.session_id)[2:]} from {self.source.name}"
```

## Steps to Reproduce
1. Run GreedyBear locally
2. Go to `/api/admin/greedybear/ioc/add/`
3. You will see `TypeError: 'NoneType' object cannot be interpreted as an integer`

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/greedybear/models.py
+++ b/greedybear/models.py
@@ -169,10 +169,11 @@ class CowrieSession(models.Model):
         indexes = [
             models.Index(fields=["source"]),
         ]
-        def __str__(self):
-            if self.session_id is None:
-                return "New Session (unsaved)"
-            return f"Session {hex(self.session_id)[2:]} from {self.source.name}"
+
+    def __str__(self):
+        if self.session_id is None:
+            return "New Session (unsaved)"
+        return f"Session {hex(self.session_id)[2:]} from {self.source.name}"
 
 
 class CowrieFileTransfer(models.Model):


ayushagarwalhere:
I would like to apologize for creating a little mess on implementing the fix, i got confused with the indentation and as it was my first PR to any org so was not aware of these copilot reviews and it gave me the impression that something was not correct !!

PR #1144, opened 2026-03-28
Redundant API fetching fix. Closes #1135

### LINKED ISSUE #1135: [Frontend] Redundant API Fetching (Data Overlap)
Caught something while looking at the dashboard components - we're hitting `IOC_ATTACKER_COUNTRIES_URI` twice on every refresh. 

Right now, `AttackOriginMap.jsx` and `AttackOriginCountriesChart` (in `charts.jsx`) are both listening to the `useTimePickerStore` range. When that range updates, they both fire off their own axios request to the exact same endpoint with the exact same params. 

Found it here:
- `AttackOriginMap.jsx` (lines 78-80) triggers a request for the map data.
- `charts.jsx` (lines 149-151) triggers an identical request for the top countries list.

It’s not breaking anything visually yet, but we're basically doubling the backend load for that specific stat for no reason. Plus, if one request fails while the other works, we'll end up with a map and a list that don't match up.

Instead of having them fetch independently, we should probably pull the fetching logic out into a shared hook or just have a single provider that fetches the data once and passes it down. This kind of redundancy is a good case for why the dashboard needs a more modular data-fetching setup.

chauhan-varun:
# Description

This PR resolves the redundant API fetching issue in the GreedyBear dashboard by centralizing data management for the `attacker-countries` endpoint into a shared Zustand store. Previously, both the World Map and the Top Countries chart would trigger identical, simultaneous requests for the same time range.

Key improvements:
- **Centralized Store**: Created `useAttackerCountriesStore` to manage all data fetching and normalization for country statistics.
- **De-duplication**: Both components now consume from the same store, resulting in exactly one API call per range change instead of two.
- **Caching**: Successfully fetched data is cached in the store, preventing redundant network requests when switching back to a previously loaded time range.
- **Race Condition Handling**: Implemented `AbortController` to automatically cancel stale in-flight requests when a user rapidly switches between time ranges.
- **Stability**: Added defensive data validation and robust error rendering to handle malformed API responses or network failures gracefully.

### Related issues

Closes #1135

### Type of change

# Checklist

### Formalities

- [x] I chose an appropriate title for the pull request in the form: `Redundant API fetching fix. Closes #1135`

### Docs and tests

### GUI changes

- [x] I have provided a screenshot of the result in the PR.

### DIFF
--- a/frontend/src/components/dashboard/AttackOriginMap.jsx
+++ b/frontend/src/components/dashboard/AttackOriginMap.jsx
@@ -106,7 +106,7 @@ export default function AttackOriginMap() {
         className="d-flex justify-content-center align-items-center text-muted"
         style={{ minHeight: 200 }}
       >
-        {typeof error === "string" ? error : error.message}
+        {error}
       </div>
     );
   }
--- a/frontend/src/components/dashboard/utils/charts.jsx
+++ b/frontend/src/components/dashboard/utils/charts.jsx
@@ -160,7 +160,7 @@ export const AttackOriginCountriesChart = React.memo(() => {
   if (error) {
     return (
       <div className="d-flex justify-content-center align-items-center py-4 text-muted">
-        {typeof error === "string" ? error : error.message}
+        {error}
       </div>
     );
   }


chauhan-varun:
Hey @regulartim, could you review this PR when you get a chance?

I also touched [register.test.jsx](https://github.com/intelowlproject/GreedyBear/pull/1144/changes#diff-f9be706669cc8120c063801e9d1387bbd1d96ef618bd88f43487bc5d40f2f911) to increase the timeout a bit, because that test was flaky in the full coverage run and was still failing CI even after the main fix

frontend/src/stores/useAttackerCountriesStore.jsx
```diff
@@ -0,0 +1,103 @@
+import axios from "axios";
+import { create } from "zustand";
+import { IOC_ATTACKER_COUNTRIES_URI } from "../constants/api";
+
+const NAME_FIXES = {
```
armoredvortex, line 6: The comment explaining why these fixes exist should be present here also so that the intent is not lost.

frontend/src/stores/useAttackerCountriesStore.jsx
```diff
@@ -0,0 +1,103 @@
...
+  loading: false,
+  error: null,
+  lastRange: null,
+  currentController: null,
+
+  fetchData: async (range) => {
+    const rangeStr = JSON.stringify(range);
+    if (get().lastRange === rangeStr && !get().error) return;
+
+    if (get().loading && get().lastRange === rangeStr) return;
```
armoredvortex: the 2nd condition seems redundant, if loading were to be `false` the line above would've already returned.

frontend/src/components/dashboard/AttackOriginMap.jsx
```diff
@@ -149,7 +106,9 @@ export default function AttackOriginMap() {
         className="d-flex justify-content-center align-items-center text-muted"
         style={{ minHeight: 200 }}
       >
-        {error}
+        {typeof error === "string"
```
armoredvortex: this is inside `if (error)` block so `error?` branch is dead code.

chauhan-varun:
Hey @armoredvortex Good catch, you were right. I removed the redundant error fallback and simplified the error rendering in both dashboard components

armoredvortex:
> Hey @armoredvortex Good catch, you were right. I removed the redundant error fallback and simplified the error rendering in both dashboard components

@chauhan-varun, please correct me if i'm wrong but, this still looks flawed.
1) error can only be a string inside that block (as per current implementation) so typeof error == "string" check is not required.
2) in the other branch which is never reached, you're using `error.message` which is non existent. where does that come from?

chauhan-varun:
Ahh yeah, you’re right. Fixed it now just rendering {error} directly, Thanks for the review @armoredvortex

frontend/tests/components/auth/Register.test.jsx
regulartim, line 1: > I also touched [register.test.jsx](https://github.com/intelowlproject/GreedyBear/pull/1144/changes#diff-f9be706669cc8120c063801e9d1387bbd1d96ef618bd88f43487bc5d40f2f911) to increase the timeout a bit, because that test was flaky in the full coverage run and was still failing CI even after the main fix

That's fine for now, but it does not really change the fact that the test is flaky, right? Should we maybe change the way we test here? If so, would you please open an issue for that?

regulartim:
@armoredvortex Also, thanks for the review! :)

frontend/tests/components/auth/Register.test.jsx
chauhan-varun: Yeah you're right, the timeout is just a band-aid. Opened #1148 for it
The main problems are leaked `localStorage` between tests and `userEvent.type` being too slow

PR #1197, opened 2026-04-07
Standardize country statistics using ISO codes. Closes #1181

### LINKED ISSUE #1181: Standardize country statistics using ISO codes
The current `/api/statistics/countries` endpoint only returns country names, which can be inconsistent. We should update the API to include the `attacker_country_code` (ISO-A2) and refactor the dashboard components to aggregate data using these codes. This will ensure consistent data across the map and charts even if country names in the database vary.

chauhan-varun:
# Description

This PR standardizes country names in the dashboard by adding ISO-A2 codes to the `/api/statistics/countries` endpoint. I've added a frontend mapping utility and updated the store to aggregate data correctly for both the map and charts. This ensures consistent naming even if database entries vary. 

Additionally, I fixed a typo in the statistics tests (`VzziewType` -> `ViewType`) and renamed some migration functions to satisfy `ruff` linting requirements.

### Related issues

Closes #1181

### Type of change

# Checklist

### Formalities

- [x] I chose an appropriate title for the pull request in the form: `Standardize country statistics using ISO codes. Closes #1181`

### Docs and tests

### GUI changes

- [x] I have provided a screenshot of the result in the PR.
- [x] I have created new frontend tests for the new component or updated existing ones.


chauhan-varun:
@armoredvortex would you like to review this pr?

frontend/src/utils/isoMapping.js
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Normalizes a country using its ISO-A2 code.
+ * If mapping exists, it returns the standard name for the map.
+ * Otherwise, it falls back to the provided name.
+ *
+ * @param {string} code - ISO-A2 code
+ * @param {string} fallback - Fallback country name
+ * @returns {string} - Resolved country name
+ */
+export function getStandardMapName(code, fallback) {
```
armoredvortex: should also handle lowercase country codes imo.

frontend/tests/utils/isoMapping.test.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(getStandardMapName("XX", "Unknown Country")).toBe("Unknown Country");
+  });
+
+  it("should return the fallback name if the ISO code is missing or null", () => {
+    expect(getStandardMapName(null, "United States")).toBe("United States");
+    expect(getStandardMapName(undefined, "China")).toBe("China");
+    expect(getStandardMapName("", "Italy")).toBe("Italy");
+  });
+
+  it("should handle mixed case ISO codes if possible (though backend should provide uppercase)", () => {
```
armoredvortex: case-insensitive seems like a better option.

frontend/src/utils/isoMapping.js
```diff
@@ -0,0 +1,195 @@
+/**
+ * A mapping from ISO 3166-1 alpha-2 country codes to the specific names
+ * used by Natural Earth / world-atlas@2 in `countries-110m.json`.
+ *
+ * This ensures that even if renaming occurs in the backend or GeoIP database,
+ * the frontend can still correctly map data to the geographical regions.
+ */
+export const ISO_TO_MAP_NAME = {
```
armoredvortex: Is this list exhaustive?
because I see that "Kosovo" exists in `countries-110m.json` but the mapping for code XK is not present

armoredvortex (changes requested):
@chauhan-varun I'm not much familiar with the backend, but from what i can see there are a lot of changes in the migrations files.
were these edits made automatically? because these look purely cosmetic changes to me, i dont think they should be part of this PR.

Also, frontend and backend tests are failing for me currently.

frontend/src/utils/isoMapping.js
chauhan-varun: Done, added Kosovo (XK)! I also added Antarctica, N. Cyprus, and Somaliland so it now matches the map's JSON exactly

frontend/tests/utils/isoMapping.test.js
chauhan-varun: good idea it's case-insensitive now. I added tests to make sure things like 'us' and 'US' both work

chauhan-varun:
Thanks @armoredvortex! I only touched the migrations to fix some ruff linting errors (N802), but I'm totally happy to revert those if you want to keep the PR focused

Also, I fixed a `VzziewType` typo in the tests, could the failures be related to that or the new country code field?

chauhan-varun:
image
image

fixed the tests too

chauhan-varun:
Hey @regulartim please take a look

tests/__init__.py
```diff
@@ -72,8 +73,8 @@ def setUpTestData(cls):
             destination_ports=[22, 23, 24],
             login_attempts=1,
             recurrence_probability=0.1,
-            expected_interactions=11.1,
```
regulartim, line 75: Why did you remove that line?

regulartim (changes requested):
Hey @chauhan-varun ! Thanks for your work! :) Two things that I stumbled upon:
- Never ever make changes to existing migrations. Please remove these changes from the PR.
- I am not sure about the ISO mapping. Maintaining that table is something I would like to avoid. Can you think of a different solution here?

chauhan-varun:
> * Never ever make changes to existing migrations. Please remove these changes from the PR.

Got it, I'll keep this in mind and revert the changes to the existing migration
> * I am not sure about the ISO mapping. Maintaining that table is something I would like to avoid. Can you think of a different solution here?

What if we just parse the map's TopoJSON directly instead? It already has the ISO codes and names in it, so we could extract the mapping from there at runtime. That way there's nothing to maintain manually and everything stays in sync

tests/__init__.py
chauhan-varun: My bad! I removed that line by mistake while adding the `attacker_country_code` field it's fixed now

regulartim:
> What if we just parse the map's TopoJSON directly instead? It already has the ISO codes and names in it, so we could extract the mapping from there at runtime. That way there's nothing to maintain manually and everything stays in sync

Sounds good! :+1:

chauhan-varun:
I looked into the TopoJSON it actually uses numeric ISO codes, not alpha-2, so there's still a small gap to bridge. I see two ways to handle it:
- Use `i18n-iso-countries` npm package to convert alpha-2 to numeric automatically (zero manual tables)
- Use the existing country.js we already have (~20 name fixes) and build the rest from the TopoJSON at runtime (no new dependency)

Which do you prefer?

regulartim:
Thanks for investigating! :) I think I prefer the `i18n-iso-countries` approach for the lower maintenance overhead.

chauhan-varun:
Done! Switched to `i18n-iso-countries` the store now groups by alpha-2 code instead of country name, and the map matches using geo.id (numeric ISO). Since we no longer match by name at all, both isoMapping.js and country.js became dead code so I removed them. All tests pass!

regulartim (approved):
Looks good, thanks! :)

PR #1211, opened 2026-04-09
fix: transition feeds_consume to allowlist for token revocation (#1190)

### LINKED ISSUE #1190: Bug: feeds_consume  accepts valid tokens that have no database record (revocation bypass)
Hey @regulartim, I was going through the share/consume/revoke token flow in the feeds API and I think I found something a bit concerning. wanted to write it up properly.

## what is going on

so when you call `/api/feeds/share`, it does two things: creates a signed token using django signing and writes the token hash to the [ShareToken](cci:2://file:///Users/rishabhshukla/Desktop/GreedyBear/greedybear/models.py:277:0-306:66) table. later when someone calls `/api/feeds/consume/<token>`, the endpoint is supposed to check if that token has been revoked before serving the feed.

the issue is the revocation check in [feeds_consume](cci:1://file:///Users/rishabhshukla/Desktop/GreedyBear/api/views/feeds.py:242:0-272:80) (line 260-262 in [api/views/feeds.py](cci:7://file:///Users/rishabhshukla/Desktop/GreedyBear/api/views/feeds.py:0:0-0:0)):

```python
token_hash = hashlib.sha256(token.encode()).hexdigest()
if ShareToken.objects.filter(token_hash=token_hash, revoked=True).exists():
    return Response({"error": "Token has been revoked"}, status=status.HTTP_400_BAD_REQUEST)
```

this is a blocklist style check. it only blocks a token if there is a [ShareToken](cci:2://file:///Users/rishabhshukla/Desktop/GreedyBear/greedybear/models.py:277:0-306:66) row in the database where `revoked=True`. but if there is no row at all for that token hash, this filter returns `False` and the code just continues and serves the feed data. no error, nothing.

## why this is a problem

there are a few ways a valid signed token can exist without a database record:

1. **db write failure during share** - [feeds_share](cci:1://file:///Users/rishabhshukla/Desktop/GreedyBear/api/views/feeds.py:199:0-239:65) uses `get_or_create` to save the token hash. if that write fails silently for any reason (network blip, transient db issue, etc.), the signed token is still returned to the client. that token now lives forever with zero way to revoke it because there is no record to set `revoked=True` on.

2. **the revoke endpoint becomes useless** - if you try to call `/api/feeds/revoke/<token>` for a token with no db record, you get a 403 (`Token not found. Only the creator can revoke a token.`). so you literally cannot revoke orphan tokens through any existing endpoint.

3. **if SECRET_KEY ever leaks** - an attacker who knows the secret key can generate tokens that will never have a db record. revocation as a security feature is then completely bypassed for those tokens.

the core design issue is that the current code uses a **blocklist** (allow everything, explicitly block revoked tokens) when it should use an **allowlist** (block everything by default, only allow tokens that actively exist in the db and are not revoked).

## how I confirmed this

to make sure I wasn't just misreading the code, I wrote a test script and ran it directly inside the running `greedybear_app` docker container against the actual postgres database. 

first, I generated a valid signed token the exact same way [feeds_share](cci:1://file:///Users/rishabhshukla/Desktop/GreedyBear/api/views/feeds.py:199:0-239:65) does (using `signing.dumps` with the `greedybear-feeds` salt), but I intentionally did NOT save it to the [ShareToken](cci:2://file:///Users/rishabhshukla/Desktop/GreedyBear/greedybear/models.py:277:0-306:66) table to simulate a db write failure. then, I ran that token through the exact `ShareToken.objects.filter(token_hash=..., revoked=True).exists()` check that [feeds_consume](cci:1://file:///Users/rishabhshukla/Desktop/GreedyBear/api/views/feeds.py:242:0-272:80) uses. the check completely missed it (returned False) because there was no row, which would allow the code to proceed and serve the feed. 

I also ran a control test where I manually inserted a revoked record into the db for a token, and the check blocked that one correctly. so the logic isn't completely broken, it just has a huge blind spot for tokens that never made it into the database.

## expected behavior

[feeds_consume](cci:1://file:///Users/rishabhshukla/Desktop/GreedyBear/api/views/feeds.py:242:0-272:80) should only serve the feed if a matching [ShareToken](cci:2://file:///Users/rishabhshukla/Desktop/GreedyBear/greedybear/models.py:277:0-306:66) record exists in the database AND is not revoked. if no record exists at all, the endpoint should return a 400 error, same as it does for revoked tokens.

happy to open a PR with the fix if this makes sense. Please let me know what you think!

R1sh0bh-1:
# Description

Hey @regulartim,

I fixed the token revocation bypass issue as described in #1190.

The old check was only blocking tokens that had `revoked=True` in the database. If no `ShareToken` record existed at all (for example due to a failed DB write or a manually crafted token), the endpoint would still serve the feed.

I changed the logic to a proper allowlist check using `get()`:

```python
try:
    share_token = ShareToken.objects.get(token_hash=token_hash)
except ShareToken.DoesNotExist:
    return Response(
        {"error": "Invalid or expired token"},
        status=status.HTTP_400_BAD_REQUEST,
    )

if share_token.revoked:
    return Response(
        {"error": "Token has been revoked"},
        status=status.HTTP_400_BAD_REQUEST,
    )
```

I tested it exactly the same way I reported in the issue,  I created a valid signed token (using `signing.dumps` with the `greedybear-feeds` salt) but did **not** save it to the `ShareToken` table. Now it correctly returns `400 "Invalid or expired token"`.  
Normal revoked tokens and valid tokens continue to work as expected.

### Related issues
Closes #1190

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes

Please let me know if any changes are required.

### DIFF
--- a/tests/api/views/test_feeds_advanced_view.py
+++ b/tests/api/views/test_feeds_advanced_view.py
@@ -1,9 +1,14 @@
+import hashlib
 from datetime import datetime, timedelta
+from unittest.mock import patch
 
 from django.conf import settings
+from django.core import signing
+from django.core.cache import cache
 from rest_framework.test import APIClient
 
-from greedybear.models import IOC, AutonomousSystem, IocType
+from api.throttles import SharedFeedRateThrottle
+from greedybear.models import IOC, AutonomousSystem, IocType, ShareToken
 from tests import CustomTestCase
 
 
@@ -246,8 +251,6 @@ class FeedsEnhancementsTestCase(CustomTestCase):
 
     def test_shareable_feed_expired_token(self):
         """Consuming a tampered/expired token returns 400."""
-        from django.core import signing
-
         data = {
             "feed_type": "all",
             "attack_type": "all",
@@ -276,10 +279,6 @@ class FeedsEnhancementsTestCase(CustomTestCase):
 
     def test_consume_valid_token_without_db_record_is_rejected(self):
         """A valid signed token that was never saved to the DB is rejected (allowlist check)."""
-        import hashlib
-
-        from django.core import signing
-
         data = {
             "feed_type": "all",
             "attack_type": "all",
@@ -302,8 +301,6 @@ class FeedsEnhancementsTestCase(CustomTestCase):
         token = signing.dumps(data, salt="greedybear-feeds")
         token_hash = hashlib.sha256(token.encode()).hexdigest()
 
-        from greedybear.models import ShareToken
-
         self.assertFalse(ShareToken.objects.filter(token_hash=token_hash).exists())
 
         self.client.logout()
@@ -313,10 +310,6 @@ class FeedsEnhancementsTestCase(CustomTestCase):
 
     def test_consume_token_deleted_from_db_is_rejected(self):
         """A token whose DB record has been deleted is rejected even though the signature is valid."""
-        import hashlib
-
-        from greedybear.models import ShareToken
-
         share_response = self.client.get("/api/feeds/share?asn=11111")
         self.assertEqual(share_response.status_code, 200)
         token = share_response.json()["url"].split("/")[-1]
@@ -339,12 +332,6 @@ class FeedsEnhancementsTestCase(CustomTestCase):
         then verifies the second request within the window returns 429.
         cache.clear() is scoped to this test to avoid leaking throttle state.
         """
-        from unittest.mock import patch
-
-        from django.core.cache import cache
-
-        from api.throttles import SharedFeedRateThrottle
-
         share_response = self.client.get("/api/feeds/share")
         token = share_response.json()["url"].split("/")[-1]
         anon = APIClient()
--- a/tests/api/views/test_feeds_advanced_view.py
+++ b/tests/api/views/test_feeds_advanced_view.py
@@ -274,6 +274,63 @@ class FeedsEnhancementsTestCase(CustomTestCase):
         self.assertEqual(response.status_code, 400)
         self.assertIn("error", response.json())
 
+    def test_consume_valid_token_without_db_record_is_rejected(self):
+        """A valid signed token that was never saved to the DB is rejected (allowlist check)."""
+        import hashlib
+
+        from django.core import signing
+
+        data = {
+            "feed_type": "all",
+            "attack_type": "all",
+            "ioc_type": "all",
+            "max_age": "3",
+            "min_days_seen": "1",
+            "include_reputation": [],
+            "exclude_reputation": [],
+            "feed_size": "5000",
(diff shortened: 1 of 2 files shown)


R1sh0bh-1:
Hey @regulartim,
I’ve pushed the fix. Tested it the same way as mentioned in the issue and it works fine.
CI is failing with a startup error. I think it's a workflow permissions issue (didn't happen in my previous PRs). So I am not sure what to do, can you please take a look and guide me?

regulartim:
> CI is failing with a startup error. I think it's a workflow permissions issue (didn't happen in my previous PRs). So I am not sure what to do, can you please take a look and guide me?

That should be fixed already. I'll take a look.

R1sh0bh-1:
Thanks! All tests are now passing
Let me know if the PR looks good or if you want any changes.

regulartim (changes requested):
Hey @R1sh0bh-1 ! Could you please add tests that explicitly test the bug you solved?

R1sh0bh-1:
hey @regulartim, I added two regression tests for this one that checks a valid signed token with no DB record gets rejected, and one that checks a token stops working once its DB record is deleted. both pass. Please let me know if you'd like anything else!

tests/api/views/test_feeds_advanced_view.py
```diff
@@ -274,6 +274,63 @@ def test_shareable_feed_expired_token(self):
         self.assertEqual(response.status_code, 400)
         self.assertIn("error", response.json())
 
+    def test_consume_valid_token_without_db_record_is_rejected(self):
+        """A valid signed token that was never saved to the DB is rejected (allowlist check)."""
+        import hashlib
+
+        from django.core import signing
+
```
regulartim: Please move the in-function imports to the top of the test file.

tests/api/views/test_feeds_advanced_view.py
```diff
@@ -274,6 +274,63 @@ def test_shareable_feed_expired_token(self):
...
+            "asn": None,
+            "min_score": None,
+            "port": None,
+            "start_date": None,
+            "end_date": None,
+        }
+        token = signing.dumps(data, salt="greedybear-feeds")
+        token_hash = hashlib.sha256(token.encode()).hexdigest()
+
+        from greedybear.models import ShareToken
```
regulartim: Also here and in the rest of the file

regulartim (changes requested):

R1sh0bh-1:
Sorry about that missed moving those imports up when I added the tests. fixed now, all imports are at the module level. thanks for catching it!

regulartim (approved):
Nice, thanks! :)

PR #1219, opened 2026-04-09
Cache get_valid_feed_types() to reduce DB load in feed endpoints

### LINKED ISSUE #952: Optimization: Caching `get_valid_feed_types()`

SupRaKoshti:
# Description

enhancement: cache `get_valid_feed_types()` to reduce database load on feed endpoints

All four feed endpoints invoke `get_valid_feed_types()` on every request, leading to redundant database connection checkouts. Since active honeypot records change infrequently (~20–30 rows), the result is now cached with a 15-minute TTL.

**Changes:**

- Cache the return value of `get_valid_feed_types()` under the key `"valid_feed_types"` with a 15-minute TTL in `api/views/utils.py`
- Introduce `greedybear/signals.py` with `post_save` and `post_delete` receivers on the `Honeypot` model to immediately invalidate the cache whenever honeypot configuration is modified
- Register the signals module via `GreedyBearConfig.ready()` in `greedybear/apps.py`

Closes #952

---

### Related issues

#952 Optimization: Caching `get_valid_feed_types()`

---

### Type of change

---

# Checklist

### Formalities

- [x] I have read and understood the rules about [[how to Contribute](https://intelowlproject.github.io/docs/GreedyBear/Contribute/)](https://intelowlproject.github.io/docs/GreedyBear/Contribute/) to this project.
- [x] I have chosen an appropriate title for the pull request in the form: `<feature name>. Closes #999`
- [x] The pull request targets the `develop` branch.

### Docs and tests

- [x] I have documented my code changes with docstrings and/or comments.
- [x] I have verified whether my changes affect user-facing behavior described in the [[docs](https://intelowlproject.github.io/docs/GreedyBear/Introduction/)](https://intelowlproject.github.io/docs/GreedyBear/Introduction/). If so, I have also submitted a pull request to the [[docs repository](https://github.com/intelowlproject/docs)](https://github.com/intelowlproject/docs).
- [x] Linter (`Ruff`) reported 0 errors. If [[pre-commit](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance)](https://intelowlproject.github.io/docs/GreedyBear/Contribute/#how-to-start-setup-project-and-development-instance) is correctly installed, these checks and fixes are applied automatically.
- [ ] I have added tests covering the feature/bug addressed in this PR.
- [x] All tests passed with 0 errors.

### DIFF
--- a/api/views/utils.py
+++ b/api/views/utils.py
@@ -146,9 +146,16 @@ def get_valid_feed_types() -> frozenset[str]:
     Returns:
         frozenset[str]: An immutable set of valid feed type strings
     """
+    cache_key = "valid_feed_types"
+    cached_types = cache.get(cache_key)
+    if cached_types is not None:
+        return cached_types
+
     honeypots = Honeypot.objects.filter(active=True)
     feed_types = ["all"] + [hp.name.lower() for hp in honeypots]
-    return frozenset(feed_types)
+    result = frozenset(feed_types)
+    cache.set(cache_key, result, (settings.EXTRACTION_INTERVAL + 5) * 60)
+    return result
 
 
 def get_queryset(request, feed_params, valid_feed_types, is_aggregated=False, serializer_class=FeedsRequestSerializer, tag_key="", tag_value=""):
--- a/greedybear/apps.py
+++ b/greedybear/apps.py
@@ -8,3 +8,4 @@ class GreedyBearConfig(AppConfig):
 
     def ready(self):
         import greedybear.checks  # noqa: F401 — register system checks
+        import greedybear.signals  # noqa: F401 — register signals
--- a/greedybear/signals.py
+++ b/greedybear/signals.py
@@ -1,0 +1,11 @@
+from django.core.cache import cache
+from django.db.models.signals import post_delete, post_save
+from django.dispatch import receiver
+
+from greedybear.models import Honeypot
+
+
+@receiver(post_save, sender=Honeypot)
+@receiver(post_delete, sender=Honeypot)
+def invalidate_valid_feed_types_cache(sender, instance, **kwargs):
+    cache.delete("valid_feed_types")


SupRaKoshti:
Hi @regulartim , ready for review! please take a look when you get a chance.

regulartim (changes requested):
Hey @SupRaKoshti ! I see a conceptual issue here. I think you are using LocMemCache here, which is not shared across workers. The same might be true for the signals. Only the cache of the worker that handles the signal would get invalidated. Would you please check if my assumption is true? If so, what other options do we have?

SupRaKoshti:
> Hey @SupRaKoshti ! I see a conceptual issue here. I think you are using LocMemCache here, which is not shared across workers. The same might be true for the signals. Only the cache of the worker that handles the signal would get invalidated. Would you please check if my assumption is true? If so, what other options do we have?

Hi @regulartim , yes, you're correct! LocMemCache is per-process, so signal invalidation only affects the worker that handled the write — other workers would continue serving stale data until the TTL expires.

One option is to switch the default cache backend to RedisCache, pointing at the existing Redis instance already in the stack. That would give all workers a shared cache, so a single signal invalidation clears it everywhere.

Let me research further to see if there are other viable approaches as well. I'll get back to you in a day or two.

regulartim:
Hey @SupRaKoshti ! 

>  pointing at the existing Redis instance already in the stack

There is no Redis in our stack.

> Let me research further to see if there are other viable approaches as well. I'll get back to you in a day or two.

Alright! :) There is a DB-based cache (Django-Q) but I don't think that using it makes any sense here.

SupRaKoshti:
Hi @regulartim, agreed — DatabaseCache still relies on the database, so while it can reduce computation and simplify queries, it doesn’t eliminate database load.

After researching, here are the viable options I found:

**Option 1: FileBasedCache (Django built-in)**
FileBasedCache is shared across Gunicorn workers and requires no extra dependencies. It’s faster than database queries but slower than in-memory caches. It does not use robust file locking, so race conditions are theoretically possible — low risk for read-heavy, non-critical caching, but not ideal for high-concurrency write scenarios.

**Option 2: Memcached or Redis (external shared cache)**
The proper long-term solution — shared across all workers with instant signal invalidation. Memcached is lighter than Redis and purpose-built for caching. However, both require adding a new service to the stack.

My preference is **Option 2** as it directly solves the issue and is the better long-term solution. We can decide between Memcached and Redis based on what fits the project best.

regulartim:
Hey @SupRaKoshti ! Thanks for investigating! 

> My preference is Option 2 as it directly solves the issue and is the better long-term solution. We can decide between Memcached and Redis based on what fits the project best.

I agree that this is the better option, but right now we really only have this single use case for it. Hard do justify a new dependency in the docker stack just to save a single (and quite cheap) DB query, which is probably cached by the DB anyway. 

> Option 1: FileBasedCache
> It’s faster than database queries but slower than in-memory caches.

I would guess that the advantage over (simple) DB queries is not significant, right? 

Sorry, but I tend to close this PR and the issue as "won't fix". I think the marginal performance bump we would gain is just not worth the added complexity here.

SupRaKoshti:
Hi @regulartim , that makes complete sense — I agree that the added complexity for a single use case isn't worth it, especially when the DB is likely caching the query anyway. You're right about FileBasedCache too — the DB query is essentially a memory lookup already (because DB cache), so FileBasedCache wouldn't offer any meaningful advantage. Thanks for walking through the options with me and for the thorough explanation. I learned a lot about cache backend tradeoffs from this discussion. Feel free to close the PR and the issue!

PR #1238, opened 2026-04-15
Fixfrontendbug#71

aditisingh70330:
I investigated the reported AppHeader rendering behavior by observing component renders during navigation and authentication state changes. The behavior appears consistent with expected React patterns, where re-renders are triggered by state and route updates.

I also reviewed the loading flow and verified the FallbackLoading component from @certego/certego-ui. It correctly handles suspended routes and provides a consistent UI fallback during lazy loading, ensuring smooth transitions while async components are resolved.

Additionally, I observed 502 Bad Gateway responses from /api/auth/authentication and /api/news, indicating that the backend/API service is not reachable in the current setup, which limits full end-to-end validation of the flow.

From the frontend perspective, the AppHeader rendering logic and loading behavior appear stable based on current observable behavior.

### DIFF
--- a/greedybear/cronjobs/extraction/utils.py
+++ b/greedybear/cronjobs/extraction/utils.py
@@ -198,18 +198,6 @@ def iocs_from_hits(hits: list[dict]) -> list[IOC]:
 
 
 
-def get_ioc_type(ioc: str) -> str:
-    """
-    Determine the type of an IOC based on its format.
-
-    Args:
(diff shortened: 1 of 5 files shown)


regulartim:
Hey @aditisingh70330 ! There are a lot of problems with this PR. For example it targets the wrong branch (that's why I will have to close it). Please go through the [contribution guidelines](https://github.com/GreedyBear-Project/GreedyBear/wiki/Contribute) again.

PR #1239, opened 2026-04-15
feat: add ?reason=param to feeds_share and token list endpoint. Closes #1224

### LINKED ISSUE #1224: Feature: ShareToken - Token List Endpoint & Reason Label
I went through the shareable feeds feature introduced in #839, really nice addition. While looking at it, I noticed two small gaps worth addressing as a follow-up.

1. The ShareToken model already has a reason field (added in migration 0041_sharetoken), but feeds_share never populates it. It only stores user and token_hash:
```python
ShareToken.objects.get_or_create(token_hash=token_hash, defaults={"user": request.user})
```

2. No way to list your own tokens: There's currently no endpoint to see what share tokens you've issued.

### Proposed Solutions:

 - **Accept an optional `?reason=` parameter on `GET /api/feeds/share`**

 - **Add `GET /api/feeds/tokens/` (authenticated)** — returns only the calling user's tokens with safe metadata (no raw token, just a hash prefix):

opbot-xd:
# Description

Addresses two gaps in the shareable feeds feature (#839):

1. **`?reason=` on `GET /api/feeds/share`** — The `ShareToken` model already has a `reason` field (added in migration `0041_sharetoken`), but `feeds_share` never populated it. Now an optional `?reason=` query parameter is accepted and persisted (truncated to 256 chars). Existing tokens are not overwritten thanks to `get_or_create` semantics.

2. **`GET /api/feeds/tokens/`** (authenticated) — New endpoint that returns only the calling user's tokens with safe metadata: `hash_prefix` (first 12 hex chars), `reason`, `created_at`, `revoked`, `revoked_at`. The raw token is never exposed.

### Related issues

Closes #1224

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes

N/A — backend-only change, no GUI modifications.


regulartim (approved):
Looks good to me, thanks! :) 
@R1sh0bh-1 , would you also take a look please?

api/views/feeds.py
```diff
@@ -323,3 +327,32 @@ def feeds_revoke(request, token):
     share_token.revoked_at = timezone.now()
     share_token.save(update_fields=["revoked", "revoked_at"])
     return Response({"detail": "Token revoked successfully."}, status=status.HTTP_200_OK)
+
+
+@api_view([GET])
```
R1sh0bh-1, line 332: Hey @opbot-xd, great work building out this new endpoint! One quick thought: since a user could potentially generate hundreds of share tokens over time, should we add pagination here (maybe using CustomPageNumberPagination)? It'll help keep the API fast and make it consistent with our other list endpoints.

api/views/feeds.py
```diff
@@ -219,20 +227,27 @@ def feeds_share(request):
...
     """
-    logger.info(f"request /api/feeds/share with params: {request.query_params}")
+    safe_params = {k: v for k, v in request.query_params.items() if k != "reason"}
+    logger.info(f"request /api/feeds/share with params: {safe_params}")
     feed_params = FeedRequestParams(request.query_params)
     data = vars(feed_params)
     # Remove internal or non-serializable objects if any
     data.pop("feed_type_sorting", None)
 
+    reason = request.query_params.get("reason", "").strip()[:256]
```
R1sh0bh-1, line 242: Nice defensive check on the string length! I noticed a minor UX edge case here though: because reason is placed inside the defaults argument, it only gets saved the very first time the token is created. If a user makes a typo in their reason and tries to fix it by generating the link again, the typo will stay there and the new reason will be ignored. Could we add a quick check to update the reason if they provide a new one for an existing token?

R1sh0bh-1:
@regulartim, I’ve completed my review as you requested and left a couple of minor suggestions. Please let me know your thoughts on them whenever you have a chance.

api/views/feeds.py
regulartim: I think it's fine without pagination, honestly. Even if a user had a thousand tokens, this APIs response would still have a reasonable size.

R1sh0bh-1: Makes sense. I was just thinking ahead for edge cases, but you're spot on that the raw JSON size would remain completely manageable anyway. Sounds good to me!

api/views/feeds.py
regulartim: You identified a real UX edge case, but I think keeping it the way it is: existing tokens are not overwritten or changed. This is a simple and clear behavior that makes things much easier, in my opinion.

regulartim:
@R1sh0bh-1 Thanks a lot for your review and you thoughtful comments!

PR #1297, opened 2026-04-23
Enhancement: bulk prefetch IOCs before processing to eliminate per-IOC SELECT queries. Closes #1241

### LINKED ISSUE #1241: Optimization: Bulk Enrichment of IOCs
Hi @regulartim, 
I had a thought on a safe optimization we can queue up, in which right now the IOC_from_hits path is functionally right, but it still queries per unique IP. 
So even when we process a huge batch efficiently in memory, DB calls still grow as 3 + 4U + 1 (U = new src_ip). That’s feels expensive. We can keep all current behavior exactly the same and just shift enrichment internals to batch operations so we stop repeating the same lookups per IOC.

If we do that, query growth drops to about 6 + U (plus a small bounded M2M thing), which is a big improvement. 

Like for Example at 1,000 unique ips, we go from roughly 4,004 queries to ~1,006, so about a 75% reduction with no intended change in output. 
The practical win is better throughput headroom, more stable worker latency, and less DB pressure during spikes 

We can do it by writing one optimized enrichment path.
Please share your thoughts on it

rahulgunwanistudy-2005:
# Description

The extraction pipeline was calling `ioc_repo.get_ioc_by_name()` once per IOC inside `add_ioc()`, so a batch of U unique IPs meant U SELECT queries. On top of that, `_add_fks()` in `BaseExtractionStrategy` was re-fetching IOCs that `add_ioc()` had just processed. Tanner had a manual local `ioc_cache` in `_classify_attacks()` as a partial workaround for the same problem.

This PR fixes all three by introducing a bulk prefetch mechanism in `IocProcessor`:
- Added `_ioc_cache` and `prefetch_iocs()` to `IocProcessor`. Each `_get_scanners()` now bulk-loads all IOCs for the batch in one query before the loop, reducing U SELECTs to 1.
- `add_ioc()` checks the cache first, falls back to the repo only on a miss, and updates the cache after every save.
- `_add_fks()` now builds an in-memory map from `self.ioc_records` instead of making extra DB calls for IOCs already processed.
- Tanner's manual local `ioc_cache` in `_classify_attacks()` is removed and replaced by the processor-level cache.

### Related issues
Closes #1241

### Type of change

# Checklist

### Formalities
- [x] I chose an appropriate title for the pull request in the form: `Enhancement: bulk prefetch IOCs before processing to eliminate per-IOC SELECT queries. Closes #1241`

### Docs and tests
- [x] I have checked if my changes affect user-facing behavior that is described in the docs. No user-facing behavior is affected.
- [x] Linter (Ruff) gave 0 errors.

### GUI changes
No GUI changes in this PR.

### DIFF
--- a/greedybear/cronjobs/extraction/ioc_processor.py
+++ b/greedybear/cronjobs/extraction/ioc_processor.py
@@ -26,6 +26,21 @@ class IocProcessor:
         self.ioc_repo = ioc_repo
         self.sensor_repo = sensor_repo
         self._whatsmyip_domains = set(WhatsMyIPDomain.objects.values_list("domain", flat=True))
+        self._ioc_cache: dict[str, IOC] = {}
+
+    def prefetch_iocs(self, names: list[str]) -> None:
+        """
+        Bulk prefetch existing IOC records into the cache.
+        Eliminates per-IOC SELECT queries in add_ioc() by loading all
+        known IOCs for the current batch in a single query.
+
+        Args:
+            names: List of IOC names (IPs or domains) to prefetch.
+        """
+        self._ioc_cache = {
+            ioc.name: ioc
+            for ioc in IOC.objects.prefetch_related("honeypots").filter(name__in=names)
+        }
 
     def add_ioc(
         self,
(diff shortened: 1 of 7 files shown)


regulartim (commented):
Hey @rahulgunwanistudy-2005 ! This is a tough one for me. Your idea and the implementation seems fine, but it adds quite a bit of complexity to the already quite complex extraction process, which is a core component of our application. 

I tested it anyways but could not measure a significant performance benefit. On a typical extraction run with ~200 IoCs I could not measure any difference in execution time. The reason might be that we are making queries to an indexed table. On a warm Postgres these queries are extremely cheap. 

I really value your work, but I will close this without merging. Sorry.

rahulgunwanistudy-2005:
Hi @regulartim, that makes total sense.
Sorry for the late reply...
Thanks for the honest feedback, will keep it in mind!

PR #1369, opened 2026-05-26
feat: Sensor creation API and Testcases. Closes #1359

### LINKED ISSUE #1359: [Event Collector API] - Sensor initialization API + test suite
Description

The goal is to create API endpoint for creating and managing sensors via Event Collector API.

Tasks
- Create Read and Write serializers 
- Create sensor creation endpoint
- Add test case

Outcome
- External users can create and manage sensors through API securely.

Motivation:
- Since the Event API requires a sensor_id, following the workflow of sensor_creation -> event_creation makes for a much smoother user flow.

Depends on : #1358

drona-gyawali:
## Sensor Create API

Available from version >= x.x.x

For authenticated users, GreedyBear provides an API to create or retrieve sensor records using an IP address as the primary identifier. This API is designed to be idempotent, meaning repeated requests with the same address will not create duplicate sensors.

Each request is associated with the user’s APISource, which is created and managed by system administrators. If no APISource is linked to the user, the request will be rejected.

You can access this API endpoint using the following URL:
```
https://<greedybear_site>/api/sensor/
```

## Authentication

This API is protected through authentication. Please reach out [Matteo Lodi](https://twitter.com/matte_lodi) or another member of [The Honeynet Project](https://twitter.com/ProjectHoneynet) if you are interested in gaining access to this API on the [Honeynet instance](https://greedybear.honeynet.org/) of GreedyBear.

* Authentication Method: CookieTokenAuthentication

* Requirement: Each user must have a valid and active APISource linked to their account. APISource is provisioned by system administrators and cannot be created by users. If no APISource is linked, or if the APISource is inactive, the request will be rejected with a 403 Forbidden response.

* Unauthorized users or users without APISource will receive a 403 response

**APISource Setup (for Administrators)**

APISource records are managed exclusively by system administrators via the Django Admin interface.

To create an APISource:

1. Log in to the Django Admin panel.
2. Navigate to the APISource section.
3. Click “Add APISource”.
4. Select the user account to associate with this APISource.
5. Set a name for identification (e.g. “User A API Source”).
6. Ensure the APISource is marked as active.

Once saved, the user will immediately be able to access API endpoints using their assigned APISource.

## Request Body

The following parameters are supported:

- address (required) : IP address of the sensor
- honeypot_type (optional) : free text, description of what kind of honeypot e.g. ssh
- honeypot_software (optional) : free text, e.g. Cowrie
- honeypot_description (optional) : free text, human readable description of the honeypot
- sensor_label (optional) : free text, identifying name of the sensor
- group_label (optional) : free text, name of a group of sensors that this sensor belongs to
- country_code (optional) : 2 characters, ISO code of the country where the sensor is located, e.g. FR
- asn (optional) : integer, AS number of the AS where the sensor is located, e.g. 12345

### Response (201 Created)

Returned when a new sensor is successfully created.

```json
{
  "id": 123,
  "message": "Sensor created successfully"
}
```

### Response (200 OK)

Returned when the sensor already exists.

```json
{
  "id": 123,
  "message": "Sensor already existed"
}
```

### Response (400 Bad Request)

Invalid input data such as:

* malformed IP address
* invalid country code
* invalid ASN format

### Response (403 Forbidden)

* User is not authenticated
* No APISource linked to the user account

---
 
### Related issues
closes: #1359 

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes

# Review process


regulartim (changes requested):
Hey @drona-gyawali ! I'll review your code tomorrow. In the mean time, could you a) please add a docstring to `sensor_create_view`, roughly in the same style as the Feeds API endpoints and b) provide a text in markdown format that I can add to the "Usage" page of the wiki? For b) the Cowrie Session API description in the wiki can be taken as a good example. 

Edit: you can put the text for b) into the PR description.

api/views/sensor.py
```diff
@@ -0,0 +1,89 @@
...
+        },
+    )
+
+    return Response(
+        {
+            "id": sensor.id,
+            "message": ("Sensor created successfully" if created else "Sensor already existed"),
+        },
+        status=status.HTTP_201_CREATED if created else status.HTTP_200_OK,
+    )
```
regulartim: What should happen when user A wants to create a sensor with IP 1.2.3.4, but user B already has a sensor with this IP? I am not sure...

Options I see:
- return 200 (as is)
- return 403 forbidden, without an ID
- loosen the address uniquenes

What's your opinion?

api/serializers.py
```diff
@@ -251,3 +251,61 @@ class FeedsResponseSerializer(serializers.Serializer):
...
         return [feed_type_validation(feed, self.context["valid_feed_types"]) for feed in feed_type]
+
+
+class SensorCreateSerializer(serializers.ModelSerializer):
+    sensor_label = serializers.CharField(source="label", required=False, allow_blank=True, max_length=128)
+
+    asn = serializers.IntegerField(
+        required=False,
+        allow_null=True,
+    )
```
regulartim, line 263: Could you enforce a positive integer here?

api/serializers.py
```diff
@@ -251,3 +251,61 @@ class FeedsResponseSerializer(serializers.Serializer):
...
+        extra_kwargs = {
+            "address": {"validators": []},
+        }
+
+        read_only_fields = ["id"]
+
+    def validate_country(self, value):
+        if value and len(value) != 2:
+            raise serializers.ValidationError("Country must be 2-character ISO code (e.g. 'NP', 'IN')")
+        return value.upper() if value else value
```
regulartim: You could use a regex here, to ensure that the country code only contains letters.

api/views/sensor.py
```diff
@@ -0,0 +1,89 @@
...
+    POST /api/sensor/
+
+    Args:
+        address (str, required): IPv4 or IPv6 address of the sensor.
+        honeypot_type (str, optional): Type of honeypot.
+        honeypot_software (str, optional): Honeypot software name.
+        honeypot_description (str, optional): Description of the sensor.
+        sensor_label (str, optional): Human-readable label.
+        group_label (str, optional): Group classification label.
+        country (str, optional): 2-letter ISO country code.
```
regulartim: If you expect users to send a country code, I would suggest to name the field "country_code".

regulartim: (I saw that I made the same mistake when writing down the requirements. But I think country_code is the better name here.)

api/serializers.py
```diff
@@ -251,3 +251,61 @@ class FeedsResponseSerializer(serializers.Serializer):
...
+        # this is done intentionally, we allow address duplication in this layer
+        # duplications are handled by get_or_create in a view logic.
+        extra_kwargs = {
+            "address": {"validators": []},
+        }
+
+        read_only_fields = ["id"]
+
+    def validate_country(self, value):
+        if value and len(value) != 2:
```
regulartim: As the country code is only 2 chars long, the model should also be restricted to 2 chars, but it has `max_length=64`. Can you please change the model accordingly? You can also rename the field to country_code. No data migration needed.
(Merge develop before doing that.)

api/serializers.py
```diff
@@ -251,3 +251,61 @@ class FeedsResponseSerializer(serializers.Serializer):
...
+            return validated_data
+
+        autonomous_system, _ = AutonomousSystem.objects.get_or_create(
+            asn=asn_value,
+            defaults={"name": ""},
+        )
+
+        validated_data["autonomous_system"] = autonomous_system
+
+        return validated_data
```
regulartim: I have to say that I am not a big fan of having DB writes inside a serializer. That's a very surprising place for that. Can you think of a better place?

regulartim (changes requested):
Hey @drona-gyawali ! Solid all in all! 👍

api/views/sensor.py
drona-gyawali: i think (option 3 )per-user uniqueness is the safer model here. with global uniqueness, one user sensor state or metadata could indirectly affect another user behavior, especially for fields like is_active.

i am thinking of scoping uniqueness to (api_source, address) instead of globally on address.

api/serializers.py
drona-gyawali: Regarding the country field update, since other core features and existing implementations already rely on the country column in the db, I think it is best to leave the db model completely untouched to avoid breaking anything or requiring a migration.

Instead, I propose we handle this entirely in the serializer:

- We expose the API field as country_code.

- We use source='country' mapping to read from and write to the existing country field in the database.

- We handle the 2-character limit and letter-only regex validation entirely within the serializer's validation layer.

I think we can also create a dedicated country_code field in the sensor model for this specific case or enforce  users to pass the full country name instead of a code in the API to match the existing column intent.

 what yours thought on this :) ?

api/serializers.py
drona-gyawali: yes, I think we can safely move this logic into api section.

api/views/sensor.py
regulartim: Yep, sounds reasonable!

api/serializers.py
regulartim: Ah, I thought the field was just added in your most recent PR, but it is already in `main`. Good that you thought about that! I am fine with what you proposed, I would prefer to create a dedicated, 2-char country_code field in the sensor model. I think it is cleaner that way. Do you agree?

drona-gyawali: yes, I agree

regulartim (commented):
Hey @drona-gyawali ! I tested your code and it works as expected. Two things though:
- In your text for the wiki you wrote: "Requirement: Each user must have a valid APISource linked to their account". A user might wonder how to create an APISource. Please also add a sentence about that. 
- You only check if an APISource exists when creating the sensor, not if it is active. Is that on purpose?

drona-gyawali:
> * You only check if an APISource exists when creating the sensor, not if it is active. Is that on purpose?

thanks for the feedback :)  I forgot to include that logic, but I'll add it shortly."

regulartim (approved):

regulartim:
I saved the text for the wiki in my note app and will add it once we release the feature.

PR #1397, opened 2026-06-04
feat: event ingestion API with asynchronous IOC processing pipeline . Closes #1363

### LINKED ISSUE #1363: [Event Collector API] - Implement Event Collector API ingestion and async processing pipeline
Description

The goal is to implement main Event Collector ingestion flow for external telemetry sources.

This includes:

* event submission API
* batch ingestion support
* raw event persistence
* async processing pipeline integration
* IOC extraction from ingested events

Incoming events should first be stored in `RawEvent` for full-fidelity persistence. After ingestion, an async task should process the events and integrate with the existing `iocs_from_hits` pipeline.

Tasks

-  Implement event submission API endpoint
- Implement Background Processing layer
- Add test case for API and Processor layer

 Outcome

External systems should be able to submit event.

Depends on:
- #1360

drona-gyawali:
## Event Ingestion API

Available from version >= x.x.x

For authenticated users, GreedyBear provides high-performance, asynchronous endpoints to ingest  event data in batches and track their background processing lifecycle.

To prevent payload parsing and database overhead from blocking HTTP responses, this API offloads execution to an asynchronous background task pipeline. The endpoint validates basic payload structures, bulk-stages the raw event data directly to the database to minimize I/O block time, and immediately hands off processing to Django-Q via an asynchronous task worker, returning a tracking ID to the client instantly.

### Endpoint URL

```http'
POST https://<greedybear_site>/api/events/add/

```
## Authentication

This API is protected through authentication. Please reach out [Matteo Lodi](https://twitter.com/matte_lodi) or another member of [The Honeynet Project](https://twitter.com/ProjectHoneynet) if you are interested in gaining access to this API on the [Honeynet instance](https://greedybear.honeynet.org/) of GreedyBear.

## Requirement

Each user must have a valid and active APISource linked to their account. APISource is provisioned by system administrators and cannot be created by users. If no APISource is linked, or if the APISource is inactive, the request will be rejected with a 403 Forbidden response.

To utilize these endpoints, each event in your payload must include a valid sensor_id. If you have not registered your sensor yet, you must first create or retrieve it using the [Sensor Create API](PLEASE_ADD_A_SESNOR_API_DOC_LINK) to obtain a valid ID.

## APISource Setup (for Administrators)

APISource records are managed exclusively by system administrators via the Django Admin interface.

To create an APISource:

1. Log in to the Django Admin panel.
2. Navigate to the APISource section.
3. Click “Add APISource”.
4. Select the user account to associate with this APISource.
5. Set a name for identification (e.g. “User A API Source”).
6. Ensure the APISource is marked as active.

Once saved, the user will immediately be able to access API endpoints using their assigned APISource.

### Request Configuration

* **Content-Type:** `application/json`
* **Batch Constraints:** Minimum **1** event, Maximum **10,000** events per payload.

### Request Body Fields

The JSON payload must contain a root key `"events"` mapping to an array of event objects.

#### Required Fields (Per Event)

| Parameter | Type | Validation / Constraints | Description |
| --- | --- | --- | --- |
| **`src_ip`** | String | Valid IP Address format | The source IP address that generated the event. |
| **`event_type`** | String | Max length: 100 | The category of the event (e.g., `bruteforce`, `connection`). |
| **`timestamp`** | String | ISO 8601 DateTime | When the event occurred. **Cannot be a future timestamp.** |
| **`sensor_id`** | Integer | Min value: 0 | The ID of the sensor. Must be provisioned beforehand. |

#### Optional Fields (Per Event)

| Parameter | Type | Default | Constraints / Description |
| --- | --- | --- | --- |
| **`session_id`** | String | `""` | Max length: 100. Unique honeypot session token. |
| **`token_id`** | String | `""` | Max length: 100. Identifier for unique authentication tokens. |
| **`protocol`** | String | `""` | Max length: 50. Network layer protocol (e.g., `tcp`, `udp`). |
| **`service_name`** | String | `""` | Max length: 100. Application service target (e.g., `ssh`, `cowrie`). |
| **`username`** | String | `""` | Max length: 255. Username used during authentication attempts. |
| **`password`** | String | `""` | Max length: 255. Password used during authentication attempts. |
| **`cve_id`** | String | `""` | Max length: 50. Target vulnerability exploit tracker. |
| **`command`** | String | `""` | Free-text. The raw command or shell payload executed. |
| **`src_port`** | Integer | `null` | Range: `1` to `65535`. Source port of the connection. |
| **`dest_port`** | Integer | `null` | Range: `1` to `65535`. Destination port of the connection. |
| **`related_url`** | String | `""` | Max length: 900. Valid URL downloaded/hit during attack. |
| **`payload_hash`** | String | `""` | Max length: 64. Hash string (SHA-256) of dropped malware. |
| **`data`** | Object | `{}` | Arbitrary nested JSON object for schema-less storage. |

#### `202 Accepted`

Returned when the payload passes structural validation, raw entries are staged, and the extraction queue successfully registers the job.

#### `400 Bad Request`

Returned if structural payload validation fails, required event fields are missing, timestamps are in the future, or no valid events could be saved.

#### `403 Forbidden`

Returned if the user is unauthenticated, lacks an administrative `APISource` link, or if their `APISource` record has been locked due to excessive payload malformations.

#### `500 Internal Server Error`

Returned if unforeseen infrastructure anomalies interrupt staging database I/O operations or background task enqueuing.

---

### Endpoint URL

This endpoint allows clients to view the processing state of an ingested event batch using its unique task_id.

```
GET https://<greedybear_site>/api/events/status/<task_id>/

```

## Authentication

This API is protected through authentication. Please reach out [Matteo Lodi](https://twitter.com/matte_lodi) or another member of [The Honeynet Project](https://twitter.com/ProjectHoneynet) if you are interested in gaining access to this API on the [Honeynet instance](https://greedybear.honeynet.org/) of GreedyBear.

### Parameters

* **`task_id`** (String): The unique string identifier (`UUID` format) received in the `202 Accepted` response from the [Event Ingestion API](PLEASE_ADD_DOC_LINK_OF_EVENT_STATUS_API).

#### `200 OK`

Returned when the tracking record exists for the authenticated user's account.

*Note: Possible values for the `status` string property include `pending`, `processing`, `completed`, and `failed`.*

#### `403 Forbidden`

Returned if the user is unauthenticated or the underlying `APISource` identity record is deactivated.

#### `404 Not Found`

Returned if the provided `task_id` does not match any asynchronous ingest lifecycle record tied to your specific `APISource`.

---

### Related issues
closes: #1363 

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes

# Review process

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -8,6 +8,8 @@ from api.views import (
     command_sequence_view,
     cowrie_session_view,
     enrichment_view,
+    event_status_view,
+    events_create_view,
     feeds,
     feeds_advanced,
     feeds_asn,
@@ -43,6 +45,8 @@ urlpatterns = [
     path("news/", news_view),
     path("health/", health_view),
     path("sensor/", sensor_create_view),
+    path("events/", events_create_view),
+    path("event/<str:task_id>/status/", event_status_view),
     # router viewsets
     path("", include(router.urls)),
     # certego_saas:
--- a/api/urls.py
+++ b/api/urls.py
@@ -45,8 +45,8 @@ urlpatterns = [
     path("news/", news_view),
     path("health/", health_view),
     path("sensor/", sensor_create_view),
-    path("events/", events_create_view),
-    path("event/<str:task_id>/status/", event_status_view),
+    path("events/add/", events_create_view),
+    path("events/status/<str:task_id>/", event_status_view),
     # router viewsets
     path("", include(router.urls)),
     # certego_saas:
--- a/api/views/event.py
+++ b/api/views/event.py
@@ -9,7 +9,7 @@ from rest_framework.response import Response
 
 from api.serializers import InjectionSerializer
 from api.views.utils import create_batch_and_events
-from greedybear.models import APISource, EventStatus
+from greedybear.models import APISource, EventStatus, EventStatusType
 
 logger = logging.getLogger(__name__)
 
@@ -66,6 +66,11 @@ def events_create_view(request):
             events_data,
             api_source,
         )
+    except ValueError as e:
+        api_source.invalid_event_count += 1
+        api_source.save(update_fields=["invalid_event_count"])
+        return Response({"error": str(e)}, status=status.HTTP_400_BAD_REQUEST)
+
     except Exception:
         logger.exception("Failed while creating batch & events")
         return Response({"error": "An internal database error occurred while staging events"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
@@ -76,12 +81,22 @@ def events_create_view(request):
         batch.save(update_fields=["status", "last_error"])
         return Response({"error": "No valid events could be stored"}, status=status.HTTP_400_BAD_REQUEST)
 
-    # enqueue background task
-    async_task(
-        "greedybear.process_event.process_incoming_event",
-        api_source.id,
-        batch.task_id,
-    )
+    try:
+        # enqueue background task
+        async_task(
+            "greedybear.process_event.process_incoming_event",
+            api_source.id,
+            batch.task_id,
+        )
+    except Exception as e:
+        logger.exception(f"Failed to enqueue background task for batch {batch.task_id}")
+
+        # marking the batch as failed so it doesn't get orphaned in a 'PENDING' state
+        batch.status = EventStatusType.FAILED
+        batch.last_error = f"Background task dispatch failed: {e!s}"
+        batch.save(update_fields=["status", "last_error"])
+
+        return Response({"error": "An internal error occurred while queueing events for processing"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
     logger.info(f"[task={batch.task_id}] Accepted {total_created} events — source={api_source.name}")
 
@@ -89,7 +104,7 @@ def events_create_view(request):
         {
             "message": f"{total_created} events accepted for processing",
             "task_id": batch.task_id,
-            "status_url": f"/api/event/{batch.task_id}/status/",
+            "status_url": f"/api/events/status/{batch.task_id}/",
         },
         status=status.HTTP_202_ACCEPTED,
     )
--- a/api/views/event.py
+++ b/api/views/event.py
@@ -1,0 +1,147 @@
+import logging
+
+from certego_saas.apps.auth.backend import CookieTokenAuthentication
+from django_q.tasks import async_task
+from rest_framework import status
+from rest_framework.decorators import api_view, authentication_classes, permission_classes
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from api.serializers import InjectionSerializer
+from api.views.utils import create_batch_and_events
+from greedybear.models import APISource, EventStatus
+
+logger = logging.getLogger(__name__)
+
+
+@api_view(["POST"])
+@authentication_classes([CookieTokenAuthentication])
+@permission_classes([IsAuthenticated])
+def events_create_view(request):
+    """
+    Ingest a batch of raw security events, persist them, and hand off processing to a background task.
+
+    This endpoint validates the payload structure, maps events to a tracking batch, saves
+    them bulk-style to the database to minimize I/O overhead, and offloads heavy parsing
+    (extracting IOCs, usernames, commands, etc.) asynchronously via Django-Q.
+
+    **Authentication:**
+        - Required: CookieTokenAuthentication
+        - User must have an active and valid associated `APISource`.
+
+    **Payload Requirements:**
+        - Content-Type: application/json
+        - Structure: {"events": [ {...event_1...}, {...event_2...} ]}
+        - Batch constraints: Min 1 event, Max 10,000 events per request.
+
+    **Responses:**
+        - `202 Accepted`: Payload verified and successfully queued for background extraction.
+        - `400 Bad Request`: Validation failure or empty event set creation.
+        - `403 Forbidden`: Unauthenticated, missing `APISource`, or locked account state.
+    """
+    try:
+        api_source = request.user.api_source
+    except APISource.DoesNotExist:
+        return Response(
+            {"error": "No APISource linked to your account"},
+            status=status.HTTP_403_FORBIDDEN,
+        )
+
+    if not api_source.is_active:
+        return Response(
+            {"error": "APISource is locked"},
+            status=status.HTTP_403_FORBIDDEN,
+        )
+
+    serializer = InjectionSerializer(data=request.data)
+    if not serializer.is_valid():
+        api_source.invalid_event_count += 1
+        api_source.save(update_fields=["invalid_event_count"])
+        return Response({"error": "Invalid data", "details": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+    events_data = serializer.validated_data["events"]
+
+    try:
+        batch, total_created = create_batch_and_events(
+            events_data,
+            api_source,
+        )
+    except Exception:
+        logger.exception("Failed while creating batch & events")
+        return Response({"error": "An internal database error occurred while staging events"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+    if total_created == 0:
+        batch.status = "failed"
+        batch.last_error = "No valid events could be stored"
+        batch.save(update_fields=["status", "last_error"])
+        return Response({"error": "No valid events could be stored"}, status=status.HTTP_400_BAD_REQUEST)
+
+    # enqueue background task
+    async_task(
+        "greedybear.process_event.process_incoming_event",
+        api_source.id,
+        batch.task_id,
+    )
+
+    logger.info(f"[task={batch.task_id}] Accepted {total_created} events — source={api_source.name}")
+
+    return Response(
+        {
+            "message": f"{total_created} events accepted for processing",
+            "task_id": batch.task_id,
+            "status_url": f"/api/event/{batch.task_id}/status/",
+        },
+        status=status.HTTP_202_ACCEPTED,
+    )
+
+
+@api_view(["GET"])
+@authentication_classes([CookieTokenAuthentication])
+@permission_classes([IsAuthenticated])
+def event_status_view(request, task_id: str):
+    """
+    Retrieve the execution and processing status of a specific event batch using its task ID.
+
+    This endpoint safely exposes the processing lifecycle phase (e.g., pending, running, success, failed)
+    of an asynchronous background extraction task triggered by Django-Q.
+
+    **Authentication:**
+        - Required: CookieTokenAuthentication
+        - User must have an active, valid associated `APISource`.
+
+    **Path Parameters:**
+        - `task_id` (str): The unique string identifier assigned to the background processing job.
+
+    **Responses:**
+        - `200 OK`: Success payload detailing batch metrics, failure reasons (if any), and state.
+        - `403 Forbidden`: Unauthenticated, missing `APISource`, or locked account state.
+        - `404 Not Found`: No event batch matching the provided `task_id` exists for this account.
+    """
+    try:
+        api_source = request.user.api_source
+    except APISource.DoesNotExist:
+        return Response({"error": "No APISource linked to your account"}, status=status.HTTP_403_FORBIDDEN)
+
+    if not api_source.is_active:
+        return Response(
+            {"error": "APISource is locked"},
+            status=status.HTTP_403_FORBIDDEN,
+        )
+
+    try:
+        batch = EventStatus.objects.get(task_id=task_id, api_source=api_source)
+    except EventStatus.DoesNotExist:
+        return Response({"error": f"No batch found for task_id={task_id}"}, status=status.HTTP_404_NOT_FOUND)
+
+    return Response(
+        {
+            "task_id": batch.task_id,
+            "batch_id": batch.id,
+            "status": batch.status,
+            "ioc_count": batch.ioc_count,
+            "last_error": batch.last_error or None,
+            "processed_at": batch.processed_at,
+            "created_at": batch.created_at,
+        },
+        status=status.HTTP_200_OK,
+    )
--- a/api/views/event.py
+++ b/api/views/event.py
@@ -111,7 +111,7 @@ def event_status_view(request, task_id: str):
     """
     Retrieve the execution and processing status of a specific event batch using its task ID.
 
-    This endpoint safely exposes the processing lifecycle phase (e.g., pending, running, success, failed)
+    This endpoint safely exposes the processing lifecycle phase (e.g., pending, processing, completed, failed)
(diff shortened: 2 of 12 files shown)


drona-gyawali:
Hi Tim, Just finished writing the first version , If you're using Docker, please ensure `DJANGO_SECRET_KEY` is set in your environment in env_file:

 If left empty, the app generates a random key on startup. This causes Django-Q to fail because the web server and the workers will end up using different keys. Setting a fixed key avoids validation errors during local testing.

I'll add  docs incrementally soon. The core logic is about 670+ lines of code, with the majority of the rest being test cases.

Waiting for your valuable feedback :)

regulartim:
Hi @drona-gyawali ! :) Wow, your progressing fast! I currently plan to review this on Monday. 

> If left empty, the app generates a random key on startup. This causes Django-Q to fail because the web server and the workers will end up using different keys. Setting a fixed key avoids validation errors during local testing.

Is this a requirement that is introduced by this PR? Can you explain why exactly this is necessary?

drona-gyawali:
> Is this a requirement that is introduced by this PR? Can you explain why exactly this is necessary?

No, this is not a requirement introduced by this PR, but rather an intentional adjustment to handle how our local development environment behaves across multiple processes.

In settings.py, the current implementation is(main branch):

```python

```
currently, our architecture relies on two separate services running concurrently:

1. Webserver: enqueues background tasks.
2.  Django-Q : Consumes and processes those background tasks 

During local development, we often don't have a DJANGO_SECRET environment variable explicitly configured. As a result:

The webserver spins up, finds no env variable, and generates Random Key X.
 The Django-Q worker spins up as a completely separate OS process, finds no env variable, and generates Random Key Y.

Because Django-Q relies on the `SECRET_KEY` to sign, serialize, and validate task payloads, this mismatch causes a `BadSignature` error. The worker ultimately rejects the tasks pushed by the webserver because the cryptographic signatures do not match.

regulartim:
Thanks for the explanation. 

> During local development, we often don't have a DJANGO_SECRET environment variable explicitly configured.

I fear that this is also true for many production environments. It was never a problem that both services don't share the same secret. But I think this PR introduced the first feature, where it matter: a task signed in app but verified in qcluster. 

So without manual intervention, production environments that don't have an explicit `DJANGO_SECRET` set will break. Am I right?

api/views/utils.py
```diff
@@ -686,3 +687,69 @@ def create_or_get_sensor(*, api_source, validated_data):
...
+    total = 0
+    try:
+        for i in range(0, len(raw_events), chunk_size):
+            chunk = raw_events[i : i + chunk_size]
+            RawEvent.objects.bulk_create(chunk)
+            total += len(chunk)
+    except Exception:
+        logger.exception(f"Database error during bulk-insert for batch {batch.task_id}")
+        return total
+
```
regulartim: I don't think we should swallow the exception here. It will prevent the DB from performing the rollback. I would suggest we log (as you did) and re-raise it. It has to reach the `@transaction.atomic` somehow, such that a rollback is triggered. Also, maybe set the batch status to "failed" in that case.

api/views/event.py
```diff
@@ -0,0 +1,147 @@
...
+        batch.last_error = "No valid events could be stored"
+        batch.save(update_fields=["status", "last_error"])
+        return Response({"error": "No valid events could be stored"}, status=status.HTTP_400_BAD_REQUEST)
+
+    # enqueue background task
+    async_task(
+        "greedybear.process_event.process_incoming_event",
+        api_source.id,
+        batch.task_id,
+    )
```
regulartim: What if this fails? Then we end up with a 500 (fine for me) and a orphaned event batch. I think I would wrap this in a try/except and mark the batch as "failed" if it throws. What do you think?

api/views/event.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if not api_source.is_active:
+        return Response(
+            {"error": "APISource is locked"},
+            status=status.HTTP_403_FORBIDDEN,
+        )
+
+    serializer = InjectionSerializer(data=request.data)
+    if not serializer.is_valid():
+        api_source.invalid_event_count += 1
```
regulartim: `invalid_event_count` is incremented but nothing ever happens with it. Is this functionality coming in a future PR?

api/urls.py
```diff
@@ -43,6 +45,8 @@
     path("news/", news_view),
     path("health/", health_view),
     path("sensor/", sensor_create_view),
+    path("events/", events_create_view),
+    path("event/<str:task_id>/status/", event_status_view),
```
regulartim: I find these paths a little weird. One might wonder: Why is the create path plural but the status path singular? Also, what if we want to add another event-related endpoint in the future? Suggestion:
- events/add/
- events/status/<task_id>

What do you think?

api/views/utils.py
regulartim, line 1: Could you add type hints to the new functions?

greedybear/process_event.py
```diff
@@ -0,0 +1,281 @@
...
+
+def _normalize_raw_event_to_hit(raw):
+    """
+    Convert a RawEvent into the hit-dict format expected by iocs_from_hits.
+    """
+    # one bad event shuouldn't crash the whole batch.
+    if not raw.src_ip:
+        logger.warning(f"RawEvent pk={raw.pk} missing src_ip — skipping")
+        return None
+
```
regulartim: Can this happen? If so, under which circumstances?

api/views/utils.py
```diff
@@ -686,3 +687,69 @@ def create_or_get_sensor(*, api_source, validated_data):
...
+        sensor = sensors_by_id.get(sensor_id)
+
+        if sensor is None:
+            logger.warning(
+                "sensor_id=%s not found or not owned by api_source=%s",
+                sensor_id,
+                api_source.id,
+            )
+            continue
+
```
regulartim: This happens if a event is received, containing a sensor, which has not been created yet, correct? If so, I think we should treat the whole batch as invalid. The reason that I am so strict here is that a) I want the user to follow the creation flow we intended and b) I'd always prefer an error over some data just silently disappearing. And the "wrong" sensor information would just silently disappear in the process, right?

greedybear/process_event.py
```diff
@@ -0,0 +1,281 @@
...
+    if batch.status in ["processing", "completed"]:
+        logger.warning(f"[task={task_id}] Batch already {batch.status} — skipping to avoid race condition")
+        return
+
+    batch.status = "processing"
+    batch.ioc_count = 0
+    batch.last_error = ""
+    batch.save(update_fields=["status", "ioc_count", "last_error"])
+    logger.info(f"[batch={task_id}] Started (source_id={source_id})")
```
regulartim, line 169: You added some logging with log level "info". I a lot of batches are injected, this can fill up the logs quickly. Suggestion: leave the "started" and "finished" on level "info", change the rest of the "info" logs to "debug".

regulartim (changes requested):
All in all very solid! Good work! 👍 
Generally, if you would add type hints on the helper functions, it would make the code much easier to read.

drona-gyawali:
> I fear that this is also true for many production environments. It was never a problem that both services don't share the same secret. But I think this PR introduced the first feature, where it matter: a task signed in app but verified in qcluster.
> 
> So without manual intervention, production environments that don't have an explicit `DJANGO_SECRET` set will break. Am I right?

Yes you are completely right, My original assumption was that the or get_random_secret_key() fallback was just a convenience helper for local setups, but I see exactly how dangerous this is for a prod env. And do you want  removing the random fallback in settings.py so the application loudly fails to start if the secret key is missing ?

api/views/event.py
drona-gyawali: Yes you are right, i will mark batch as failed.

api/urls.py
drona-gyawali: thanks for suggestion, i will update the endpoint path.

api/views/utils.py
drona-gyawali: Sorry for that, I will add type hints .

greedybear/process_event.py
drona-gyawali: thanks for the feedback :), and Yes this should not be happen both src_ip and sensor are required field so this checks are dead block I will remove them.

api/views/utils.py
drona-gyawali: I agree with this design. Also, this behavior might surprise users, for example, the API returns success but the IOC count is 0. Thanks for this feedback, I will refactor this part.

api/views/event.py
drona-gyawali: The reason behind invalid_event_count is for immediate monitoring of a single user's behavior, so we can track how often they are failing to ingest events in the db.

Also, in my mind, I had a plan to implement account locking, meaning if a user reaches a certain threshold of errors, we automatically lock the account to safeguard our endpoint from broken automated scripts.

I wanted your help on this topic  Does this locking mechanism make sense? If yes, what threshold would you recommend before triggering a lock

regulartim:
> Yes you are completely right, My original assumption was that the or get_random_secret_key() fallback was just a convenience helper for local setups, but I see exactly how dangerous this is for a prod env. And do you want removing the random fallback in settings.py so the application loudly fails to start if the secret key is missing ?

We'll solve that in a separate issue/PR. I'll open an issue later that you _can_ pick up.

api/views/utils.py
regulartim: No problem! :)

api/views/event.py
regulartim: Maybe locking the APISource instead of the user is better. The user could then log in and unlock the APISource again (given they have the corresponding permission), if they fixed the underlying issue. I don't fear that users will abuse the endpoint in some way, I fear that users will write shitty code that spams the endpoint with invalid data. You know what I mean? 

I am not sure about the threshold though. We are counting event batches that are containing at least one invalid event, right?

drona-gyawali: That makes perfect sense. Locking the APISource instead of the whole user profile is a much better approach because it makes the specific API token inaccessible while still allowing the user to log in normally and fix things.

You are completely right about the counter logic too. My current implementation tracks the number of failed batch request attempts rather than individual events, as it throws an error and rejects the whole batch.  I  also think setting a threshold of 10 failed batches is resonable..

Does that sound good?

regulartim: Yes, sounds good. Maybe we can periodically reset the counter for unlocked APISources in a future PR (we need retention mechanisms anyway).

regulartim:
@drona-gyawali Can you ping me again when you implemented all changes want me to review again?

api/views/event.py
drona-gyawali: Yes I agree on periodical reset and I will raise separate issue for this.

drona-gyawali:
Hi @regulartim, I have pushed the new changes and they are ready for your review :)

api/views/event.py
```diff
@@ -0,0 +1,162 @@
...
+    except Exception:
+        logger.exception("Failed while creating batch & events")
+        return Response({"error": "An internal database error occurred while staging events"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+    if total_created == 0:
+        batch.status = "failed"
+        batch.last_error = "No valid events could be stored"
+        batch.save(update_fields=["status", "last_error"])
+        return Response({"error": "No valid events could be stored"}, status=status.HTTP_400_BAD_REQUEST)
+
```
regulartim: This branch is kind of dead code, right? The serializer enforces a minimum of 1 events. Or is there a case where `total_created == 0` can be the case? If not, I think I would remove this.

greedybear/process_event.py
```diff
@@ -0,0 +1,275 @@
...
+        for raw in raw_events:
+            hit = _normalize_raw_event_to_hit(raw)
+            if hit is not None:
+                hits.append(hit)
+            else:
+                skipped += 1
+
+        if skipped:
+            logger.warning(f"[task={task_id}] {skipped} RawEvent(s) skipped (bad data)")
+
```
regulartim: Can hit really be None? If not, please simplify.

api/views/utils.py
```diff
@@ -686,3 +687,119 @@ def create_or_get_sensor(*, api_source, validated_data):
...
+    Validates sensor ownership and bulk-inserts raw event payloads into the database.
+
+    This internal utility performs an optimized prefetch lookup of all referenced sensors to
+    ensure they exist and belong to the calling `APISource`. It validates the incoming list
+    upfront to fail fast on foreign or missing sensor identifiers, securely strips the mapping
+    fields to prevent schema mismatch unpacking errors, and stages the data into memory.
+
+    Args:
+        events_data (list[dict]): A collection of un-persisted, validated event data dictionaries.
+        batch (EventStatus): The tracking tracking batch model instance these events belong to.
```
regulartim: "tracking tracking batch" -> typo

api/views/utils.py
```diff
@@ -686,3 +687,119 @@ def create_or_get_sensor(*, api_source, validated_data):
...
+
+
+def increment_and_evaluate_lock(api_source: APISource) -> Response | None:
+    """
+    Increments the failed batch attempts for an APISource and checks if the safety
+    threshold has been reached. If exceeded, it automatically locks the source.
+
+    Returns a 403 Response if locked, otherwise returns None.
+    """
+    api_source.invalid_event_count += 1
```
regulartim: What if two concurrent requests try to increment this at the same time? Then one increment gets lost. This might be the better way of doing that: https://stackoverflow.com/questions/11316047/incrementing-values-in-django-database

regulartim (changes requested):
Nice! 👍 This is basically mergeable. Only found minor nits. 

In your wiki text, you state that status values are pending, running, success, failed, but the actual EventStatusType enum is pending, processing, completed, failed. So there's a mismatch, right?

greedybear/process_event.py
drona-gyawali: Yes, hit never return none,  i will simply that.

api/views/event.py
drona-gyawali: Yes, I will update that .

drona-gyawali:
Hi @regulartim, i have pushed the new changes and addressed the concerns you raised. thanks a lot for the feedback

api/views/event.py
```diff
@@ -0,0 +1,156 @@
...
+
+@api_view(["GET"])
+@authentication_classes([CookieTokenAuthentication])
+@permission_classes([IsAuthenticated])
+def event_status_view(request, task_id: str):
+    """
+    Retrieve the execution and processing status of a specific event batch using its task ID.
+
+    This endpoint safely exposes the processing lifecycle phase (e.g., pending, running, success, failed)
```
regulartim: Oh, here is another mismatch!  Please fix.

regulartim (changes requested):
Nice! :) The documentation mismatches are the last think you have to fix before merging.

api/views/event.py
drona-gyawali: sorry for this, I will fix it soon.

regulartim (approved):
👍

PR #1419, opened 2026-06-18
Implement NDJSON renderer for feed endpoints.Closes #1408

### LINKED ISSUE #1408: Add NDJSON responses to Feeds API
As a follow up on #1350 , we could add NDJSON as a new (streamable!) output format to the Feeds API endpoints and make it the default on the next major release.

suvani-ctrl:
# Description
Added NDJSON (Newline Delimited JSON) as a new streamable output format for the feed endpoints.
The standard JSON format, buffers the entire response payload in memory before sending it the nwer feature uses the NDJSON implementation that streams IOCs line-by-line using a StreamingHttpResponse. This feature significantly reduces memory consumption and prevents potential memory exhaustion or timeouts when handling exceptionally large feeds compared to json.

### Related issues

Closes #1408
### Type of change

# Checklist

### Formalities

- [X] I have read and understood the rules about [how to Contribute](https://github.com/GreedyBear-Project/GreedyBear/wiki/Contribute) to this project.
- [X] I chose an appropriate title for the pull request in the form: `<feature name>. Closes #999`
- [X] My branch is based on `develop`.
- [X] The pull request is for the branch `develop`.
- [X] I have reviewed and verified any LLM-generated code included in this PR.

### Docs and tests

- [X] I documented my code changes with docstrings and/or comments.
- [] I have checked if my changes affect user-facing behavior that is described in the [docs](https://github.com/GreedyBear-Project/GreedyBear/wiki). If so, I also included an update to the [wiki](https://github.com/GreedyBear-Project/GreedyBear/wiki) in the description of this PR.
- [X] Linter (`Ruff`) gave 0 errors. If you have correctly installed [pre-commit](https://github.com/GreedyBear-Project/GreedyBear/wiki/Contribute), it does these checks and adjustments on your behalf.
- [ ] All the tests gave 0 errors.

### GUI changes

# Review process


api/views/feeds.py
```diff
@@ -174,10 +171,15 @@ def render_response(self, request: Request, iocs_queryset: QuerySet) -> HttpResp
...
             request.accepted_media_type = FeedJSONRenderer.media_type
             return paginator.get_paginated_response(resp_data)
+        if requested_format == "ndjson" or request.accepted_media_type == "application/x-ndjson":
+            verbose = self.request_params.get("verbose", False)
 
-        # Hand the raw queryset to the renderer and flag it for envelope shaping.
-        # Set only here so error/exception responses (which bypass render_response)
-        # and the pre-built paginated/ASN payloads are left as plain JSON.
+            def stream_iocs():
+                feed = build_feed_dict(iocs_queryset, verbose, include_sensors=self.include_sensors)
```
regulartim, line 189: In your PR description, you write that " This feature significantly reduces memory consumption and prevents potential memory exhaustion or timeouts when handling exceptionally large feeds compared to json.". This is kind of wrong though, because here the `feed` variable accumulates all IOCs in memory. We should maybe make `build_feed_dict` return a generator (or iterator) in a follow up. What do you think?

api/views/feeds.py
```diff
@@ -174,10 +171,15 @@ def render_response(self, request: Request, iocs_queryset: QuerySet) -> HttpResp
...
+                for ioc in feed["iocs"]:
+                    yield json.dumps(ioc) + "\n"
+
+            return StreamingHttpResponse(stream_iocs(), content_type="application/x-ndjson")
```
regulartim, line 193: You don't use the `FeedNDJSONRenderer` here, right? Is that intended? Doesn't it support streaming?

api/views/feeds.py
```diff
@@ -174,10 +171,15 @@ def render_response(self, request: Request, iocs_queryset: QuerySet) -> HttpResp
...
+        if requested_format == "ndjson" or request.accepted_media_type == "application/x-ndjson":
+            verbose = self.request_params.get("verbose", False)
 
-        # Hand the raw queryset to the renderer and flag it for envelope shaping.
-        # Set only here so error/exception responses (which bypass render_response)
-        # and the pre-built paginated/ASN payloads are left as plain JSON.
+            def stream_iocs():
+                feed = build_feed_dict(iocs_queryset, verbose, include_sensors=self.include_sensors)
+                for ioc in feed["iocs"]:
+                    yield json.dumps(ioc) + "\n"
```
regulartim: `json.dumps()` might not work with `verbose==True`. Please write a test case for that.

regulartim (changes requested):
Solid first draft! 👍 
I think you forgot to update `AdvancedFeedView.get_queryset`. Please check if it needs to be updated.

api/views/feeds.py
suvani-ctrl: @regulartim  agreed, `build_feed_dict` still loads everything into memory before streaming starts. Can i open a follow-up issue to refactor it to return a generator instead  because changing the  build_feed_dict might break those files and could go out of what the pr scopes for ?
image

api/views/feeds.py
suvani-ctrl: My thought process was : bypassing `FeedNDJSONRenderer` because  DRF's renderer pipeline calls `render()` and it wold expects bytes back  which means the whole response gets built in memory before anything is sent and to actually stream, i have to return a `StreamingHttpResponse` directly from `render_response()` and skip the renderer entirely. the `FeedNDJSONRenderer` is still there for potential future use but streaming can't go through the normal DRF pipeline...

api/views/feeds.py
regulartim: Yes, good idea! :+1:

api/views/feeds.py
regulartim: Ok :)

suvani-ctrl:
Added NDJSON support as a streamable output format for the Feeds API.
The main change is in `api/views/feeds.py` within `render_response()`. When `format=ndjson` is requested, the view now bypasses DRF's standard renderer pipeline and returns a `StreamingHttpResponse`. The response streams data using a Python generator that takes one IOC per line in NDJSON format:

**json.dumps(ioc, default=str) + "\n"**

An NDJSON renderer was also added for content negotiation purposes. Its `render()` method intentionally returns `b""`; it exists solely so requests with `Accept: application/x-ndjson` are handled correctly by DRF. Thee actual streaming implementation remains in the view layer because `StreamingHttpResponse` is not compatible with DRF's renderer pipeline, which expects `render()` to return a byte string.

* Fixed `PaginatedFeedView.get_request_data()` to default to JSON only when no format is explicitly specified.
* Added `ndjson` to the sensor annotation condition in `AdvancedFeedView`.
* Added  some test cases as well

Also i added few iocs and it seems to be in NDJSON format.
image

api/renderers.py
```diff
@@ -26,6 +26,15 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
 
+class FeedNDJSONRenderer(BaseRenderer):
+    media_type = "application/x-ndjson"
+    format = "ndjson"
+    charset = "utf-8"
+
+    def render(self, data, accepted_media_type=None, renderer_context=None):
+        return b""
+
+
```
regulartim, line 38: Please add a comment, why this is a no-op and where the actual rendering happens.

api/views/feeds.py
```diff
@@ -166,6 +162,7 @@ def sort_and_slice_queryset(self, qs: QuerySet) -> QuerySet:
 
     def render_response(self, request: Request, iocs_queryset: QuerySet) -> HttpResponseBase:
         """Select the renderer for the validated format and hand it the prepared data."""
+        requested_format = self.request_params.get("format_") or self.request_params.get("format")
```
regulartim: This should be dead code. `request_params` is the serializer's validated_data, which never contains `format_` . Could you please check?

api/views/feeds.py
```diff
@@ -259,8 +261,13 @@ def should_paginate(self, request_data: dict) -> bool:
...
     def get_request_data(self, request: Request, **kwargs) -> dict:
-        """Pagination requires JSON response."""
-        return request.query_params.dict() | {"format": "json"}
+        """Pagination requires JSON response unless NDJSON is specified"""
+        query_data = request.query_params.dict()
+
+        requested_format = query_data.get("format")
+        if requested_format not in ["json", "ndjson"]:
+            query_data["format"] = "json"
+        return query_data
```
regulartim: This is very misleading, no? This code suggests that PaginatedFeedView is able to return ndjson, which it is not. That view's `should_paginate()` always returns `True` and therefore `render_response` forces  JSON. Or am I overlooking something?

api/views/feeds.py
```diff
@@ -302,6 +309,7 @@ def get_queryset(self) -> QuerySet:
                     distinct=True,
                 )
             )
+
```
regulartim, line 318: Why did you add that line? Please try to minimize diff noise.

api/views/feeds.py
```diff
@@ -45,12 +46,7 @@
 from greedybear.consts import SHARE_TOKEN_SALT
 from greedybear.models import IOC, ShareToken
 
-RENDERERS_BY_FORMAT = {
-    "json": FeedJSONRenderer,
-    "txt": FeedTextRenderer,
-    "csv": FeedCSVRenderer,
-    "stix21": Stix21Renderer,
-}
+RENDERERS_BY_FORMAT = {"json": FeedJSONRenderer, "txt": FeedTextRenderer, "csv": FeedCSVRenderer, "stix21": Stix21Renderer, "ndjson": FeedNDJSONRenderer}
```
regulartim: I liked the old style better. Having a separate line for every renderer is much easier to read.

regulartim (changes requested):
Hey @suvani-ctrl ! Apart from my comments, there is another thing that I want to discuss: NDJSON is never cached, right? And every attempt to do so would introduce a conflict with the streaming goal. I never really thought about that conflict, to be honest, but I think it is a fundamental one. What's your opinion on that?

api/views/feeds.py
suvani-ctrl: On PaginatedFeedView, should_paginate() always returns True, which means the NDJSON streaming logic inside render_response() is currently unreachable. To handle this I’ve made get_request_data() to default back to forcing standard JSON for this specific endpoint.

However, what about NDJSON on paginated endpoints ?Does it also needs pagination like cursor based pagination for ndjson..?

api/views/feeds.py
suvani-ctrl: Yes . I have now cleared it to be self.request_params.get("format")

api/renderers.py
suvani-ctrl: Noted! I have written comment as     #  FeedNDJSONRenderer exists only for DRF content negotiation, actual streaming is handled via StreamingHttpResponse in render_response().

api/views/feeds.py
suvani-ctrl: ok i eliminated that space

api/views/feeds.py
suvani-ctrl: sure ! Formatted.

api/views/feeds.py
regulartim: Nice! I think we can keep it that way. Streaming NDJSON makes sense for endpoints that transfer a lot of data. The paginated endpoints usually don't.

regulartim:
Hey @suvani-ctrl ! Before giving your PR another review, I want to hear your opinion on what I wrote earlier: 

> NDJSON is never cached, right? And every attempt to do so would introduce a conflict with the streaming goal. I never really thought about that conflict, to be honest, but I think it is a fundamental one. What's your opinion on that?

suvani-ctrl:
Hey @regulartim, after tracing through CachedResponseMixin._store_api_response(), I found that NDJSON responses are already automatically excluded from caching the isinstance(response, Response) check returns False for StreamingHttpResponse, so no cache writes happen at all. (Please correct me if i got it wrong) The fundamental reason is that caching requires a complete, replayable response object, but a generator can only be consumed once there's nothing to store.
The tradeoff could be: every NDJSON request hits the database with no cache protection. At minimal traffic  the rate limiting could provide some protection and feed data only changes on a cron cycle anyway, so the same data gets queried repeatedly with no benefit. At scale with many concurrent NDJSON requests this could become a bottleneck.
I think the right fix isn't response-level caching but data-level caching pre-serializing IOC chunks into Redis during the cron run and having the generator pull from cache instead of hitting the database. This could pair nicely with streaming since we arent trying to cache a response, but the underlying data. Cache invalidation also becomes natural since the cron job already knows when data changes.

For now could the automatic bypass be enough.? Also i think refractoring this  iocs_queryset.iterator(chunk_size=x) to cut the memory pressure could be nice .

regulartim:
Hey @suvani-ctrl ! Thanks for your assessment. Redis could be an option if we would use it in this project. But currently we do not. Could you open a follow up issue describing the necessary refactoring of `build_feed_dict`?

regulartim (approved):

suvani-ctrl:
> Hey @suvani-ctrl ! Thanks for your assessment. Redis could be an option if we would use it in this project. But currently we do not. Could you open a follow up issue describing the necessary refactoring of `build_feed_dict`?

Yes. Sure Thankyou! @regulartim

PR #1422, opened 2026-06-20
feat: introduce HoneypotPayload model with custom quarantine storage …

### LINKED ISSUE #1403: [Access payload files] Implement HoneypotPayload Model & Quarantine Storage
### Problem

GreedyBear needs a database model to track payload metadata and a secure way to store downloaded malicious binaries without risking accidental execution.

### Suggested approach

- Create the `HoneypotPayload` Django model with fields for hashes, file type, source honeypot, and relationships to `Ioc` and `CowrieSession`.
- Implement a custom `QuarantineStorage` backend (subclassing `FileSystemStorage`) to restrict file saves to a `media/quarantine/` directory and force a `.vir` extension.
- Register the model in the Django admin and write unit tests.

### Additional context

Part of Phase 2 of the GSoC 2026 `Access payload files` project.

opbot-xd:
# Description

Implemented the `HoneypotPayload` Django model to track payload metadata (`md5`, `sha1`, `sha256`, `mime_type`, `size`, `locator`, etc.) and its relationships to `Honeypot`, `IOC`, and `CowrieSession`. Also implemented a custom `QuarantineStorage` backend that inherits from `FileSystemStorage` to securely restrict downloaded malicious payloads to the quarantine directory (`/var/lib/greedybear/quarantine`) with enforced `.vir` extensions to prevent accidental execution. Registered the model in the Django admin dashboard and added unit tests for both the model and the custom storage backend.

### Related issues

Closes #1403

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes


opbot-xd:
Hi @regulartim, the PR is ready for review. Please take a look when you get time... :D

greedybear/models.py
```diff
@@ -481,3 +482,24 @@ class Meta:
...
+
+class HoneypotPayload(models.Model):
+    """
+    Model for the payload that would be received by the Tpot Payload server to quarantine it in the Quarantine directory.
+    """
+
+    payload_file = models.FileField(storage=QuarantineStorage(), blank=True, null=True)
+    md5 = models.CharField(max_length=32, blank=True, default="")
+    sha1 = models.CharField(max_length=40, blank=True, default="")
+    sha256 = models.CharField(max_length=64, blank=True, default="")
```
regulartim: What do you think of adding a unique constraint to `sha256`?

greedybear/settings.py
```diff
@@ -19,6 +19,7 @@
 DEBUG = os.environ.get("DEBUG", "False") == "True"
 
 DJANGO_LOG_DIRECTORY = "/var/log/greedybear/django"
+QUARANTINE_DIR = "/var/lib/greedybear/quarantine"
```
regulartim, line 22: This directory is never provisioned (in Dockerfile , default.yml and / or entrypoint). Do you plan to do that in an upcoming PR?

regulartim (changes requested):
Looks good all in all! 👍

greedybear/models.py
opbot-xd: This seems fine just a small catch; We are linking the payload to `ioc` and `cowrie_session` as foreign key. Unique `sha256` means we can only link a payload to a single session/IOC (presumably the 'first seen' one). If another session drops the exact same payload, we would either have to skip creating a new record or we'd get an Error. Should we consider changing `ioc` and `cowrie_session` to ManyToManyField so we can link a single unique payload to all the different sessions that drop it?

greedybear/settings.py
opbot-xd: Good catch! I totally forgot to do that. I created a `quarantine_data` volume in `default.yml` and updated the `Dockerfile` and entrypoint scripts to handle the directory creation and www-data permissions.

greedybear/models.py
regulartim: > Should we consider changing ioc and cowrie_session to ManyToManyField so we can link a single unique payload to all the different sessions that drop it?

I think that would be cleaner, yes. Or do you see any problems? 

And it's semantically more correct: two different IOCs can download the exact same payload and a single IOC can download two different payloads, so it is a N:M relation. I guess `source_honeypot` is N:M too?

opbot-xd: yes, source_honeypot should definitely be N:M too, since the exact same payload binary could theoretically be captured by both Cowrie and Log4pot independently.

greedybear/models.py
```diff
@@ -481,3 +482,24 @@ class Meta:
...
+
+class HoneypotPayload(models.Model):
+    """
+    Model for the payload that would be received by the Tpot Payload server to quarantine it in the Quarantine directory.
+    """
+
+    payload_file = models.FileField(storage=QuarantineStorage(), blank=True, null=True)
+    md5 = models.CharField(max_length=32, blank=True, default="")
+    sha1 = models.CharField(max_length=40, blank=True, default="")
+    sha256 = models.CharField(max_length=64, blank=True, default="", unique=True)
```
regulartim: Now that `sha256` has a unique constraint, a default value does not make sense anymore. The second time a HoneypotPayload object without a `sha256` value is created, the code will throw an Integrity error. 

Another side effect is that we now always need to get a sha256 value from the T-Pot side. But that should not be a problem, right?

If it can legitimately be absent, we could allow `null=True`, but I think it's a good idea to require the hash to prevent duplication.

regulartim (changes requested):
Nice, basically ready to get merged! 👍 There's one last thing I found.

greedybear/models.py
opbot-xd: Yes, since we will always get the sha256 value from the T-Pot side, this shouldn't cause any issues!

regulartim (approved):
Nice! :)

PR #1435, opened 2026-06-29
Implement Django Q2 Task for Payload Extraction. Closes #1404

### LINKED ISSUE #1404: [Access payload files] Implement Django Q2 Task for Payload Extraction
### Problem

GreedyBear needs a mechanism to routinely fetch new payloads from the `tpot-payload-server`, deduplicate them, and save them locally without blowing up disk space.

### Suggested approach

- Implement a Django Q2 scheduled task `extract_honeypot_payloads` to run periodically.
- Deduplicate by checking if the SHA256 hash already exists in PostgreSQL before initiating any downloads.
- Enforce a `MAX_QUARANTINE_SIZE_GB` setting to cap local disk usage for payloads.

### Additional context

Part of Phase 2 of the GSoC 2026 `Access payload files` project.

opbot-xd:
# Description

This PR implements the `PayloadExtractionJob` cronjob, which periodically fetches recently modified payloads from the `tpot-payload-server`, deduplicates them using SHA256 hashes against PostgreSQL, checks local quarantine disk usage against `MAX_QUARANTINE_SIZE_GB`, and downloads/stores them safely in the quarantine directory using `QuarantineStorage`.

Key components implemented:
- **`PayloadExtractionJob`**: Cronjob class under `greedybear/cronjobs/payload_extraction.py` implementing the fetching, deduplication, disk quota enforcement, and download logic.
- **Integration**: Added `extract_honeypot_payloads` task in `greedybear/tasks.py` and integrated it to run after every extraction cycle inside `extract_all()`.
- **Settings**: Introduced `TPOT_PAYLOAD_SERVER_URL`, `TPOT_PAYLOAD_SERVER_API_KEY`, and `MAX_QUARANTINE_SIZE_GB` configuration variables in `greedybear/settings.py`.
- **Environment variables**: Added documentation and variables to `docker/env_file_template`.
- **Test coverage**: Created `tests/test_payload_extraction.py` achieving 98.89% coverage of the payload extraction logic, and updated existing wrapper tests in `tests/test_tasks.py`.

### Related issues

Closes #1404

### Type of change

# Checklist

### Formalities

### Docs and tests

- [x] I have checked if my changes affect user-facing behavior that is described in the [docs](https://github.com/GreedyBear-Project/GreedyBear/wiki). (N/A)

### GUI changes

Ignore this section as no GUI changes were made.

### DIFF
--- a/docker/env_file_template
+++ b/docker/env_file_template
@@ -106,7 +106,7 @@ FEEDS_LICENSE=
 
 # T-Pot Payload Server URL for downloading honeypot payloads.
 # If not set, payload extraction is disabled.
-# Example: https://tpot.example.com
+# Example: https://tpot.example.com:64299
 TPOT_PAYLOAD_SERVER_URL=
 
 # Optional API key for authenticating with the T-Pot Payload Server.
--- a/docker/env_file_template
+++ b/docker/env_file_template
@@ -104,6 +104,19 @@ TRENDING_BUCKET_RETENTION_HOURS=744
 # Example: https://github.com/honeynet/GreedyBear/blob/main/FEEDS_LICENSE.md
 FEEDS_LICENSE=
 
+# T-Pot Payload Server URL for downloading honeypot payloads.
+# If not set, payload extraction is disabled.
+# Example: https://tpot.example.com
+TPOT_PAYLOAD_SERVER_URL=
+
+# Optional API key for authenticating with the T-Pot Payload Server.
+# When empty, authentication is disabled on both sides.
+TPOT_PAYLOAD_SERVER_API_KEY=
+
+# Maximum disk usage (in GB) for the quarantine directory.
+# Payload downloads are paused when this limit is reached.
+MAX_QUARANTINE_SIZE_GB=5
+
 # Optional IntelOwl base URL. When set, a link to analyze each IOC on IntelOwl
 # will appear in the Feeds table.
 # Example: https://your-intelowl-instance.example.com
--- a/greedybear/cronjobs/payload_extraction.py
+++ b/greedybear/cronjobs/payload_extraction.py
@@ -1,8 +1,8 @@
 # This file is a part of GreedyBear https://github.com/honeynet/GreedyBear
 # See the file 'LICENSE' for copying permission.
 
-import shutil
 import time
+from pathlib import Path
 
 import requests
 from django.conf import settings
@@ -128,15 +128,12 @@ class PayloadExtractionJob(Cronjob):
 
     def _quarantine_usage_bytes(self):
         """
-        Return the current disk usage of the quarantine directory in bytes.
+        Return the total size of files in the quarantine directory in bytes.
         """
-        quarantine_dir = settings.QUARANTINE_DIR
-        try:
-            usage = shutil.disk_usage(quarantine_dir)
-        except FileNotFoundError:
+        quarantine_path = Path(settings.QUARANTINE_DIR)
+        if not quarantine_path.is_dir():
             return 0
-        else:
-            return usage.used
+        return sum(f.stat().st_size for f in quarantine_path.iterdir() if f.is_file())
 
     def _download_and_store(self, client, server_url, payload_meta):
         """
--- a/greedybear/cronjobs/payload_extraction.py
+++ b/greedybear/cronjobs/payload_extraction.py
@@ -1,0 +1,188 @@
+# This file is a part of GreedyBear https://github.com/honeynet/GreedyBear
+# See the file 'LICENSE' for copying permission.
+
+import shutil
+import time
+
+import requests
+from django.conf import settings
+from django.core.files.base import ContentFile
+
+from greedybear.cronjobs.base import Cronjob
+from greedybear.cronjobs.http_client import HttpClient
+from greedybear.models import HoneypotPayload
+
+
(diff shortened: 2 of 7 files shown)


opbot-xd:
Hi @regulartim,

I ran into a CI issue: 4–5 new tests in `test_payload_extraction.py` call `.save()` on `payload_file`, which uses `QuarantineStorage()`. Since that defaults to `QUARANTINE_DIR = "/var/lib/greedybear/quarantine"`, Django tries to create that path on the GitHub runner and fails with `PermissionError` (runner isn’t in Docker and can’t write to `/var/lib/...`).

Two possible fixes:

1. create the directory before tests:

```yaml
custom_command: sudo mkdir -p /var/lib/greedybear/quarantine && sudo chown -R $USER:$USER /var/lib/greedybear && python manage.py test --tag=migration --failfast
```

2. use `/tmp` in CI/tests:

```python
# in settings.py
import sys
if STAGE_CI or (len(sys.argv) > 1 and sys.argv[1] == "test"):
    QUARANTINE_DIR = "/tmp/greedybear/quarantine"
```

Do you prefer one of these, or is there another storage-testing pattern?

regulartim:
Hey @opbot-xd ! 

> Do you prefer one of these, or is there another storage-testing pattern?

I think like the 2nd one better. Feel free to implement it like this.

opbot-xd:
Hi @regulartim, the PR is ready for review please take a look when you get time.

Also added S108 to the per-file-ignores for `greedybear/settings.py` in `pyproject.toml`. This suppresses ruff's "insecure temp directory" warning, the `/tmp/greedybear/quarantine` path is only used when STAGE_CI is true or manage.py test is running.

opbot-xd:
### Test Coverage Report

```text
Name                                        Stmts   Miss   Cover   Missing
--------------------------------------------------------------------------
greedybear/cronjobs/payload_extraction.py      88      1  98.86%   136
--------------------------------------------------------------------------
TOTAL                                          88      1  98.86%
```

docker/env_file_template
```diff
@@ -104,6 +104,19 @@ TRENDING_BUCKET_RETENTION_HOURS=744
 # Example: https://github.com/honeynet/GreedyBear/blob/main/FEEDS_LICENSE.md
 FEEDS_LICENSE=
 
+# T-Pot Payload Server URL for downloading honeypot payloads.
+# If not set, payload extraction is disabled.
+# Example: https://tpot.example.com
```
regulartim: Isn't the example a bit misleading? You would also need to set the correct prot on the URL, as the payload server usually does not listen on 443, right?

greedybear/cronjobs/payload_extraction.py
```diff
@@ -0,0 +1,185 @@
...
+            payloads: List of payload metadata dicts (must contain 'sha256' key).
+
+        Returns:
+            list[dict]: Only the payloads not yet stored locally.
+        """
+        incoming_hashes = {p["sha256"] for p in payloads if "sha256" in p}
+        existing_hashes = set(HoneypotPayload.objects.filter(sha256__in=incoming_hashes).values_list("sha256", flat=True))
+        new_hashes = incoming_hashes - existing_hashes
+        self.log.debug(f"Deduplication: {len(incoming_hashes)} incoming, {len(existing_hashes)} existing, {len(new_hashes)} new.")
+        return [p for p in payloads if p.get("sha256") in new_hashes]
```
regulartim: What if `payloads` contains two identical payloads? This would lead to a integrity error from the DB, or am I mistaking? If so, either handle duplicates in `payloads` or handle the following IntegrityError.

greedybear/cronjobs/payload_extraction.py
regulartim, line 1: Could you please add type hints to the function / method definitions in this file?

greedybear/cronjobs/payload_extraction.py
regulartim: Also, you are assuming that `payloads` is a iterable of payloads here. Is there a situation where this assumption does not hold? For example if the payload server returns a non-200 response?

regulartim (changes requested):
Hey @opbot-xd ! Really nice work! I have some robustness concerns but other than that, this is mergeable in my opinion. 

@mlodic Since you originally requested this feature, do you also want to take a look?

docker/env_file_template
opbot-xd: Done. Updated the example to https://tpot.example.com:64299

greedybear/cronjobs/payload_extraction.py
opbot-xd: > Also, you are assuming that `payloads` is a iterable of payloads here. Is there a situation where this assumption does not hold? For example if the payload server returns a non-200 response?

You're right, the server could return a non-200 response body (e.g. an error dict). Added a isinstance(data, list) check in _fetch_metadata



> What if `payloads` contains two identical payloads? This would lead to a integrity error from the DB, or am I mistaking? If so, either handle duplicates in `payloads` or handle the following IntegrityError.

Copilot flagged this too, I initially skimmed past it, but now I am realising, this is a real scenario since the same payload could appear across multiple honeypot directories with different locators. Fixed `_deduplicate` now tracks seen hashes and keeps only the first occurrence of each sha256

regulartim (approved):
Nice, thanks! :)

Issue #962, opened 2026-03-05
Add OpenAPI/Swagger schema documentation for API endpoints


manik3160:
## Description

While going through the codebase to understand the API layer (in preparation for GSoC Idea— Event Collector API), I noticed that GreedyBear has a rich set of API endpoints but lacks formal OpenAPI/Swagger documentation.

It looks like `drf-spectacular` is already present in our dependency tree (pulled in via `certego-saas`), but we aren't currently configuring it or exposing the schema routing in [urls.py](cci:7://file:///Users/manik/GreedyBear/api/urls.py:0:0-0:0). 

Currently, an external developer integrating with GreedyBear needs to read the source code of `api/views/` and `api/serializers.py` to understand the request/response contracts for endpoints like:
- `/api/feeds/advanced/`
- `/api/enrichment`
- `/api/cowrie_session`

## Benefits

Exposing the OpenAPI documentation would:
- Auto-generate interactive API docs (Swagger UI / Redoc) out of the box
- Help external consumers understand the API contract instantly without reading source code

## Proposed Solution

Wire up `drf-spectacular` which is already available in the environment. 

1. Configure the `DEFAULT_SCHEMA_CLASS` in `greedybear/settings.py`:
```python
REST_FRAMEWORK = {
    # ... existing settings
    "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema",
}

SPECTACULAR_SETTINGS = {
    "TITLE": "GreedyBear API",
    "DESCRIPTION": "Threat Intel Platform for T-POTs",
    "VERSION": "1.0.0",
}
```
2. Expose the schema and Swagger UI routes in `api/urls.py`:
```python
from drf_spectacular.views import SpectacularAPIView, SpectacularSwaggerView

urlpatterns = [
    # ... existing routes
    path("schema/", SpectacularAPIView.as_view(), name="schema"),
    path("docs/", SpectacularSwaggerView.as_view(url_name="schema"), name="swagger-ui"),
]

manik3160:
Hey @regulartim, if this approach makes sense, I’d be happy to work on implementing it.

regulartim:
Hey @manik3160 ! We already have a documentation for our API endpoints in our [official docs](https://intelowlproject.github.io/docs/GreedyBear), which is (mostly) auto generated. What would be the main advantage of doing it the way you suggest?

manik3160:
[...]

PR #1491, opened 2026-08-04
[Dashboard modularization] Backend persistence for react-grid-layout. Closes #1483

### LINKED ISSUE #1483: [Dashboard Modularization] Backend persistence for dashboard configuration
Deliverable 5 for the Dashboard Modularization project.
Currently the config for the dashboard is stored in localstorage only.

### Proposed Changes: 
- Add `DashboardConfig` model. Stores a single global JSON blob.
- Create a migration for the new model.
- New DashboardConfigView (APIView): GET, PUT (superuser only), DELETE (superuser only)

This would replace the current implementation localstorage for the config.

armoredvortex:
# Description
Added backend persistence for the dashboard configuration. 
A `DashboardConfig` Django model stores the layout as a global JSON blob, exposed via a DRF endpoint (GET/PUT/DELETE /api/dashboard-config/).

### Related issues
#1483 

### Type of change

# Checklist
### Formalities

### Docs and tests

### GUI changes


api/views/dashboard_config.py
```diff
@@ -0,0 +1,108 @@
...
+            request.user,
+            request.user.is_superuser,
+        )
+        record = DashboardConfig.objects.first()
+        if record is None:
+            # No config saved yet - signal the frontend to use built-in defaults.
+            return Response({"layout": None}, status=status.HTTP_200_OK)
+
+        return Response({"layout": record.layout}, status=status.HTTP_200_OK)
+
```
regulartim, line 90: This endpoint is authenticated. So an anonymous visitor would fail to call it. What happens then? I guess they see the default config, but not the config crafted by superusers. I think we should make the  GET request `AllowAny`. What do you think?

api/views/dashboard_config.py
```diff
@@ -0,0 +1,108 @@
...
+                {"detail": "Request body must contain a 'layout' key."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        if not isinstance(layout, dict) or "widgetConfigs" not in layout or "layouts" not in layout:
+            return Response(
+                {"detail": ("'layout' must be an object with 'widgetConfigs' and 'layouts' keys.")},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
```
regulartim: The validation you are doing here should be done by a serializer. We are using these on almost every public endpoint and especially when user data is written, it is the safer approach.

api/views/dashboard_config.py
```diff
@@ -0,0 +1,108 @@
...
+        if not request.user.is_superuser:
+            logger.warning(
+                "DashboardConfig PUT rejected: user=%s is not a superuser",
+                request.user,
+            )
+            return Response(
+                {"detail": "Only superusers may modify the dashboard configuration."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
```
regulartim: Normally this should be handled by a permission class, but I think there is no fitting one here. Suggestion: we make a custom permission class for this endpoint, somethink like `IsSuperUserOrReadOnly`, such that anyone can read, but only superusers can modify. 
https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions

api/views/dashboard_config.py
```diff
@@ -0,0 +1,108 @@
...
+class DashboardConfigView(APIView):
+    """
+    GET  /api/dashboard-config/
+        Returns the globally saved dashboard layout, or null when no record
+        exists yet (frontend falls back to defaultDashboardConfig.js).
+        Requires authentication so anonymous scraping is not possible.
+
+    PUT  /api/dashboard-config/
+        Replaces the global layout.  Restricted to superusers.
+        Expects JSON body: { "widgetConfigs": [...], "layouts": {...} }
```
regulartim: What about DELETE? The proper way of documentation would be the OpenAPI syle, introduced in #962 . But I leave it up to you if you want to to that.

regulartim (changes requested):
Hey @armoredvortex ! This might be the first time reviewing python code from you! :D Nice job, architecturally sound. 

Please don't push merge migrations. The clean way to do it is  to rebase on develop and regenerate your migration as a single `0061_dashboardconfig`, dropping the merge file.

Aslo, it would be cool to have the new model in the Django admin interface. For this, you have to register it in registered in `greedybear/admin.py`.

api/views/dashboard_config.py
armoredvortex: yes, fixed the bug.

api/views/dashboard_config.py
```diff
@@ -0,0 +1,121 @@
...
+
+from api.permissions import IsSuperuserOrReadOnly
+from greedybear.models import DashboardConfig
+
+logger = logging.getLogger(__name__)
+
+
+class DashboardLayoutSerializer(serializers.Serializer):
+    layout = serializers.DictField(
+        allow_null=True,
```
regulartim: Why `allow_null=True`? I think that could raise an integrty error, as you don't allow null values in the model, if someone fires a PUT request with `"layout": null` . I'll write a test case for that and we'll see what happens. :D

greedybear/models.py
```diff
@@ -518,3 +518,34 @@ class Meta:
...
+
+
+class DashboardConfig(models.Model):
+    """
+    Stores a single global JSON blob representing the dashboard layout shared
+    across all users. Only one row should ever exist (singleton pattern).
+
+    Superusers write it via the PUT /api/dashboard-config/ endpoint.
+    Any authenticated (or anonymous, depending on deployment policy) user
+    reads it via GET /api/dashboard-config/.
```
regulartim: This is kind of wrong now, isn't it?

regulartim (changes requested):
Very nice! 👍 Only details left. 

I'll push a test, as said in the comments. If it fails, please provide a fix. 

Also, I just found out that you can prevent users from adding objects in django admin. I tested it, and it works. I will also push that change, as it prevents admins from adding a 2nd dashboard config manually.

api/views/dashboard_config.py
regulartim: Ok, the test fails. Two options: (1) drop `allow_null=True` from the serializer or (2) change the model to allow null values and adapt the test to expect a 200. I'd prefer (1), but it's up to you!

armoredvortex: yes (1) seems good

regulartim (approved):
Thanks! :)

PR #1552, opened 2026-09-05
feat: Link HoneypotPayload to Cowrie sessions and attacker IOCs by SHA256. Closes #1509

### LINKED ISSUE #1509: Feature Request: IOC correlation with attacker IPs and Cowrie sessions
### Problem / What it aims to solve

Currently, malware payloads are successfully downloaded and quarantined, but they lack automatic correlation with the broader attack context. Analysts need to manually connect a downloaded payload to the attacker IP or specific Cowrie session that delivered it.

### Suggested approach

Automatically link downloaded payloads to the attacker IPs and Cowrie sessions by cross-referencing payload hashes against T-Pot's structured event logs. This could be implemented by extending the existing ingestion pipeline to query T-Pot's Elasticsearch stack for matching hashes and updating the relationships on the `HoneypotPayload` model.

### Additional context

**Note:** Implementing this feature may require changes across multiple codebases (both `GreedyBear` and `tpot-payload-server`).

kikiscodedeliveryservice:
# Description

Adds the join GreedyBear was missing between downloaded payloads and the Cowrie sessions/attacker IOCs that delivered them. `HoneypotPayload.cowrie_sessions`/`.iocs` existed but were not being populated. This join occurs bi-directionally, in both `PayloadExtractionJob._download_and_store` and `CowrieExtractionStrategy._process_session_hit`, so it works regardless of whether the session or the payload shows up first in a given extraction window.

**_IMPORTANT NOTE_**: This adds one change beyond the strict scope of issue #1509. It exposes both fields on the payloads API (`iocs`, `cowrie_sessions`). I have added it since populated fields that are not returned in the corresponding APIs for the appropriate users would be incomplete development on this feature. Please refer to the Wiki Update section.

### Related issues

Closes #1509

### Type of change

# Checklist

### Formalities

### Docs and tests

### Wiki update

The Payloads API response gains two new fields, not yet reflected in the Usage page:

`iocs` (list[str]): Attacker IPs (or other IOCs) linked to this payload.
`cowrie_sessions` (list[str]): IDs of the Cowrie sessions that transferred this payload, as hex strings.

### DIFF
--- a/greedybear/cronjobs/repositories/payload.py
+++ b/greedybear/cronjobs/repositories/payload.py
@@ -1,5 +1,7 @@
 import logging
 
+from django.db.models.functions import Lower
+
 from greedybear.models import CowrieSession, HoneypotPayload
 
 
@@ -29,7 +31,9 @@ class PayloadRepository:
         Returns:
             Number of sessions linked.
         """
-        sessions = CowrieSession.objects.filter(file_transfers__shasum__iexact=payload.sha256).select_related("source").distinct()
+        # assumption: Cowrie always emits lowercase hex for shasum (hashlib.hexdigest()), and payload.sha256 is
+        # always stored in lowercase too, hence this plain comparison works with CowrieFileTransfer's shasum index.
+        sessions = CowrieSession.objects.filter(file_transfers__shasum=payload.sha256.lower()).select_related("source")
         linked = 0
         for session in sessions:
             payload.cowrie_sessions.add(session)
@@ -49,7 +53,7 @@ class PayloadRepository:
         Returns:
             True if a matching payload was found and linked.
         """
-        payload = HoneypotPayload.objects.filter(sha256__iexact=shasum).first()
+        payload = HoneypotPayload.objects.annotate(sha256_lower=Lower("sha256")).filter(sha256_lower=shasum.lower()).first()
         if payload is None:
             return False
         payload.cowrie_sessions.add(session)
--- a/greedybear/cronjobs/repositories/payload.py
+++ b/greedybear/cronjobs/repositories/payload.py
@@ -1,0 +1,57 @@
+import logging
+
+from greedybear.models import CowrieSession, HoneypotPayload
+
+
+class PayloadRepository:
+    """
+    Repository joining HoneypotPayload records with the CowrieSession(s) that
+    transferred the same file, matched by SHA256 hash.
+
+    ExtractionJob (Cowrie) and PayloadExtractionJob run independently and can
+    complete in either order within the same tick, so this repository is used
+    from both directions: once a payload is downloaded, and once a Cowrie file
+    transfer is recorded, each side looks for a match already stored by the
+    other.
+    """
+
+    def __init__(self):
+        self.log = logging.getLogger(f"{__name__}.{self.__class__.__name__}")
+
+    def link_sessions_to_payload(self, payload: HoneypotPayload) -> int:
+        """
+        Find Cowrie sessions that transferred a file matching this payload's
+        SHA256, and link them (and their source IOCs) to the payload.
+
+        Args:
+            payload: HoneypotPayload instance to link, already saved.
+
+        Returns:
+            Number of sessions linked.
+        """
+        sessions = CowrieSession.objects.filter(file_transfers__shasum__iexact=payload.sha256).select_related("source").distinct()
+        linked = 0
+        for session in sessions:
+            payload.cowrie_sessions.add(session)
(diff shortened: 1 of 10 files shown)


greedybear/cronjobs/repositories/payload.py
```diff
@@ -0,0 +1,57 @@
...
+        it (and the session's source IOC) to the session.
+
+        Args:
+            session: CowrieSession instance that transferred the file.
+            shasum: SHA256 checksum of the transferred file.
+
+        Returns:
+            True if a matching payload was found and linked.
+        """
+        payload = HoneypotPayload.objects.filter(sha256__iexact=shasum).first()
```
regulartim: I am not sure if the `iexact` filter is a good choice here. If you take a look at the model definition, it has a unique constraint on its `sha256` field that uses lower case. This constraint implies an index on that field and we should try to use that index to answer the query. Maybe you can use a `Lower()`-annotated lookup here? Something like this: 

```
    payload = HoneypotPayload.objects
    .annotate(sha256_lower=Lower("sha256"))
    .filter(sha256_lower__in=shasum)
    .first()
```

I am not a 100% sure if this really. What do you think?

greedybear/cronjobs/repositories/payload.py
```diff
@@ -0,0 +1,57 @@
...
+        Find Cowrie sessions that transferred a file matching this payload's
+        SHA256, and link them (and their source IOCs) to the payload.
+
+        Args:
+            payload: HoneypotPayload instance to link, already saved.
+
+        Returns:
+            Number of sessions linked.
+        """
+        sessions = CowrieSession.objects.filter(file_transfers__shasum__iexact=payload.sha256).select_related("source").distinct()
```
regulartim: Same here as in line 52, but it's a little more complicated. There is an index on `shasum` but it is case insensitive. As fat as I know, Cowrie always emits lowercase hex, so we could just use `exact=payload.sha256.lower()`. Another option would be to add a corresponding index to CowrieFileTransfer. What do you think? 

Note that on live instance, there are million of cowrie sessions. So using `iexact` can cause really bad performance issues.

regulartim (changes requested):
Good job, thank you! :) 

Since you introduced a repository for the payloads, do you think it makes sense to add CRUD methods to the repository class and also use them in other places?

greedybear/cronjobs/repositories/payload.py
kikiscodedeliveryservice: Yes, you're right about this. The indexing was constructed with lower cased values and would not be utilised by `iexact` as it internally upper cases it for comparison. Your fix would enforce that the index is utilised, I will implement it as is (except for the `__in` which I'm assuming was a typo because shasum is a single string and not a list).

greedybear/cronjobs/repositories/payload.py
kikiscodedeliveryservice: Yes, similar issue with the indexing as a plain, case sensitive index cannot be used for the built-in upper cased comparisons of `iexact`. I also verified that Cowrie's shasum is computed using Python's `hexdigest()` which is unconditionally lowercase.

I think we should avoid adding another index, it might be overkill. Currently, `CowrieFileTransfer` is created only from one path and the shasum is taken verbatim from Cowrie's own log data. It is unlikely that such longstanding behaviour would change. Moreover, as you mentioned, in production instances there can be millions of records which would cause two issues: 1) increase the DB write costs, especially bulk writes could potentially slow down noticeably and 2) greater the number of records, the longer the DB would be locked for applying the index's migration which may impact the reliability of the live instance and user experience. 

The `CowrieFileTransfer` model already has 2 shasum-related indices as opposed to `HoneypotPayload` which has 1. It would be better to not add indices unless there is a performance bottleneck that simply cannot be resolved any other way. Here, we can do the lowercase comparison similar to line 52, and mention our assumption clearly in a comment in case this comes up again in future development. We can always add the index later if it becomes absolutely necessary.

However, I do understand that this would make us too reliant on Cowrie behaving properly, and if that is not aligned with the way GreedyBear is typically designed, I can add the index and generate the migrations file.

greedybear/cronjobs/repositories/payload.py
regulartim: Yes, the `__in` was a mistake by me, sorry.

greedybear/cronjobs/repositories/payload.py
regulartim: I agree, let's not add another index.

regulartim (approved):
Good job, thank you!

Issue #1530, opened 2026-09-01
`PayloadExtractionJob._deduplicate` skips files that were first seen as hash-only stubs


opbot-xd:
### Context

This was first raised by @kikiscodedeliveryservice in [Discussion #1514](https://github.com/GreedyBear-Project/GreedyBear/discussions/1514#discussioncomment-18208753)

### Description

`PayloadExtractionJob._deduplicate` filters out any payload whose `sha256` already has a row in the `HoneypotPayload` table, regardless of whether that row has an actual file attached or is just a metadata-only stub (`payload_file=None`).

Hash-only stubs are created by `_process_payload_hashes` (introduced in PR #1463) when a raw event reports a payload hash before the file is available from the `tpot-payload-server`. The two PRs were written at different times without accounting for each other's assumptions:

- PR #1435 (`_deduplicate`) assumed: *any row with this hash = file already downloaded.*
- PR #1463 (`_process_payload_hashes`) introduced rows where `payload_file=None`.

This means once a hash-only stub exists, the actual file can never be automatically downloaded by the pipeline.

### Expected behavior

`_deduplicate` should only skip hashes whose `HoneypotPayload` row already has a non-null `payload_file`. Stub rows (`payload_file=None`) should pass through the filter so the file can still be downloaded.

Additionally, `_download_and_store` must use `get_or_create` instead of blindly constructing a new object, to avoid an `IntegrityError` from the unique constraint on `sha256` when updating an existing stub.

kikiscodedeliveryservice:
#1509 (PR #1552) already touches `PayloadExtractionJob`, the `_deduplicate` filter fix and the `_download_and_store` fix would be small, related changes in the same file. I can take this issue up in the same PR if that's okay. If not, I can make a separate PR for it.

@opbot-xd @regulartim, let me know what you think.

regulartim:
Hey @kikiscodedeliveryservice ! I'd like to keep these changes separate. When your PR is merged, I am happy to assign you here.

kikiscodedeliveryservice:
@regulartim Since my PR is merged, please assign me here next.

kikiscodedeliveryservice:
Hey, @regulartim! Raised a [PR](https://github.com/GreedyBear-Project/GreedyBear/pull/1593).

PR #1565, opened 2026-09-08
fix: normalize SHA256 case in payload extraction. Closes #1556

### LINKED ISSUE #1556: Latent IntegrityError in payload extraction
In the `_download_and_store` method, `payload_meta["sha256"]` is used raw but the underlying data model has a constraint: `UniqueConstraint(Lower("sha256"))`. An uppercase hash from the server might miss the case-sensitive sha256 lookup in `_deduplicate`, reach the insert and trigger the unique constraint. This would throw an exception and abort the whole run.

aashu2006:
# Description

HoneypotPayload is unique on Lower("sha256"), but the dedup check looked up hashes case-sensitively and stored them exactly as the payload server sent them. So an uppercase hash for a payload already saved in lower case slipped past dedup, hit the insert, and broke the constraint.

That kills more than one payload. The save has no error handling and the cronjob re-raises, so one bad hash ends the whole run and the rest of the batch is dropped.

Fix normalizes both sides: incoming hashes get lower-cased, and existing rows are matched on Lower("sha256"). That second bit is a little beyond the issue, but without it any mixed-case row already in the DB would still slip through.

kept this to case normalization only, as agreed in #1556, the get_or_create change stays with #1530

### Related issues

Closes #1556

Related: #1530 (get_or_create is left out on purpose), #1552 (also touches payload_extraction.py)

### Type of change

### Formalities

### Docs and tests

### Tests

5 new tests. No case-sensitivity coverage existed before.

I checked they are real regression tests by reverting the fix and rerunning themwith the tests untouched, all four behavioural ones fail, including the 'CCCC…' != 'cccc…' assertion.

### DIFF
--- a/greedybear/cronjobs/payload_extraction.py
+++ b/greedybear/cronjobs/payload_extraction.py
@@ -7,10 +7,31 @@ from pathlib import Path
 import requests
 from django.conf import settings
 from django.core.files.base import ContentFile
+from django.db.models.functions import Lower
 
 from greedybear.cronjobs.base import Cronjob
 from greedybear.cronjobs.http_client import HttpClient
 from greedybear.models import HoneypotPayload
+
+
+def normalize_sha256(value: str) -> str:
+    """
+    Lower-case a SHA256 hash so it matches the way hashes are stored.
+
+    HoneypotPayload enforces uniqueness with UniqueConstraint(Lower("sha256")),
+    and the event-injection path already lower-cases before writing (see
+    greedybear/process_event.py::_process_payload_hashes). Normalizing the
+    payload-server path too keeps both writers consistent, so a hash that
+    differs only in case is recognised as a duplicate instead of reaching the
+    insert and violating the constraint.
+
+    Args:
+        value: SHA256 hash as received from the payload server.
+
+    Returns:
+        The hash in lower case.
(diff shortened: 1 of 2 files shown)


greedybear/cronjobs/payload_extraction.py
```diff
@@ -7,12 +7,33 @@
...
+
+    Args:
+        value: SHA256 hash as received from the payload server.
+
+    Returns:
+        The hash in lower case.
+    """
+    return value.lower()
+
+
```
regulartim: I appreciate the effort, but I think a 17-line wrapper around `str.lower()` is a bit much. Please just make an inline `.lower()`call in the methods.

regulartim (changes requested):
Good job, your fix is correct, minimal and well-tested. Only one nit.

greedybear/cronjobs/payload_extraction.py
aashu2006: Done, I have made inline `.lower()` in both methods and dropped the wrapper, kept a short comment in `_deduplicate` explaining the Lower("sha256") constraint.

regulartim (approved):
Nice, thanks!

Issue #1555, opened 2026-09-07
No notification when payload quota is reached


regulartim:
During payload extraction, when the defined storage quota is reached, the job logs this at level WARNING. I think we should increase the log level to ERROR, such that admins get notified via Slack or ntfy.

joshuavetos:
Hi! I’d like to work on this issue if it’s available. I’ll keep the change focused on making the quota condition log at the level needed for Slack/ntfy notifications, and I’ll add or update tests if there’s existing coverage around the extraction path.

regulartim:
Hey @joshuavetos ! alright, I'll assign you!

PR #1570, opened 2026-09-08
Keep metadata for payloads skipped at quarantine quota. Closes #1532

### LINKED ISSUE #1532: Payloads skipped due to quarantine quota are permanently lost (no stub saved, no backfill)
This was first raised by @kikiscodedeliveryservice in [Discussion #1514](https://github.com/GreedyBear-Project/GreedyBear/discussions/1514#discussioncomment-18208753)

### Description

When the quarantine directory reaches `MAX_QUARANTINE_SIZE_GB`, `PayloadExtractionJob.run()` breaks out of the download loop immediately:

```python
if self._quarantine_usage_bytes() >= max_size_bytes:
  self.log.warning("Quarantine directory has reached the limit. Stopping downloads.")
  break
```

No metadata-only record is created for the remaining payloads in the batch. Because the `/recent` endpoint uses strictly sequential, non-overlapping time windows based on `EXTRACTION_INTERVAL`, each payload appears in **exactly one** time window. If that window is skipped due to the quota, the automated pipeline will never see those payloads again.

There is no backfill, no retry path, and no manual recovery mechanism.

### Expected behavior

When the quota is reached, the job should still persist metadata-only stubs (`payload_file=None`) for all remaining payloads in the current batch.

### Next step
A follow-on improvement (separate issue) would be a management command or periodic task that sweeps `HoneypotPayload.objects.filter(payload_file__isnull=True, locator__gt="")` and retries downloads automatically when quota is available.

Prasad8830:
## Description

When the quarantine directory reaches `MAX_QUARANTINE_SIZE_GB`, `PayloadExtractionJob.run()` is invoked to `break` out of the download loop and silently drop all remaining payloads in the batch.

That loss is permanent. `_fetch_metadata` builds its window as `end_ts = time.time()` and `start_ts = end_ts - (EXTRACTION_INTERVAL * 60)`, and the job is scheduled at that same interval, so the windows are strictly sequential and non-overlapping. Each payload, therefore, appears in exactly one `/recent` window. Once a window is skipped, there is no cursor, no overlap and no retry path, so those files are never seen again.

This PR makes the job record what it did not download, rather than discarding it.

Key details:

- `run()` now passes the unprocessed remainder of the batch to a new `_store_metadata_only()` helper before breaking out of the loop.
- `_store_metadata_only()` writes rows with `payload_file` left empty and preserves the `locator`, which is the only handle the file can be fetched with later and exactly what was being thrown away. It also carries over `md5`, `sha1`, `mime_type`, `size` and `mtime` from the listing.
- Payloads with no locator are skipped rather than stored, since such a row would shadow the hash without offering any way to recover the file. This matches the existing behaviour in `_download_and_store`.
- The helper uses `get_or_create`, so a concurrent write from `_process_payload_hashes` cannot raise an `IntegrityError` against the `Lower("sha256")` unique constraint. If it finds an existing hash-only stub with no locator, it fills the locator in so the payload stays recoverable.
- Hashes are lower-cased on the way in, consistent with the normalisation added in #1556.
- The `log.error` call and its message are left untouched, since the quota notification added in #1555 depends on them.
- The completion log line now reports the deferred count separately from `skipped_count`, which already means "download failed".

### Note on scope

This PR only preserves the metadata. It does not add a sweep that retries these rows later. As discussed in #1532, nothing currently deletes quarantine files, so a sweep would find no freed quota to use until an admin removes files or raises the limit. That is tracked separately.

This also interacts with #1530: `_deduplicate` currently skips any hash that has at least one row, so the rows written here are not yet picked up for download. Both changes should ship in the same release.

## Related issues

- Closes #1532
- Related to #1530, which should ship in the same release
- Builds on #1555 and #1556, which touch the same code path

## Type of change

### Formalities

### Docs and tests

- [x] I have checked if my changes affect user-facing behaviour that is described in the docs. If so, I also included an update to the wiki in this PR's description.
- [x] Linter (Ruff) gave 0 errors. If you have installed pre-commit correctly, it performs these checks and adjustments on your behalf.

Tests added and updated in `tests/test_payload_extraction.py`

### GUI changes

N/A

### DIFF
--- a/greedybear/cronjobs/payload_extraction.py
+++ b/greedybear/cronjobs/payload_extraction.py
@@ -27,8 +27,10 @@ class PayloadExtractionJob(Cronjob):
     3. Checks quarantine disk usage against MAX_QUARANTINE_SIZE_GB before downloading.
     4. Downloads new payload files via ``/api/v1/payloads/download/{locator}``
        and saves them via QuarantineStorage.
-    5. Links each downloaded payload to any Cowrie sessions that already
-       transferred a file with the same SHA256.
+    5. Records metadata-only rows for any payload left undownloaded once the
+       quota is reached, so it is not lost with the extraction window.
+    6. Links every payload it records, downloaded or not, to the Cowrie sessions
+       and attacker IOCs that already transferred a file with the same SHA256.
     """
 
     # Timeout for the metadata listing request (seconds).
@@ -66,10 +68,15 @@ class PayloadExtractionJob(Cronjob):
             # Step 3: Download and store each new payload.
             downloaded = 0
             skipped_count = 0
-            for payload_meta in new_payloads:
+            deferred_count = 0
+            for index, payload_meta in enumerate(new_payloads):
                 # Check disk usage before each download.
                 if self._quarantine_usage_bytes() >= max_size_bytes:
                     self.log.error(f"Quarantine directory has reached the {settings.MAX_QUARANTINE_SIZE_GB} GB limit. Stopping downloads.")
+                    # Record what we did not get to. Every payload shows up in exactly
+                    # one /recent window, so dropping the rest of the batch here would
+                    # lose these files permanently.
+                    deferred_count = self._store_metadata_only(new_payloads[index:])
                     break
 
                 if self._download_and_store(client, server_url, payload_meta):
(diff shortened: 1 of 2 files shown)


regulartim (changes requested):
Hey @Prasad8830 ! Solid, thanks! 👍 A few things though: 

1. There's a merge conflict now, please resolve it.
2. The stub is never linked to Cowrie sessions or IOCs. I just merged PR #1552 that introduced a function that does that. Would be cool if you could integrate it here. 
3. Somethoing I just realized and that I want to bring up, so you're aware of it:  `payload_file` has two different values to express non-existence:  "" and NULL . We should change that in a follow-up.

regulartim (approved):
Hey @Prasad8830 ! Good work, thank you!

PR #1580, opened 2026-09-09
Normalize SHA256 case in payload API lookups. Closes #1579

### LINKED ISSUE #1579: Normalize hashes in payload API
The API endpoint at `/api/payloads/{sha256}` takes a sha256 hash and returns meta data for a corresponding payload. However the hash is user provided and might be upper case. We should normalize it, such that it matches the hashes in our DB.

aashu2006:
# Description

`/api/payloads/{sha256}` matched the hash exactly, so an uppercase hash from a user returned 404 even when the payload was there. Stored hashes are lower-cased on write, so anything not already lowercase missed

fixed by overriding `get_object` to lowercase the URL value and match on Lower("sha256"), matching on Lower() also covers older rows that were saved before hashes got normalized

This fixes `download` too, since it resolves the payload through the same `get_object()`

### Related issues

Closes #1579

Related: #1556 (same hash casing problem, but on the write side)

### Type of change

### Formalities

### Docs and tests

### Tests

5 new tests: uppercase URL, mixed case URL, a row stored uppercase, uppercase on the download route, and one checking the RBAC gate still applies since I rewrote the method that calls check_object_permissions.

also I reverted the fix and reran to confirm they are real, 4 of 5 fail without it. The RBAC one passes either way by design, it guards my change rather than the original bug.

### DIFF
--- a/api/views/payloads.py
+++ b/api/views/payloads.py
@@ -2,11 +2,13 @@
 # See the file 'LICENSE' for copying permission.
 import logging
 
+from django.db.models.functions import Lower
 from django.http import FileResponse
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiResponse, extend_schema, extend_schema_view
 from rest_framework import status, viewsets
 from rest_framework.decorators import action
+from rest_framework.generics import get_object_or_404
 from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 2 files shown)


regulartim (approved):
Nice, minimal fix. Thank you! :)

PR #1586, opened 2026-09-09
Allow querying Cowrie sessions by ID. Closes #1578

### LINKED ISSUE #1578: Allow querying Corwrie sessions by ID
Our new payload API now returns a list of session IDs. It would be nice if the Cowrie session API would be able to take an IDs and return session data of the corresponding session.

aashu2006:
# Description

The payloads API returns Cowrie session IDs, but there was no way to look a session up by one. I have added an `id` query parameter to /api/cowrie_session that takes the hex ID in the same format the payloads API returns

went with a separate parameter rather than another branch on `query`, since a hex ID and a password can look identical and the ID is not really an observable. `query` and `id` are mutually exclusive, giving both is a 400

The ID lookup does not apply the `duration__gt=0` filter other branches use, so asking for a specific session returns it either way

### Related issues

Closes #1578

Related to #1509 (added the session IDs to the payloads API that this consumes)

### Type of change

### Formalities

### Docs and tests

### Tests

7 new tests: basic ID lookup, with session data, a zero-duration session still being returned, case-insensitive hex, unknown ID 404, non-hex 400, and both parameters together 400.

reverted the source and reran to check they're real, 6 of the 7 fail without it. The non-hex one passes either way since a request with no `query` was already a 400, but it is still the right assertion to keep.

Note: `id` shadows a builtin so I checked ruff's flake8-builtins rules, no complaints across all 205 files. Happy to rename if you would rather avoid it.

### DIFF
--- a/api/serializers/cowrie_session.py
+++ b/api/serializers/cowrie_session.py
@@ -3,12 +3,20 @@ from rest_framework import serializers
 
 class CowrieSessionRequestSerializer(serializers.Serializer):
     query = serializers.CharField(
+        required=False,
         max_length=256,
         help_text=(
             "The search term, can be an IP address, the SHA-256 hash of a command sequence, or a password. "
-            'SHA-256 hashes should match command sequences generated using Python\'s `"\n".join(sequence)` format.'
+            'SHA-256 hashes should match command sequences generated using Python\'s `"\n".join(sequence)` format. '
+            "Mutually exclusive with `id`."
         ),
     )
+    id = serializers.CharField(
+        required=False,
+        max_length=32,
+        help_text=("Hex session ID, in the same format the payloads API returns. Mutually exclusive with `query`."),
+    )
+
     include_similar = serializers.BooleanField(
         required=False,
         default=False,
@@ -25,6 +33,20 @@ class CowrieSessionRequestSerializer(serializers.Serializer):
         required=False, default=False, help_text="When `true`, includes detailed information about matching Cowrie sessions."
     )
 
+    def validate_id(self, value: str) -> str:
+        """Session IDs are stored as integers, so the hex has to parse."""
+        try:
+            int(value, 16)
+        except ValueError as exc:
+            raise serializers.ValidationError(f"Not a valid hex session ID: {value}") from exc
+        return value
+
+    def validate(self, data: dict) -> dict:
+        """Exactly one of query or id, they select different things."""
+        if bool(data.get("query")) == bool(data.get("id")):
+            raise serializers.ValidationError("Provide either `query` or `id`, not both.")
+        return data
+
 
 class SessionDetailSerializer(serializers.Serializer):
     """A single matching Cowrie session."""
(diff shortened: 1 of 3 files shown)


api/serializers/cowrie_session.py
```diff
@@ -3,12 +3,20 @@
...
         help_text=(
             "The search term, can be an IP address, the SHA-256 hash of a command sequence, or a password. "
-            'SHA-256 hashes should match command sequences generated using Python\'s `"\n".join(sequence)` format.'
+            'SHA-256 hashes should match command sequences generated using Python\'s `"\n".join(sequence)` format. '
+            "Mutually exclusive with `id`."
         ),
     )
+    id = serializers.CharField(
+        required=False,
+        max_length=32,
```
regulartim: `CowrieSession.session_id` is a `BigIntegerField` which can store 64 bit. So 16 hex digits are sufficient, right?

api/serializers/cowrie_session.py
```diff
@@ -25,6 +33,20 @@ class CowrieSessionRequestSerializer(serializers.Serializer):
         required=False, default=False, help_text="When `true`, includes detailed information about matching Cowrie sessions."
     )
 
+    def validate_id(self, value: str) -> str:
+        """Session IDs are stored as integers, so the hex has to parse."""
+        try:
+            int(value, 16)
```
regulartim: This is not a good hex validator. Maybe create a regex in `greedybear/regex.py` and use that to validate.

api/serializers/cowrie_session.py
```diff
@@ -25,6 +33,20 @@ class CowrieSessionRequestSerializer(serializers.Serializer):
...
+        except ValueError as exc:
+            raise serializers.ValidationError(f"Not a valid hex session ID: {value}") from exc
+        return value
+
+    def validate(self, data: dict) -> dict:
+        """Exactly one of query or id, they select different things."""
+        if bool(data.get("query")) == bool(data.get("id")):
+            raise serializers.ValidationError("Provide either `query` or `id`, not both.")
```
regulartim: This will produce a misleading error message when neither parameter is given. Suggestion:

```python
if data.get("query") and data.get("id"):
      raise serializers.ValidationError("Provide either `query` or `id`, not both.")
if not data.get("query") and not data.get("id"):
      raise serializers.ValidationError("Provide either `query` or `id`.")

File: tests/api/views/test_cowrie_session_view.py [hunk 4]
```diff
@@ -165,6 +165,58 @@ def test_case_insensitive_boolean_parameters(self):
...
+        session = CowrieSession.objects.create(
+            session_id=int("abc123", 16),
+            start_time=self.current_time,
+            duration=0,
+            source=self.ioc,
+        )
+        response = self.client.get("/api/cowrie_session?id=abc123&include_session_data=true")
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.data["sessions"]), 1)
+        session.delete()
```
regulartim: I guess you don't need to delete the session here. CustomTestCase rolls back per test. Could you please check?

regulartim (changes requested):
Hey @aashu2006 ! Looks good, all in all, thanks! :) 

Two things apart from my comments:
1. Please provide an update for the wiki. Just a short text that describes the new parameter is sufficient. 
2. I think we talked about the `duration__gt=0` filter and after reviewing, I think we should also add it here. It's better to not have incomplete or broken sessions in the result, especially when clustering is enabled. Sorry for the confusion.

api/serializers/cowrie_session.py
aashu2006: yup, changed to 16. It's signed so the real max is 7fffffffffffffff, but I checked and bigger values just 404, no crash, added a test for that.

api/serializers/cowrie_session.py
aashu2006: `int()` was taking stuff like 0xff, -ff and f_f. I have added `REGEX_COWRIE_SESSION_ID` to `regex.py` and tests for those

api/serializers/cowrie_session.py
aashu2006: Done, used your version, added a test for each message.

aashu2006: Done, used your version, added a test for each message.

tests/api/views/test_cowrie_session_view.py
aashu2006: Removed it 👍

regulartim (approved):
Looks good, thank you! :)

Issue #1551, opened 2026-09-04
Inconsistencies in and around the tag model


regulartim:
In #522 we introduced the `Tag` model for external enrichment data, e.g. from AbuseIPDB and Abuse.ch. This is what the data model currently looks like: 

```
class Tag(models.Model):
    """Tags for IOCs from enrichment sources like ThreatFox and AbuseIPDB."""

    ioc = models.ForeignKey(IOC, on_delete=models.CASCADE, related_name="tags")
    key = models.CharField(max_length=128, db_index=True)
    value = models.CharField(max_length=256)
    source = models.CharField(max_length=64)  # e.g., "threatfox", "abuseipdb"
    added = models.DateTimeField(db_default=Now())
```

Currently, 5 sources write data to this model:
| Source | Keys written | Conditional | Write fn |
| --- | --- | --- | --- |
| threatfox | malware, threat_type, confidence_level | yes - checks THREATFOX_API_KEY | replace_tags_for_source |
| abuseipdb | confidence_of_abuse | yes - checks ABUSEIPDB_API_KEY | replace_tags_for_source |
| rdns | ptr_record | no | add_tags |
| credential_reuse | behavior | no | add_tags |
| tanner | attack_type | no | add_tags |

A few things are inconsistent regarding this model and/or could be improved:
1. As mentioned in #522, `ip_reputation` and the `firehol_categories` are external data and therefore _should_ live in the `Tag` model but currently don't. 
2. Data from `credential_reuse` and `tanner` are honeypot data and therefore _should not_ live in the `Tag` model.
3. Some source use `replace_tags_for_source` (all corresponding tags are deleted at the start of every job run), others use `add_tags` (new tags are added, existing tags are not deleted). This might make sense, but it should alt least be checked and maybe written in a comment what was the rationale behind the decision. 
4. Some of these enrichment jobs do very similar thing. Maybe we could create a common base class from which they all inherit, to reduce repetitive / duplicate code. 

All these points should be split up into individual issues. It is also important to know that the API needs to be adapted, which means those are braking changes that will require a new major release once they are merged.

rainnj:
I would like to work on this issue

regulartim:
Hey @rainnj ! Is my summary of the issue is correct in your opinion? What do you think is the best thing to start with? Maybe 3.? 

You can open separate issues for all the points I listed and assign yourself.

rainnj:
[...]

PR #1590, opened 2026-09-11
Add BaseEnrichmentJob/HttpEnrichmentJob base classes. Closes #1566

### LINKED ISSUE #1566: Share common base class across Tag model cronjobs. Progresses #1551 -4
this progresses point 4 of #1551

threatfox_feed.py, abuseipdb_feed.py, reverse_dns.py, and credential_reuse.py all follow a similar shape, a Cronjob subclass, an __init__ that sets up a TagRepository, and a run() method that builds tag entries and writes them.

Went through all 4 files and found one line thats duplicated in all of them

self.tag_repo = tag_repo if tag_repo is not None else TagRepository()

(credential_reuse.py writes it slightly differently using "or" instead of "if...else" but does the same thing)

Didnt find much else duplicated beyond this, each files run() logic, imports, and SOURCE_NAME are unique to that source. reverse_dns.py also has an extra ioc_repo dependency the other 3 dont need

What im planning to do

I want to create a new BaseEnrichmentCron class (extending the existing Cronjob base) that holds just this shared tag_repo setup. the 4 files would extend this instead of Cronjob directly, and drop their duplicate __init__ lines. reverse_dns.py would keep its own __init__ for the extra ioc_repo, just calling super().__init__(tag_repo) for the shared part

happy to adjust this approach if theres a better way to structure it.

rainnj:
# Description
This progresses point 4 of #1551.

## What I did

Created two new shared classes:

### `BaseEnrichmentJob` (`greedybear/cronjobs/enrichment/base_enrichment.py`)
- holds the shared `tag_repo` setup 
- adds a `WRITE_METHOD` attribute + `_write_tags()` helper, so subclasses can pick `replace_tags_for_source` or `add_tags` without duplicating that logic
- adds a default `_should_skip()` returning `False`

### `HttpEnrichmentJob` (`greedybear/cronjobs/enrichment/http_enrichment.py`)
- for jobs that fetch data via HTTP API calls
- defines 4 abstract methods each subclass must implement: `_should_skip`, `_fetch_feed`, `_parse_tags`, `_build_tag_entries`
- shares the IOC-matching query (`_match_iocs`) and the overall `run()` workflow, since these were the same between `threatfox_feed.py` and `abuseipdb_feed.py`

Moved `threatfox_feed.py`, `abuseipdb_feed.py`, and `reverse_dns.py` into a new `greedybear/cronjobs/enrichment/` folder.

`ThreatFoxCron` and `AbuseIPDBCron` now extend `HttpEnrichmentJob`. `ReverseDNSCron` extends `BaseEnrichmentJob` directly, since it doesn't fetch over HTTP (does DNS lookups + its own DB query instead).

## Testing

Updated all import references in `tasks.py` and the affected test files. Full test suite passes (1348/1348).

### Related issues

Progresses #1551 
Closes #1566 

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes

No GUI changes


greedybear/cronjobs/enrichment/base_enrichment.py
```diff
@@ -0,0 +1,24 @@
...
+        super().__init__()
+        self.tag_repo = tag_repo if tag_repo is not None else TagRepository()
+
+    def _write_tags(self, tag_entries):
+        if self.WRITE_METHOD == "replace":
+            return self.tag_repo.replace_tags_for_source(self.SOURCE_NAME, tag_entries)
+        return self.tag_repo.add_tags(self.SOURCE_NAME, tag_entries)
+
+    def _should_skip(self) -> bool:
+        return False
```
regulartim: I looks like this is dead code, the method isn't called anywhere, right? Or do you already anticipate that it will be used by a future subclass?

greedybear/cronjobs/enrichment/base_enrichment.py
```diff
@@ -0,0 +1,24 @@
...
+    WRITE_METHOD = "replace"  # default, matches majority of sources
+
+    def __init__(self, tag_repo=None):
+        super().__init__()
+        self.tag_repo = tag_repo if tag_repo is not None else TagRepository()
+
+    def _write_tags(self, tag_entries):
+        if self.WRITE_METHOD == "replace":
+            return self.tag_repo.replace_tags_for_source(self.SOURCE_NAME, tag_entries)
+        return self.tag_repo.add_tags(self.SOURCE_NAME, tag_entries)
```
regulartim: This method fails when there is a typo in `WRITE_METHOD`. I see two options: (1) raise a `ValueError` on an unrecognized value or (2) use an enum. Your decision! :)

regulartim: Also: `SOURCE_NAME` is never declared/enforced in the base class. Maybe you could enforce it by defining it as a abstract property?

greedybear/cronjobs/enrichment/threatfox_feed.py
```diff
@@ -0,0 +1,144 @@
...
+        with HttpClient() as client:
+            response = client.post(url, json=data, headers=headers, timeout=30)
+
+        json_data = response.json()
+
+        if json_data.get("query_status") != "ok":
+            self.log.warning(f"ThreatFox API returned non-OK status: {json_data.get('query_status')}")
+            return []
+
+        return json_data.get("data", [])
```
regulartim, line 45: If the API returns a non-OK status, the method returns an empty list. This means that effectively all tags are deleted without replacement. But the method's doc string says "on errors or non-OK status, existing tags are preserved to avoid losing enrichment data", which I think is the better behavior. 

If you agree I would suggest to (1) go back to the original behavior (e.g. return `None` and catch that in the caller method) and (2) add a test case that covers this (seed a tag and asserts it survives).

greedybear/cronjobs/enrichment/http_enrichment.py
```diff
@@ -0,0 +1,57 @@
...
+        pass
+
+    @abstractmethod
+    def _fetch_feed(self):
+        pass
+
+    @abstractmethod
+    def _parse_tags(self, raw_data) -> dict:
+        pass
+
```
regulartim, line 45: `_parse_tags` doesn't parse tags, it parses the feed, right? If so, `_parse_feed` would be a better name.

greedybear/cronjobs/enrichment/reverse_dns.py
```diff
@@ -34,6 +33,9 @@ class ReverseDNSCron(Cronjob):
     are rechecked on subsequent runs.
     """
 
+    SOURCE_NAME = "rdns"
```
regulartim, line 34: Source name is defined twice in this file.

greedybear/cronjobs/enrichment/http_enrichment.py
```diff
@@ -0,0 +1,57 @@
+from abc import abstractmethod
+
+import requests
+
+from greedybear.cronjobs.enrichment.base_enrichment import BaseEnrichmentJob
+from greedybear.models import IOC
+
+
+class HttpEnrichmentJob(BaseEnrichmentJob):
+    """
```
regulartim, line 10: The abstract methods in this class should have doc strings. Otherwise, if you implement a subclass, you don't know what these methods should do.

regulartim (changes requested):
Hey @rainnj ! Although I wrote some comments, I am very satisfied with your work. This is exactly what I thought of, good job! :)

regulartim (approved):
Nice, thank you! :)

greedybear/cronjobs/enrichment/base_enrichment.py
rainnj: Yes that was dead code, I have removed it from base_enrichment.py now

greedybear/cronjobs/enrichment/base_enrichment.py
rainnj: Went with the `ValueError` approach, added an explicit check for both `"replace"` and `"add"`, raising `ValueError` for anything else.

Also added `SOURCE_NAME` as an abstract property in `BaseEnrichmentJob` .

Both done

greedybear/cronjobs/enrichment/threatfox_feed.py
rainnj: I totally agree, went with option 1. `_fetch_feed()` now returns `None` on a non-OK status, and `run()` returns early in that case instead of calling `_write_tags`, so existing tags are preserved.

Added `test_tag_persists_failed_update` to test this

greedybear/cronjobs/enrichment/http_enrichment.py
rainnj: Agreed, All Done!

Issue #1568, opened 2026-09-08
Feature Request: Extend PayloadRepository with full CRUD methods and use it for all HoneypotPayload access


kikiscodedeliveryservice:
### Problem Context

Follow-up issue raised from @regulartim's review on #1552: since #1509 introduced `PayloadRepository`, it should own all `HoneypotPayload` access the way `IocRepository` and `CowrieSessionRepository` already do for their models, instead of only the two linking methods it currently has.

### Scope
Right now `HoneypotPayload` is still touched directly in a few other places:

1. `process_event.py::_process_payload_hashes` — creates hash-only stub rows via `get_or_create`.
2. `payload_extraction.py::_deduplicate` — looks up existing hashes to filter incoming payloads.
3. `payload_extraction.py::_download_and_store` — looks up an existing row via `get_or_create` and updates it, or creates a new one.

These should move onto proper `PayloadRepository` methods (e.g. `get_or_create_stub`, `get_existing_hashes`, `create_payload`).

### Note
This change needs to go after the #1530 bug is fixed so that it refactors the latest, bug-free code.

kikiscodedeliveryservice:
Hey @regulartim, wanted to clarify if these come under our scope as well:

- `malwarebazaar.py` also queries `HoneypotPayload` directly (`is_uploaded_mb`/`size` filtering), but it's a fully separate cronjob, should that be included too?

- `api/views/payloads.py`'s `HoneypotPayloadViewSet.queryset` also queries the model directly. Looking at how repositories get used elsewhere in the API layer — `TrendingBucketRepository` shows up in `feeds.py`, but only because that view method deals with business logic, not as a plain class-level `queryset`. `payloads.py`'s queryset is just a one-line `prefetch_related().order_by()` with no logic to encapsulate. Is that meant to stay a plain Django/DRF queryset, or should it move into our scope too?

Overall, these 5 call sites cover every place that ORM accesses `HoneypotPayload`.

regulartim:
I would also include the malwarebazaar cronjob. For the API, I am not so sure. The API only does read operations and not using the repository usually provides the necessary flexibility for special operations on these queries (like different orderings). So I would leave that out. What do you think?

kikiscodedeliveryservice:
I would include the malwarebazaar cronjob too.

[...]

PR #1593, opened 2026-09-12
fix: let payload extraction upgrade hash-only stubs instead of skipping them. Closes #1530

### LINKED ISSUE #1530: `PayloadExtractionJob._deduplicate` skips files that were first seen as hash-only stubs

kikiscodedeliveryservice:
# Description

`PayloadExtractionJob` was skipping any hash that already had a `HoneypotPayload` row, even if that row was just a metadata-only stub (no file yet) created by the event pipeline. Those stubs never had their file attached since dedup treated an existing record as one with the file already downloaded.

Now dedup only counts a hash as "already downloaded" if the row has a real file attached. Stub-only hashes are treated as new and go through the normal download flow, and `_download_and_store` now finds-and-updates an existing stub instead of blindly constructing a new row, so it doesn't hit the unique constraint on `sha256` when one already exists.

### Related issues

Closes #1530

### Type of change

# Checklist

### Formalities

### Docs and tests

- [x] I have checked if my changes affect user-facing behavior that is described in the docs. If so, I also included an update to the wiki in the description of this PR.

# Note
Instead of `get_or_create`, I used an explicit case-insensitive lookup (because the unique constraint is on `Lower(sha256)`, not the raw column) which might be overkill. I did it because I wanted the Extraction Job to not be dependent on the writes to sha256 always being lower cased (especially in cases where it could have been added from Django Admin).

### DIFF
--- a/greedybear/cronjobs/payload_extraction.py
+++ b/greedybear/cronjobs/payload_extraction.py
@@ -125,8 +125,13 @@ class PayloadExtractionJob(Cronjob):
 
     def _deduplicate(self, payloads: list[dict]) -> list[dict]:
         """
-        Filter out payloads whose SHA256 already exists in the database
-        and remove duplicates within the response itself.
+        Filter out payloads whose SHA256 already has a downloaded file in the
+        database, and remove duplicates within the response itself.
+
+        A hash can already have a HoneypotPayload row without a file attached -
+        a metadata-only stub created from a raw event before the file was
+        available. Those hashes are treated as new so the file can still be
+        downloaded.
 
         Args:
             payloads: List of payload metadata dicts (must contain 'sha256' key).
@@ -140,7 +145,10 @@ class PayloadExtractionJob(Cronjob):
         # through to fail on the constraint instead.
         incoming_hashes = {p["sha256"].lower() for p in payloads if "sha256" in p}
         existing_hashes = set(
-            HoneypotPayload.objects.annotate(sha256_lower=Lower("sha256")).filter(sha256_lower__in=incoming_hashes).values_list("sha256_lower", flat=True)
+            HoneypotPayload.objects.annotate(sha256_lower=Lower("sha256"))
+            .filter(sha256_lower__in=incoming_hashes)
+            .exclude(payload_file="")  # Django stores an unset FileField as an empty string.
+            .values_list("sha256_lower", flat=True)
         )
         new_hashes = incoming_hashes - existing_hashes
         self.log.debug(f"Deduplication: {len(incoming_hashes)} incoming, {len(existing_hashes)} existing, {len(new_hashes)} new.")
@@ -168,8 +176,9 @@ class PayloadExtractionJob(Cronjob):
 
     def _download_and_store(self, client: HttpClient, server_url: str, payload_meta: dict) -> bool:
         """
-        Download a single payload file, create its database record, and link
-        it to any Cowrie sessions that already transferred the same file.
+        Download a single payload file, create its database record (or upgrade
+        an existing hash-only stub with the real file), and link it to any
+        Cowrie sessions that already transferred the same file.
 
         Args:
             client: HttpClient instance.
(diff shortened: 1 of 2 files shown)


greedybear/cronjobs/payload_extraction.py
```diff
@@ -140,7 +145,10 @@ def _deduplicate(self, payloads: list[dict]) -> list[dict]:
         # through to fail on the constraint instead.
         incoming_hashes = {p["sha256"].lower() for p in payloads if "sha256" in p}
         existing_hashes = set(
-            HoneypotPayload.objects.annotate(sha256_lower=Lower("sha256")).filter(sha256_lower__in=incoming_hashes).values_list("sha256_lower", flat=True)
+            HoneypotPayload.objects.annotate(sha256_lower=Lower("sha256"))
+            .filter(sha256_lower__in=incoming_hashes)
+            .exclude(payload_file="")  # Django stores an unset FileField as an empty string.
```
regulartim, line 150: What if the field is NULL for some reason? I think we should cover that case.

greedybear/cronjobs/payload_extraction.py
```diff
@@ -199,23 +208,33 @@ def _download_and_store(self, client: HttpClient, server_url: str, payload_meta:
...
+        # Lower("sha256"), so a case-sensitive get_or_create(sha256=sha256) could
+        # miss a differently-cased row and hit IntegrityError on insert instead.
+        fields = {
+            "md5": payload_meta.get("md5", ""),
+            "sha1": payload_meta.get("sha1", ""),
+            "mime_type": payload_meta.get("mime_type", ""),
+            "size": len(file_content),
+            "locator": locator,
+            "mtime": payload_meta.get("mtime"),
+        }
```
regulartim, line 221: When upgrading, better only overwrite on truthy incoming values, with size and locator excepted. Otherwise a populated md5 might get overwritten with an empty string.

greedybear/cronjobs/payload_extraction.py
regulartim, line 26: This comment is now outdated.

greedybear/cronjobs/payload_extraction.py
```diff
@@ -199,23 +208,33 @@ def _download_and_store(self, client: HttpClient, server_url: str, payload_meta:
...
+        payload_obj = HoneypotPayload.objects.annotate(sha256_lower=Lower("sha256")).filter(sha256_lower=sha256).first()
+        created = payload_obj is None
+        if created:
+            payload_obj = HoneypotPayload(sha256=sha256)
+        for field, value in fields.items():
```
regulartim, line 226: This should be a method in `PayloadRepository`, do you agree? We can defer this to the PR that will close #1568 .

regulartim (changes requested):
Hey @kikiscodedeliveryservice ! Nice, correct fix. Well done. 

> Instead of get_or_create, I used an explicit case-insensitive lookup (because the unique constraint is on Lower(sha256), not the raw column) which might be overkill. I did it because I wanted the Extraction Job to not be dependent on the writes to sha256 always being lower cased (especially in cases where it could have been added from Django Admin).

I think you can assume that no payload will ever be added via Django admin and this is not something I want to support. So you can ignore this scenario. Is there any other reason not to use `get_or_create`?

In the tests, you are creating _real_ files in the filesystem. I would prefer if you could use a mock instead. `tests/test_malwarebazaar.py` does that, AFAIK.

greedybear/cronjobs/payload_extraction.py
kikiscodedeliveryservice: Yes, will add this to the issue #1568's scope.

kikiscodedeliveryservice:
Hey @regulartim, addressed your comments! No other reason not to use get_or_create(), so I've gone ahead with that. Mocked the files in the tests instead of creating them. Please re-review.

regulartim (approved):
Looks good, thank you! :)

PR #1600, opened 2026-09-15
Recover event batches stuck in processing after worker crash. Closes #1601

### LINKED ISSUE #1601: Event batch stuck in processing forever after qcluster worker is killed
## Description

An event batch can get stuck in `processing` forever if the qcluster worker dies while it is working on it (Django-Q timeout, OOM kill, container restart during a deploy). The sensor already got a `202 Accepted`, `GET /api/events/status/<task_id>/` keeps returning `processing`, and the batch's `RawEvent`s are never turned into IOCs.

## Cause

In `greedybear/process_event.py`, `process_incoming_event()` saves `status=PROCESSING` before entering `transaction.atomic()`:

```python
if batch.status in [EventStatusType.PROCESSING, EventStatusType.COMPLETED]:
    logger.warning(f"[task={task_id}] Batch already {batch.status} — skipping to avoid race condition")
    return

batch.status = EventStatusType.PROCESSING
...
batch.save(update_fields=["status", "ioc_count", "last_error"])
```

When the worker is killed inside the atomic block:

1. Postgres rolls back the IOC work, but `PROCESSING` was already committed.
2. The task was never acked, so the ORM broker hands it out again once `retry` (1860s) expires.
3. The redelivered call sees `PROCESSING`, logs the warning above and returns.
4. Django-Q stores that as a successful run and deletes the queue entry.
5. Nothing else ever enqueues the batch again (`async_task` is only called from `EventsCreateView.post`).

With `workers: 1` there is never a second worker actually processing the same batch, so in practice the guard only fires on redelivery, which is exactly the case where the batch should be retried.

## Steps to reproduce

1. Start the app and qcluster, create an `APISource`, a sensor and a token.
2. `POST /api/events/add/` with a batch.
3. While the batch is inside the atomic block (e.g. hold `LOCK TABLE greedybear_ioc IN ACCESS EXCLUSIVE MODE` from psql to keep it there), `kill -9` the qcluster processes, then start qcluster again.
4. Wait for the broker `retry` to expire.

Observed:

```
status: processing, ioc_count: 0, processed_at: null
RawEvents: total=2 processed=0
IOC for the source IP: none
django_q_task: success=true
log: [task=...] Batch already processing — skipping to avoid race condition
```

Setting the batch back to `pending` and enqueueing it again processes the same events fine, so the data is valid and the guard is the only blocker.

## Expected behaviour

A batch left in `processing` by a dead worker should be picked up again by the redelivered task and end up `completed` (or `failed`), instead of staying in `processing` forever.

## Environment

GreedyBear `develop` (also on `main` / v3.6.2), django-q2 1.11.1, PostgreSQL 17.

dev-aditya-hub:
# Description

`process_incoming_event` saves `PROCESSING` before entering `transaction.atomic()`. If the qcluster worker dies while a batch is inside that block (Django-Q timeout, OOM kill, `docker compose` restart), Postgres rolls the work back but the status stays committed. Because the task was never acked, the ORM broker hands it out again after `retry`. The redelivered call then hits the `already processing` guard and returns, Django-Q stores that as a success and deletes the queue entry. Nothing re-enqueues the batch after that: the sensor got a `202`, the status endpoint shows `processing` forever and the `RawEvent`s stay `processed=False`.

I reproduced this locally with the real API and qcluster on Postgres 17: posted a batch, held a lock on `greedybear_ioc` so the worker was mid-transaction, `kill -9`'d the cluster and restarted it. After redelivery the log shows `Batch already processing — skipping to avoid race condition`, the task row is `success=true`, and the batch never completes. Resetting the row to `pending` and enqueueing it again processes the same events fine, so the guard is the only thing in the way.

The fix:
- adds a nullable `EventStatus.started_at` (migration `0063`)
- replaces the read-check-save with `_claim_batch()`, a single conditional `UPDATE`: pending/failed batches can be claimed as before, and a `PROCESSING` batch can be taken over only once `started_at` is older than `Q_CLUSTER["timeout"]`. Django-Q kills tasks at `timeout` and only redelivers after `retry` (> `timeout`), so a batch that old can't still be running.

Since the claim is one `UPDATE ... WHERE`, the small window between reading and saving the status is gone too. Running the same crash scenario with this branch, the redelivered task picks the batch up and it ends `completed` with its IOC created.

Batches that are already stuck on an existing instance have no `started_at`, so this doesn't touch them (the existing guard tests rely on that); they need a one-off reset to `pending` and a re-enqueue.

### Related issues

Closes #1601

Related: #1363 (event ingestion API)

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes


regulartim:
Hey @dev-aditya-hub Your work looks solid on first sight. But you did not completely fill the checklist and I am wondering why.

> I have read and understood the rules about how to Contribute to this project.

This is really important, especially the "Rules" section. I would kindly ask you to read it carefully if you want to contribute here. On that note: the usual workflow in this project is to raise an issue **first** -> get assigned -> raise a PR. 

> I have reviewed and verified any LLM-generated code included in this PR.

This worries me. It seems you deliberately left that unchecked, which could mean that either your code is written without LLMs (which would be completely fine) or your code is written by a LLM but you did not review it before pushing. Could you please tell me what was your thought when leaving this box unchecked?

dev-aditya-hub:
Hi @regulartim thank you for taking the time to look at this and for explaining the process so clearly I really appreciate it

You are right about the workflow and that one is on me I came across the bug while going through the event ingestion code and got excited and opened the PR first then raised the issue afterwards as #1601 without waiting to be assigned
I have read the Rules section properly now and I understand why the order matters it helps avoid two people working on the same thing without knowing about each other
Please assign me to #1601 if you are happy to keep this PR or just let me know and I will close it and raise it again the right way I am completely fine with either option and I can move it back to draft in the meantime

dev-aditya-hub:
About the LLM checkbox I left it unchecked because I didnt want to tick something without properly thinking about it and not because I wanted to hide anything from you To be honest I did use an AI assistant while working on this so the honest answer is that it is LLM assisted and then reviewed and tested by me Before pushing the fix I reproduced the bug end to end on Postgres with the real API and qcluster running I posted a batch then held a lock on the IOC table so the worker was stuck inside the transaction then killed the cluster and restarted it The batch stayed in processing forever the raw events were never processed and django q recorded the retried task as a success

dev-aditya-hub:
I also went through the django q2 source to confirm how the retry behaviour actually works instead of just assuming it My first attempt at the fix broke two of the existing guard tests which showed me that the rule was too wide so I narrowed it down to only take over a batch when the started_at proves that it is older than the Django Q retry window
I also ran the full suite of 1375 tests and ruff and both were clean and then ran the crash scenario one more time to confirm that the batch now completes properly If you would prefer not to take LLM assisted code in this project please just tell me and I will close it without any hard feelings Otherwise I am happy to tick the box change it carefully and explain any part of the fix

Thanks again for the guidance I would rather learn the workflow properly now than keep getting it wrong

regulartim:
Hey @dev-aditya-hub ! Thanks for your detailed explanation. 

> Please assign me to https://github.com/GreedyBear-Project/GreedyBear/issues/1601 if you are happy to keep this PR 

Yep, I already assigned you. Please keep this PR open. 

> About the LLM checkbox I left it unchecked because I didnt want to tick something without properly thinking about it and not because I wanted to hide anything from you To be honest I did use an AI assistant while working on this so the honest answer is that it is LLM assisted and then reviewed and tested by me Before pushing the fix I reproduced the bug end to end on Postgres with the real API and qcluster running I posted a batch then held a lock on the IOC table so the worker was stuck inside the transaction then killed the cluster and restarted it The batch stayed in processing forever the raw events were never processed and django q recorded the retried task as a success

Thanks for your honesty, I appreciate that! :+1: The reason why I think it is important to check this box is that it should remind the author that what they push here is ultimately their responsibility. So even if the code is 100% LLM generated, you are the one pushing it and you are the one who will be held responsible for it. Therefore the author must confirm that they not just blindly push LLM code, by checking that box. It's a hard requirement for me. 

> I also went through the django q2 source to confirm how the retry behaviour actually works instead of just assuming it My first attempt at the fix broke two of the existing guard tests which showed me that the rule was too wide so I narrowed it down to only take over a batch when the started_at proves that it is older than the Django Q retry window
I also ran the full suite of 1375 tests and ruff and both were clean and then ran the crash scenario one more time to confirm that the batch now completes properly If you would prefer not to take LLM assisted code in this project please just tell me and I will close it without any hard feelings Otherwise I am happy to tick the box change it carefully and explain any part of the fix

We do accept LLM assisted code, I think almost every contribution from the last year is mostly LLM generated. No problem here. :) If you have reviewed and tested your code, as you wrote, you fulfilled our requirements and can confidently tick the box. Your code looked good at first sight and I'll be happy to give it a proper review.

dev-aditya-hub:
Thanks a lot @regulartim, and thanks for assigning me to #1601
That makes sense about the checkbox Whatever ends up in the PR is my responsibility no matter how it was written, so I am happy to stand behind this one I have ticked the box now

One small correction to my earlier comment I wrote that a batch is taken over once it is older than the Django Q retry window but it is actually the timeout The check in _claim_batch uses Q_CLUSTER timeout, which is 1800 seconds That is the safe value because Django Q kills a task once it hits the timeout and only redelivers it after retry which is longer So by the time a batch is that old no worker can still be processing it

Looking forward to your review, and I will pick up any changes you ask for

greedybear/process_event.py
```diff
@@ -192,6 +195,27 @@ def _process_payload_hashes(saved_ioc: IOC, ip_hits: list[dict]) -> None:
...
 
+def _claim_batch(task_id: str) -> bool:
+    """
+    Atomically move a batch to PROCESSING, returning False if it can't be claimed.
+
+    A PROCESSING batch older than the Django-Q timeout belongs to a worker that was
+    killed mid-run (timeout, OOM, restart). Its transaction was rolled back, so the
+    redelivered task takes it over instead of skipping it forever.
+    """
+    now = timezone.now()
```
regulartim: Please use `datetime.now()` for consistency with the rest of the project. I know that it has drawbacks, but we're only working with UTC time anyway, so it should not make a difference.

greedybear/process_event.py
```diff
@@ -192,6 +195,27 @@ def _process_payload_hashes(saved_ioc: IOC, ip_hits: list[dict]) -> None:
...
+    stale_before = now - timedelta(seconds=settings.Q_CLUSTER["timeout"])
+    claimable = Q(status__in=[EventStatusType.PENDING, EventStatusType.FAILED]) | Q(status=EventStatusType.PROCESSING, started_at__lt=stale_before)
```
regulartim: What if `started_at` is NULL? Such a batch would be permanently unclaimable, right?

regulartim (changes requested):
Hey @dev-aditya-hub ! Looks good! 👍 
Two things, besides my comments:
- a "test_completed_batch_is_not_reclaimed" would be nice
- the new filed is not visible in Django admin and also not in the API response, but I think both should contain it

greedybear/process_event.py
dev-aditya-hub: Changed to datetime.now() in 242757e, and the tests use it too now. USE_TZ is False so it makes no difference here, you are right.

greedybear/process_event.py
dev-aditya-hub: Good catch, you are right. A PROCESSING batch with started_at NULL could never be claimed, which is exactly the batches that are already stuck on a running instance before this change lands.

It is now treated as stale as well, so those get picked up on the next delivery instead of needing a manual reset. I also added a test for it. Two existing guard tests set the status to processing directly without a started_at, so I gave them one, since under the new code a batch a worker is really working on always has it.

dev-aditya-hub:
Thanks for the review @regulartim, all four points are done in 242757e.

The NULL started_at one was a real gap, thanks for catching it. Such a batch could never be claimed, which is exactly the state of batches that are already stuck on a running instance today. They are now treated as stale too, so they recover on the next delivery instead of needing a manual reset. Because of that I also had to give a started_at to the two existing guard tests that set the status to processing directly, since under the new logic a batch that a worker is really running always has one.

Also added test_completed_batch_is_not_reclaimed, switched to datetime.now(), and started_at now shows up in the Django admin list and detail view and in the batch status API response.

Full suite is green at 1378 tests and ruff is clean. Happy to change anything else.

dev-aditya-hub:
Hi @regulartim, everything from your review is in now, so this is ready for another look whenever you get a moment. Thanks again for the careful review, the NULL started_at catch really made the fix better :)

regulartim (approved):
Good job, thank you!

PR #1610, opened 2026-09-15
Keep feed cache version from expiring on bump. Closes #1609

### LINKED ISSUE #1609: Feeds serve stale data because the cache version expires after 5 minutes
## Description

The feeds sometimes serve an old list of IOCs even right after an extraction run has invalidated the cache. New IOCs are missing from the response and IOCs that changed since can show up in their old state, for up to an hour.

## Cause

`Cache.bump_data_version()` in `greedybear/cache.py`:

```python
def bump_data_version(self, version_key: str) -> None:
    cache = self._get_cache()
    try:
        cache.incr(version_key)
    except ValueError:
        cache.set(version_key, 2, timeout=None)
```

The first bump stores the version with `timeout=None`, but every bump after that goes through `cache.incr()`. The `api` cache is a `DatabaseCache`, which does not implement `incr()` itself, so Django's `BaseCache.incr()` is used. That one does a `get` and then `self.set(key, new_value)` without a timeout, so the version is saved again with the default timeout of 300 seconds.

With the default `EXTRACTION_INTERVAL` of 10 minutes this is what happens:

1. A run bumps the version from 2 to 3. The key now expires after 5 minutes.
2. 5 minutes later the key is gone and `get_data_version()` returns 1.
3. The next run finds no key, `incr()` raises `ValueError` and the version is set to 2 again.
4. Cached feed responses are stored for an hour under keys that include the version, so the response cached the last time the version was 2 is still there and gets served.

So the version keeps going 2, 3, 1, 2, 3, 1 instead of only going up, and each time it comes back to a number that was used before, an old response is returned.

The trending feed uses the same method with its own version key, so it has the same problem.

## Steps to reproduce

With the default `api` DatabaseCache on Postgres:

1. Create an IOC on an active honeypot, call `Cache("api").bump_data_version(IOC_DATA_VERSION_KEY)` and request `GET /api/feeds/`.
2. Create a second IOC, bump again and request the feed.
3. Wait a bit more than 300 seconds.
4. Create a third IOC, bump again and request the feed.

Output:

```
bump -> version=2 expires=9999-12-31 23:59:59
GET /api/feeds/ -> ['1.1.1.1']                          db has ['1.1.1.1']
bump -> version=3 expires=2026-09-15 18:43:45
GET /api/feeds/ -> ['1.1.1.1', '2.2.2.2']               db has ['1.1.1.1', '2.2.2.2']
wait 305s -> version=1
bump -> version=2 expires=9999-12-31 23:59:59
GET /api/feeds/ -> ['1.1.1.1']                          db has ['1.1.1.1', '2.2.2.2', '3.3.3.3']
```

The last response is the one cached in step 1.

## Why the tests don't catch it

`tests/test_cache.py` overrides the `api` cache with `LocMemCache`. `LocMemCache.incr()` updates the value in place and keeps the existing expiry, so `test_repeated_bumps_keep_incrementing` passes there while `DatabaseCache` behaves differently.

## Expected behaviour

Bumping the version should never let it expire, so that every bump moves to a version that has not been used before and the feeds show the current data.

## Environment

GreedyBear `develop`, Django 5.2.17, PostgreSQL 17.

dev-aditya-hub:
# Description

`Cache.bump_data_version()` used `cache.incr()` for every bump after the first one. On the `api` DatabaseCache, `incr()` comes from Django's `BaseCache`, which saves the new value with the default timeout of 300 seconds. With extraction running every 10 minutes the version key expires between runs, falls back to 1 and then cycles through numbers that were already used. Feed responses are cached for an hour under keys that include the version, so after a bump the feeds can return a response cached one or two runs earlier. Details and a reproduction are in #1609.

The fix reads the current version and stores `version + 1` with `timeout=None`, so the version never expires and every bump moves to a number that has not been used yet. This works the same on every cache backend. It is not less safe than before, since `incr()` on DatabaseCache was already a separate get and set, and extraction runs one at a time.

Existing instances do not need any manual step: the first bump after the update stores the version without expiry.

I added a test that uses a DatabaseCache instead of the LocMemCache override used by the other cache tests. It sets `TIMEOUT: 0`, so anything saved with the default timeout expires immediately. On the old code it fails with `1 != 3`, with the fix it passes.

I also ran the same scenario from the issue again on Postgres with the fix, including the 305 second wait. The version stays at 3 after the wait and the feed returns all three IOCs after the next bump.

### Related issues

Closes #1609

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes


greedybear/cache.py
```diff
@@ -40,9 +40,10 @@ def get_data_version(self, version_key: str) -> int:
...
-            cache.incr(version_key)
-        except ValueError:
-            cache.set(version_key, 2, timeout=None)
+        """Invalidate every cache entry keyed off version_key by incrementing it.
+
+        The version is always stored without expiry. cache.incr() can't be used here:
+        on DatabaseCache it re-saves with the default timeout, so the version would
+        expire, fall back to 1 and start matching old cached responses again.
+        """
+        self._get_cache().set(version_key, self.get_data_version(version_key) + 1, timeout=None)
```
regulartim: I guess this is the correct fix. Two things here: 
- Please shorten the comment a bit. The explanation why `cache.incr()` can not be used belongs into the issue and/or PR (where it is already present). The comment about the expiry can stay. 
- It looks like you could use the class's own wrapper here: `self.set` already defaults to `timeout=None`.

regulartim (changes requested):
Very nice fix and a genuine regression test! 👍 Only two minor nits.

greedybear/cache.py
dev-aditya-hub: Both done in 8595f86. The comment now only mentions the expiry, and it uses self.set, which already defaults to timeout=None. Thanks for spotting that, it reads better.

Full suite and ruff are still clean.

dev-aditya-hub:
Thanks a lot for the review @regulartim, I have re-requested it now.

I really appreciate the guidance through this one. Without your feedback I would not have realised where I was going wrong, both with the workflow and with small things like keeping the reasoning in the issue instead of the code comment. That is the kind of thing I would have kept repeating otherwise.

Looking forward to contributing more here and learning as I go. Happy to pick up another issue whenever you have one that suits.

dev-aditya-hub:
Hi @regulartim, both nits are done, so this one is ready for another look too whenever you have time. Thanks a lot :)

regulartim (approved):
Nice, thanks! :)

Issue #1543, opened 2026-09-03
Add a list of payload hashes to the Feeds API responses


regulartim:
We started to collect payload data, but the Feeds API currently does not return any information about which payloads were used by certain IP addresses. We could do that by adding a list of SHA256 hashes to the responses.

Demiserular:
i would love to wok on it @regulartim

regulartim:
Hey @Demiserular ! :) Alright, I'll assign you.

Prasad8830:
Hi @Demiserular, are you working on this issue? Because my issue #1532 is blocked due to this issue. If you are busy, I can take a look here also, like possible fixes, and all, so you can implement the fixes

regulartim:
Hi @Demiserular ! Since I want to see this implemented and you did not raise a PR for two weeks, I'll assign @Prasad8830 . Sorry.

Demiserular:
I had few questions, but surely @Prasad8830 can work on it.

Prasad8830:
> I had few questions, but surely [@Prasad8830](https://github.com/Prasad8830) can work on it.

Let's discuss them here before any implementation. So that we will reach at better decisions

Demiserular:
Before implementation, I wanted to clarify the expected API shape so we don’t accidentally expose the wrong relationship.

A few questions:
1. Should the payload SHA256 hashes be returned directly on each IP/feed item, e.g. `payload_hashes: string[]`, or should they be nested under a more explicit field like `payloads` for future metadata?
2. Do we want unique hashes only per IP, or should repeated payload usage be represented somehow?
3. Should this be added to all Feeds API responses, or only the endpoints where IP-level payload data is already available in the backend?
4. If an IP has no payloads, should the response include an empty list or omit the field?

Once these are clear, the implementation should be straightforward.

Prasad8830:
> Once these are clear, the implementation should be straightforward.

thanks, I will also thorugh them

Prasad8830:
Hi @regulartim, @Demiserular, I went through the code. Here is my understanding and proposed plan.

**Understanding**

The data already exists, only the API output is missing:
- `HoneypotPayload.iocs` (M2M, `related_name="payloads"`) is populated from two paths: Cowrie file transfers matched by SHA256 in `PayloadRepository`, and `_process_payload_hashes` in the events API flow.
[...]

PR #1612, opened 2026-09-17
Add payload hashes to the advanced feed responses. Closes #1543

Prasad8830:
## Description

GreedyBear links collected payloads to the attacker IOCs that transferred them (`HoneypotPayload.iocs`, filled from matched Cowrie file transfers and from `_process_payload_hashes` in the events flow). The Feeds API did not expose that link, so there was no way to go from an IP in a feed to the payloads it used.

This PR adds a `payload_hashes` list to each IOC returned by the authenticated advanced feed, as agreed in #1543.

Key details:

- `AdvancedFeedView.get_queryset()` annotates `payload_hashes` only for the `json` and `ndjson` formats. The public simple and paginated feeds, and the `txt`, `csv` and `stix21` outputs, are unchanged.
- The hashes come from an `ArraySubquery` rather than another `ArrayAgg`. The queryset already aggregates honeypots, tags, sensors and credentials in one `GROUP BY`, and each additional multi-valued join multiplies the rows per IOC. The subquery avoids that, and `ARRAY(subquery)` returns `[]` when an IOC has no payloads.
- Hashes are lowercased with `Lower("sha256")` and sorted, so the output is consistent even for rows stored before the case normalisation, and responses stay stable.
- `stream_ioc_objects` only emits `payload_hashes` when the annotation is present, following the existing `sensors_json` pattern. Callers without the annotation, such as the ML scoring path, keep their current row shape.
- The shared token endpoint (`/api/feeds/consume/<token>`) inherits the advanced feed response, so it includes the hashes too, as confirmed in the issue.
- `payload_hashes` is added to `AdvancedFeedResponseSerializer`, so the field shows up in the API schema.

Example of an IOC row in the advanced feed:

```json
{
  "value": "140.246.171.141",
  "credential_count": 1,
  "sensors": [],
  "payload_hashes": [
    "0a1b...e9f0",
    "7c2d...41aa"
  ]
}
```

### Note on scope

STIX 2.1 support for payload hashes is left for a follow-up, as discussed in #1543. I will open a separate issue for it.

### Wiki update

Proposed addition to the "Advanced Feeds API" section of the [Usage](https://github.com/GreedyBear-Project/GreedyBear/wiki/Usage) page:

> The `json` and `ndjson` responses include an additional field:
>
> _Available from version >= X.Y.Z_
>
> - `payload_hashes`: sorted list of lowercase SHA256 hashes of the payloads observed from the IOC. Empty list if none were observed.

(Version to be filled in with the release this ships in.)

## Related issues

- Closes #1543
- Builds on #1552, which links `HoneypotPayload` to Cowrie sessions and attacker IOCs
- Related to #1580, which normalises SHA256 case in payload lookups

## Type of change

### Formalities

### Docs and tests

- [x] I have checked if my changes affect user-facing behaviour that is described in the docs. If so, I also included an update to the wiki in this PR's description.
- [x] Linter (Ruff) gave 0 errors. If you have installed pre-commit correctly, it performs these checks and adjustments on your behalf.

Tests added in:
- `tests/api/views/test_feeds_advanced_view.py`: hashes in the JSON, NDJSON, paginated and shared token responses; lowercase and sorted output; no hashes leaking between IOCs; `[]` for IOCs without payloads; `txt` and `csv` unchanged
- `tests/api/views/test_feeds_view.py`: the field is absent from the public simple, paginated and NDJSON feeds
- `tests/api/views/test_utils.py`: `stream_ioc_objects` emits the field only when the annotation is present

### GUI changes

N/A


regulartim (changes requested):
Hey @Prasad8830 ! Looks good, thanks! :) One thing that came to my mind: the list is basically an unbound array that could get arbitrarily large. So maybe we should gate it behind the `verbose` flag, i.e. only return it if `verbos=True`. What do you think?

Prasad8830:
> Hey @Prasad8830 ! Looks good, thanks! :) One thing that came to my mind: the list is basically an unbound array that could get arbitrarily large. So maybe we should gate it behind the `verbose` flag, i.e. only return it if `verbos=True`. What do you think?

That makes sense, @regulartim, thanks for flagging! Unlike `tags`/`sensors`, payload hashes have no natural cap, a persistent scanner could accumulate hundreds of them over time, so it shouldn't be in the default payload.

I'll gate it behind `verbose=true`, same as `days_seen` and `firehol_categories`, so it only computes and returns it when requested. Will update the PR shortly.

Prasad8830:
Hi @regulartim, could you please re-review?

regulartim (approved):
Nice, thank you! :) I'll shorten your comment a bit and merge.

Issue #1584, opened 2026-09-09
Remove Tanner honeypot data from `Tag` model


regulartim:
Tanner is a http honeypot and part of T-Pot. Some of its data is currently written into the `Tag` model, although it arguably belongs into the `IOC` model directly. 

Suggestion: Instead of writing a tag called `attack_type`, the `TannerExtractionStrategy` should write in a new `IOC.http_attack_types` field of type `pg_fields.ArrayField`. 

**Note**: this involves changes in multiple parts of GreedyBear:
- changes in the `IOC` model
- the `TannerExtractionStrategy` and subsequent method calls, like `IocProcessor._merge`
- several test cases and fixtures
- all API endpoints returning IoC data (they should include the new field)
- Django admin interface
- maybe more

Also we have to migrate the existing data from the `Tag` model into the new field.

Issue #1559, opened 2026-09-07
Tag model: document replace_tags_for_source vs add_tags rationale. Progresses #1551-point3


rainnj:
This issue progresses point 3 of #1551.

Currently, 5 sources write to the Tag model, using 2 different write methods:

replace_tags_for_source (deletes existing tags from that source, then writes fresh ones):
- threatfox
- abuseipdb

add_tags (adds new tags, never deletes existing ones):
- reverse_dns (rdns)
- credential_reuse
- tanner

I looked into the PRs/issues that introduced each of these to check if this split was intentional or accidental.

Findings:

threatfox and abuseipdb use replace_tags_for_source because both jobs re-fetch and re-check their entire external feed against all
IOCs every run, producing a full, current snapshot each time. Confirmed in PR #867: "tags are replaced atomically on each run
for data freshness."

reverse_dns, credential_reuse, and tanner use add_tags because these jobs only process a small batch of new candidates each run
(each excludes IOCs already tagged from that source in their query). If they used replace_tags_for_source instead, every tag created in a previous run would be deleted just because that IOC isn't part of today's batch, wiping out still-valid data.

This was explicitly confirmed for reverse_dns in PR #934, where the author states they used add_tags "for incremental tag creation, as opposed to the atomic replace used by ThreatFox/AbuseIPDB." credential_reuse and tanner don't have this explicitly discussed in their PRs, but both follow the exact same query pattern, so the same reasoning applies even though it wasn't written down at the time.

Proposed next step:

Add a short comment above the add_tags call in reverse_dns.py, credential_reuse.py, and tanner.py, explaining why add_tags is used instead of replace_tags_for_source, so this reasoning is documented directly in the code and not just in PR history.

Happy to hear if this reasoning holds up or if there's something I'm missing.

regulartim:
Sounds good! 👍 Feel free to add these comments and open a PR.

regulartim:
Thanks! :) Could you also add comments to the `add_tags` function calls? After that, I guess we can close this issue.

PR #1623, opened 2026-09-18
Move Tanner attack types to IOC.http_attack_types field. Closes #1584

rainnj:
# Description

Tanner's attack types  were stored in the Tag model, but that model is meant for external enrichment data. Tanner is our own honeypot, so this data belongs on the IOC record directly. This PR moves it to a new `IOC.http_attack_types` ArrayField and migrates the existing tags over.

For the data migration I followed the approach from #902 and did it in raw SQL (array_agg + GROUP BY) so it runs as one statement instead of looping in Python.

I also added `bulk_add_http_attack_types` to IocRepository. It merges with the existing values rather than replacing them, because Tanner only processes the current extraction chunk each run - same reasoning as why `add_tags` was used over `replace_tags_for_source` before (#1559).

### Related issues

Progresses #1551 
Closes #1584 

### Type of change

# Checklist

### Formalities

### Docs and tests

### GUI changes

No GUI changes
  
# Review process
One point of discussion that I have, I put the field in `JSON_BASE_FIELDS` instead of `JSON_VERBOSE_FIELDS`. This data used to reach API users through `tags`, which is in the base list, but `firehol_categories` is in verbose, so there's an argument the other way.


regulartim:
> One point of discussion that I have, I put the field in JSON_BASE_FIELDS instead of JSON_VERBOSE_FIELDS. This data used to reach API users through tags, which is in the base list, but firehol_categories is in verbose, so there's an argument the other way.

I think it belongs into the base list.

api/views/utils.py
```diff
@@ -80,6 +80,7 @@ def ioc_as_dict(ioc, fields: set) -> dict:
     "interaction_count",
     "scanner",
     "payload_request",
+    "http_attack_types",
```
regulartim, line 83: The addition here should make the field present in the API response, but not in the OpenAPI docs. Please also add it to the serializers (`SimpleFeedResponseSerializer` might be sufficient).

greedybear/models.py
```diff
@@ -169,6 +169,7 @@ class IOC(models.Model):
     payload_request = models.BooleanField(default=False)
     related_ioc = models.ManyToManyField("self", blank=True, symmetrical=True)
     related_urls = pg_fields.ArrayField(models.CharField(max_length=900, blank=True), blank=True, default=list)
+    http_attack_types = pg_fields.ArrayField(models.CharField(max_length=64, blank=True), blank=True, default=list)
```
regulartim, line 172: The max_length seems sufficient here, but what if during the data migration there's a stray long value? It would aborts it. Either  increase the length or use something like `LEFT(value, 64)` in the data migration.

greedybear/cronjobs/repositories/ioc.py
```diff
@@ -302,3 +302,34 @@ def bulk_update_ioc_reputation(self, ip_addresses: list[str], reputation: str) -
         if not ip_addresses:
             return 0
         return IOC.objects.filter(name__in=ip_addresses).update(ip_reputation=reputation)
+
+    def bulk_add_http_attack_types(self, attack_types_by_ioc: dict[int, set[str]], batch_size: int = 1000) -> int:
```
regulartim: Do we really need this method? Can't we just create a new IOC object in the tanner strategy and merge it?

regulartim (changes requested):
Hey @rainnj ! Looks good, all in all! Thank you! One thing regarding the tests:  there's no test which asserts the field reaches the API, but that's quite important.

greedybear/cronjobs/repositories/ioc.py
rainnj: Okay so I had a look at the code, I agree that it is redundant, I'm thinking of using add_ioc and _merge_iocs

regulartim: Alright, thanks! :) So I just wanted to start investigating, but since you already seem to have a plan, I'll wait until you pushed your changes. :)

regulartim:
Hey @rainnj ! I'll give this another review now. What clearly needs to be done is solving the merge conflict and renumbering the migrations.

regulartim (changes requested):
I really like what you did here, this is a genuine improvement over your first commits. What's still missing (please correct me if I am wrong) is a test which asserts the new field reaches the API, right?

greedybear/cronjobs/repositories/ioc.py
rainnj: Done in 786ad3a. Detection now runs first and the types are set on the IOC before add_ioc, so _merge_iocs handles the union and the method is gone. Had to split _classify_attacks in two so the RFI part still runs after the scanners are saved.

rainnj:
> I really like what you did here, this is a genuine improvement over your first commits. What's still missing (please correct me if I am wrong) is a test which asserts the new field reaches the API, right?

Thank you @regulartim! Yes sorry I updated it now, let me know if there is anything else you need me to change.

regulartim (approved):
Hi @rainnj ! Good job! Really impressive what you did here. I'll merge it as soon as the next patch release is out.

rainnj:
> Hi @rainnj ! Good job! Really impressive what you did here. I'll merge it as soon as the next patch release is out. 

Thank you Tim!!

Issue #1625, opened 2026-09-19
bug : Scoring jobs crash on empty IoC data instead of skipping cleanly


AdeshDeshmukh:
## Description

While reading through the ML scoring pipeline (`greedybear/cronjobs/scoring/scoring_jobs.py`) to understand how model training and score updates fit into the nightly jobs, I noticed both entry points assume there is always data to work with — and I don't think that assumption always holds.

In `TrainModels.run()` (`scoring_jobs.py:96-98`), the freshly fetched data goes straight into `max()` with no empty check:

```python
self.current_data = get_current_data()
current_date = max(row["last_seen"] for row in self.current_data)
```
UpdateScores.run() has the identical pattern at :262-265. The existing if not training_data: guard a few lines below (:103) only protects the previously saved data, not the fresh fetch.
get_current_data() (scoring/utils.py:159-180) returns scanners seen in the last 30 days associated with active honeypots — so it comes back empty on fresh installs before the first hits land, on quiet private deployments, or if all honeypots are deactivated. On any such night, the midnight train_and_update() (tasks.py:46-54) dies on the ValueError, and since failures propagate (base.py:21-23 re-raises), the task fails loudly. Worse, trainer.execute() raising means the UpdateScores step never runs either, and save_training_data() is never reached — so one quiet night silences both training and scoring, with no self-recovery.
## Cause
Unguarded max(row["last_seen"] ...) over a possibly-empty dataset in TrainModels.run() (:98) and UpdateScores.run() (:265).

## Steps to Reproduce
1. Start from a state with no scanners seen in the last 30 days (fresh install, or deactivate all honeypots).
2. Run the midnight job (train_and_update() → TrainModels().execute()).
3. Observe: ValueError: max() arg is an empty sequence, task fails, neither training data nor scores are refreshed.

## Expected Behaviour
An empty data window should log a warning and skip the run cleanly — matching the existing if not training_data: style a few lines below — instead of crashing the nightly jobs.

## Proposed Approach
Early-return guards in both run() methods (house log style, mirroring :104):
self.current_data = get_current_data()
if not self.current_data:
```python 
    self.log.warning("no current IoC data, skipping training run")
    return
```
and the same for self.data in UpdateScores.run(). Both guards are needed as a unit: tasks.py:52-54 hands trainer.current_data straight into updater.data, so guarding only the trainer would still crash the updater on the handoff ([] is not None, so its own fetch is skipped).
[...]

PR #1627, opened 2026-09-21
Guard scoring jobs against empty IoC data. Closes #1625

AdeshDeshmukh:
# Description

Small robustness fix in the ML scoring jobs. Both `TrainModels.run()` and `UpdateScores.run()` feed freshly fetched data straight into `max()` with no empty check — on a night with no scanners seen in the last 30 days (fresh installs, no-T-Pot instances, all honeypots deactivated), the midnight job dies on `ValueError`, and since failures propagate, training data never refreshes either.

Four early-return guards (two per `run()` method): empty input list *and* empty feature frame — the second trapdoor exists because `get_features()` skips every IOC with empty `days_seen`, so non-empty input can still yield an empty table. All four log at error level and return cleanly. No migration, no API change, `score_only` untouched.

### Related issues

Closes #1625

### Type of change

# Checklist

### Formalities

- [x] I chose an appropriate title for the pull request in the form: `Guard scoring jobs against empty IoC data. Closes #1625`

### Docs and tests

- [x] I have checked if my changes affect user-facing behavior that is described in the [docs](https://github.com/GreedyBear-Project/GreedyBear/wiki). If so, I also included an update to the [wiki](https://github.com/GreedyBear-Project/GreedyBear/wiki) in the description of this PR. (Nightly-job robustness fix, no docs change needed.)
- [x] I have added tests for the feature/bug I solved (3 tests in new `tests/test_scoring_jobs.py`: empty-fetch skip, empty-data skip, empty-frame skip with train/save/update asserted uncalled).
- [ ] All the tests gave 0 errors. (Ruff + format green locally; full Django suite needs CI — local `pg_config` blocker.)


greedybear/cronjobs/scoring/scoring_jobs.py
```diff
@@ -125,6 +128,9 @@ def run(self):
 
         self.log.info("extracting features from training data")
         training_df = get_features(training_data, training_date)
+        if training_df.empty:
+            self.log.error("no features extracted from training data, skipping training run")
+            return
```
regulartim, line 134: I think we should call `save_training_data` here before returning. Otherwise we might enter an unrecoverable state.

regulartim (changes requested):
Hey @AdeshDeshmukh ! Good job, thank you! Only found one thing.

AdeshDeshmukh:
Hi @regulartim, done — applied exactly as you suggested...!

The training_df.empty guard now calls self.save_training_data() before returning, so the snapshot keeps advancing on skipped nights and the next trainable run computes deltas against a fresh baseline instead of a frozen one. That was the right call — without it the stale-snapshot rot you described would just sit there silently.

SCR-20260922-mukt

One deliberate non-change I want to confirm: the empty-input guard stays a bare return with no save. Saving there would overwrite training_data.json with [] and destroy the last good snapshot, so I left it — and added a test asserting save is not called on that path, mirroring the assert_called_once on the frame path, so the distinction is locked in for future edits.

Ruff + format green locally; full suite via CI once pushed. score_only untouched as agreed.

thanks for ur suggestion.!

regulartim:
> done — applied exactly as you suggested

But did you push these changes? I can't see them.

regulartim:
> One deliberate non-change I want to confirm: the empty-input guard stays a bare return with no save. Saving there would overwrite training_data.json with [] and destroy the last good snapshot, so I left it — and added a test asserting save is not called on that path, mirroring the assert_called_once on the frame path, so the distinction is locked in for future edits.

That's fine I think.

AdeshDeshmukh:
> But did you push these changes? I can't see them.
>
I missed to push in the rush of college lecture!
Done now!

regulartim (approved):
Thanks! :)

